PR #1, opened 2016-04-06
add a fetch function to the WorldArgs

csherratt:
This is a bit more verbose, and I am open to alternatives.

The `WorldArgs` from before was a bit implicit in when the next system was triggered to go. It was triggered off of access to `entities`. This caused issue for me when the first thing I tried was a call that did not use this api.


kvark:
Nice, thank you!

PR #7, opened 2016-04-08
Entity deletion via a deferred queue

### LINKED ISSUE #3: Entity deletion
May require some sort of recycling with generational IDs.

kvark:
Fixes #3 

Needs to be discussed and evaluated before merging, since there are use-cases that would make it leak data. For example, if any component is not locked for writing for a long time, it's deletion queue will grow unbound even though the affected entities may not even have this component.

On the other hand, this change is not visible in the API and can be reversed at any point later on.


kvark:
Closing in favor of #8

PR #9, opened 2016-04-09
use fnv hasher

csherratt:
In my benchmarks this is twice as fast as the Sip hasher fore only 5 extra lines.


kvark:
Sweeet, thanks!
If you happen to benchmark some of the `parsec` stuff, please share the results.

PR #10, opened 2016-04-09
add some tests!

csherratt:
VecStorage::sub would panic if you delete an out of generation id.
Add tests to make sure the behavior of vec/hashmap is consistent.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -30,6 +30,12 @@ pub type Index = u32;
 pub struct Entity(Index, Generation);
 
 impl Entity {
+    #[cfg(test)]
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -30,6 +30,12 @@ pub type Index = u32;
 pub struct Entity(Index, Generation);
 
 impl Entity {
+    #[cfg(test)]
+    /// Create a new entity (externally from ECS)
+    pub fn new(index: u32, gen: i32) -> Entity {
```
kvark: I don't want `Entity::new()` to be exposed. Can we get away without it?

PR #14, opened 2016-04-09
Dynamic entity insertion/removal

### LINKED ISSUE #11: RFC: dynamic entity creation and deletion
Dynamic operation means that it happens during a system execution as opposed to being outside of the system processing (this is already in).

### Appendix
We could add `RwLock<Appendix>` to the `Scheduler`, where:
```rust
struct Appendix {
    next: Entity,
    queue: Vec<Entity>,
}
```
Note: `World` doesn't know anything about it.
Then, `WorldArg` would get methods `insert() -> Entity` and `remove(Entity)`. These would lock the `Appendix` for writing and update both `next` and the `queue`.

Note: when removing an entity, the value added to the queue can have its generation inverted right away. This allows to avoid having two separate queues for insertion/removal, or an additional enum type.
Note: we propose `next: Entity` instead of `next_free: Index`, because when an entity gets dynamically deleted, we can no longer use `generations` vector for its new generation.

### Latency
We want the newly created entities to start participating right away. The use-case is: player input generates a bullet, it gets its first movement and rendered the same frame.
Note: removing entities doesn't have to have low latency. E.g. if the player killed something, it's more important to show the explosion particles quickly than hiding the object itself.

Locking `generations` for writing is not an option, since it's getting used simultaneously by multiple systems all the time (fore reading). Thus, we record entity insertion and removal in the `Appendix`.

In order to be able to work with the new entities, we propose having `entities_extra(&self) -> EntitiesExtraIter` method in the `WorldArg`. It would lock the appendix for reading, go through the queue, and return those entities that were added. We can also update the `run` method to always process extra entities by default, or make it configurable.

Note: we don't force the user to iterate over extra entities in their custom systems. It's their decision, and they can even decide on that dynamically during the system execution.

### Rest
Of course, we should not be using the appendix across multiple frames without merging it with the world. Otherwise we'd suffer from the lock waits. To address this, we could have a `rest()` method in the scheduler. The implementation would lock both `world.generations` and `appendix` for writing, thus waiting for all the systems to finish. Then we'd record the queued changes into the generation changes, and clear the queue.

User is supposed to call `rest` at the end of the frame, where waiting is allowed. It's an open question of how to enforce that. Perhaps, we can assert or warn when the size of the queue reaches a certain fraction of the world.

kvark:
Closes #11

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -146,9 +146,33 @@ impl World {
 }
 
 
+struct Appendix {
+    next: Entity,
+    add_queue: Vec<Entity>,
+    sub_queue: Vec<Entity>
+}
+
+/// A custom entity iterator for dynamically added entities.
+pub struct ExtraEntityIter<'a> {
+    guard: RwLockReadGuard<'a, Appendix>,
+    index: usize,
+}
+
+impl<'a> Iterator for ExtraEntityIter<'a> {
+    type Item = Entity;
+    fn next(&mut self) -> Option<Entity> {
+        let ent = self.guard.add_queue.get(self.index);
+        self.index += 1;
+        ent.map(|e| *e)
+    }
+}
 
 /// World argument for a system closure.
-pub struct WorldArg(Arc<World>, RefCell<Option<Pulse>>);
+pub struct WorldArg {
+    world: Arc<World>,
+    pulse: RefCell<Option<Pulse>>,
+    app: Arc<RwLock<Appendix>>,
+}
 
 impl WorldArg {
     /// Borrows the world, allowing the system lock some components and get the entity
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -136,24 +136,80 @@ impl World {
...
-        let pulse = self.1.borrow_mut().take().expect("fetch may only be called once.");
-        let u = f(&self.0);
+        let pulse = self.pulse.borrow_mut().take()
+                        .expect("fetch may only be called once.");
+        let u = f(&self.world);
         pulse.pulse();
         u
     }
+    /// Insert a new entity dynamically.
+    pub fn insert(&self) -> Entity {
```
csherratt: Maybe we should just go with `create`?

kvark: agreed, I'm revamping the change now

PR #17, opened 2016-04-10
actully wait

csherratt:
Use a barrier to stall the scheduler when wait is called.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -13,7 +13,7 @@ extern crate fnv;
 
 use std::cell::RefCell;
 use std::sync::Arc;
-use pulse::{Pulse, Signal};
+use pulse::{Pulse, Signal, Barrier, Signals};
 use threadpool::ThreadPool;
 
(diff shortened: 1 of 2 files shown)


kvark:
thanks!

kvark:
Just realized the actual wait needs to go before the `merge`, not after. That's not going to change much though, just a cosmetic aspect.

csherratt:
@kvark I just discovered that doing the merge before the barrier can deadlock the scheduler. Wonderful, I'll submit a fix with #19

kvark:
Hmm, it's the deadlock that I fixed. Are you sure your were testing with my #15 ?

csherratt:
Yep, was testing master.

```rust
#[test]
fn dynamic_create() {
    let mut scheduler = create_world();

    for _ in 0..1_000 {
        scheduler.run(|arg| {
            arg.fetch(|_| ());
            arg.create();
        });
        scheduler.wait();
    }
}
```

This use to reproduce the deadlock, but actually fails right now since there is another bug I'm isolating.

PR #20, opened 2016-04-10
Fix entity reuse using dynamic create.

csherratt:
Using the length of generations is not a reliable source of the
next entity as it is out of date when using dynamic entity creation

### DIFF
--- a/src/world.rs
+++ b/src/world.rs
@@ -55,13 +55,18 @@ impl<'a> EntityBuilder<'a> {
 struct Appendix {
     next: Entity,
     add_queue: Vec<Entity>,
-    sub_queue: Vec<Entity>
+    sub_queue: Vec<Entity>,
 }
 
-fn find_next(gens: &[Generation], base: usize) -> Entity {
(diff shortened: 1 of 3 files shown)


kvark:
@csherratt please clarify the cast that you found.

csherratt:
`dynamic_create` was the test that reproduced the issue.

src/world.rs
```diff
@@ -55,13 +55,20 @@ impl<'a> EntityBuilder<'a> {
...
+    max_eid: Index
 }
 
-fn find_next(gens: &[Generation], base: usize) -> Entity {
+fn find_next(gens: &[Generation], base: usize, max_eid: &mut Index) -> Entity {
     match gens.iter().enumerate().skip(base).find(|&(_, g)| *g <= 0) {
-        Some((id, gen)) => Entity(id as Index, 1 - gen),
-        None => Entity(gens.len() as Index, 1),
+        Some((id, gen)) => {
+            println!("Some({}, {})", id, gen);
```
csherratt: oops

PR #21, opened 2016-04-11
Added comments to example

White-Oak:
Also added an error message for unregistered types and fixed several clippy warnings like elided lifetimes, I hope noone is against that.


kvark:
Very nice, thank you!

PR #28, opened 2016-04-13
Hide implementation of Generation

futile:
Instead, expose `is_alive()` and `is_dead()` methods which have a clear
name, and refactor existing code to use these.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -30,14 +30,26 @@ mod world;
 pub struct Generation(i32);
 
 impl Generation {
-    /// Returns `true` if entities of this generation are alive
+    /// Returns `true` if entities of this `Generation` are alive
     pub fn is_alive(&self) -> bool {
         self.0 > 0
     }
 
-    /// Returns `true` if entities of this generation are dead
-    pub fn is_dead(&self) -> bool {
-        !self.is_alive()
+    /// Kills this `Generation`
+    fn die(&mut self) {
+        debug_assert!(self.is_alive());
+        self.0 = -self.0;
+    }
+
+    /// Revive and increment a dead `Generation`
+    fn raised(self) -> Generation {
+        debug_assert!(!self.is_alive());
+        Generation(1 - self.0)
+    }
+
+    /// Returns `true` if this is a first generation, i.e. has value `1`
+    fn is_first(&self) -> bool {
+        self.0 == 1
     }
 }
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -23,13 +23,24 @@ pub use world::{Component, World, FetchArg,
 mod storage;
 mod world;
 
-
 /// Index generation. When a new entity is placed at the old index,
 /// it bumps the generation by 1. This allows to avoid using components
 /// from the entities that were deleted.
-/// G<=0 - the entity of generation G is dead
-/// G >0 - the entity of generation G is alive
-pub type Generation = i32;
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -23,13 +23,24 @@ pub use world::{Component, World, FetchArg,
...
+pub struct Generation(i32);
+
+impl Generation {
+    /// Returns `true` if entities of this generation are alive
+    pub fn is_alive(&self) -> bool {
+        self.0 > 0
+    }
+
+    /// Returns `true` if entities of this generation are dead
+    pub fn is_dead(&self) -> bool {
```
kvark: I'd rather not have a separate method here, `!x.is_alive()` is clear enough

src/world.rs
```diff
@@ -59,14 +59,14 @@ struct Appendix {
 }
 
 fn find_next(gens: &[Generation], lowest_free_index: usize) -> Entity {
-    if let Some((id, gen)) = gens.iter().enumerate().skip(lowest_free_index).find(|&(_, g)| *g <= 0) {
-        return Entity(id as Index, 1 - gen);
+    if let Some((id, gen)) = gens.iter().enumerate().skip(lowest_free_index).find(|&(_, g)| g.is_dead()) {
+        return Entity(id as Index, Generation(1 - gen.0));
```
kvark: we better have a method here, like `gen.die()`

src/world.rs
```diff
@@ -59,14 +59,14 @@ struct Appendix {
...
     }
 
     if lowest_free_index > gens.len() {
-        Entity(lowest_free_index as Index, 1)
+        Entity(lowest_free_index as Index, Generation(1))
```
kvark: Please refactor this to only return the index id from the condition, and then return the `Entity` in a separate line

src/world.rs
```diff
@@ -219,12 +219,12 @@ impl World {
         }
         let mut gens = self.generations.write().unwrap();
         let mut gen = &mut gens[entity.get_id() as usize];
-        assert!(*gen > 0);
+        assert!(gen.is_alive());
         let mut app = self.appendix.write().unwrap();
         if entity.get_id() < app.next.get_id() {
-            app.next = Entity(entity.0, *gen+1);
+            app.next = Entity(entity.0, Generation(gen.0 + 1));
```
kvark: This was rather a hack - to do `gen.0+1` before the `gen` is inverted. Instead, we should do:
```rust
gen = gen.die();
...
app.next = Entity(..., gen.raise());
```

src/world.rs
```diff
@@ -253,19 +253,19 @@ impl World {
...
+                assert_eq!(ent.get_gen().0, 1 - gens[ent.get_id()].0);
                 gens[ent.get_id()] = ent.get_gen();
             }
             let mut next = app.next;
             for ent in app.sub_queue.drain(..) {
                 assert_eq!(ent.get_gen(), gens[ent.get_id()]);
                 temp_list.push(ent);
                 if ent.get_id() < next.get_id() {
-                    next = Entity(ent.0, ent.1 + 1);
+                    next = Entity(ent.0, Generation(ent.get_gen().0 + 1));
```
kvark: same here (as my previous comment)

src/world.rs
futile: `gen.raise()` should basically return `Generation(gen.0 + 1)`?

kvark:
Thanks a bunch!

Issue #24, opened 2016-04-11
Avoid iterating through all the entities


kvark:
In a lot of cases, the fraction of entities that do need to be processed is just a fraction of all entities. We lose a lot of time by checking all the entity components to exist. This information could be cached and re-used.

We could get a small speedup (in general) by early exiting if one of the components is not found. Similarly, we could advice users to convert problematic `runXwYr` calls to custom `run` calls and early-exit manually.

As a last resort, one could separate the ECS into two, so that the component coverage is denser in each (like, don't mix particle entities with UI entities with game entities, etc).

One way to properly address this is to use [tables](https://github.com/csherratt/tables). @csherratt reported the speedup from ~260us to ~50us by building on top of our `Storage` interface.

Another way would be to have `Aspect` as a special construct that tracks which entities have a required set of components, proposed by @OvermindDL1.

OvermindDL1:
For note, my 'Aspects' only defined where they looked in the Aspect cache. If someone in my system creates an Aspect.match<HealthComponent, DamageComponent>, Aspect.optional<ArmorComponent>>() it just returns a single pointer into the aspect cache, the cache itself was a recursive map of maps with keys of component atoms/ids and values of the matched ones with sub-matches with other components requested. You 'should' create any Aspects as early as possible as the first time a certain matchspec is defined it will iterate over all entities to see which go in, but from then on that cache is kept up to date among all aspects that match on the same matchspec.
The old system I based the style off of (Artemis, don't bother looking at it, this was the only good part about its system) used each Aspect as its own cache and they just registered event handlers to hear when components were added/removed, which caused more indirect function calls compared to mine
Plus duplicated data

OvermindDL1:
I just looked in the Tables library a bit more (I still have so much more to learn about rust syntax, I have the ideas, but just the syntax...  >.>), looks like it would work very well as a backing store while also using Aspects, they are not competing concepts but rather seem to attack two different issues in their own efficient ways.  Perhaps they should both be integrated.

kvark:
Fixed by #29

PR #29, opened 2016-04-14
Remove the need for iteration over entities

### LINKED ISSUE #13: Optimize VecStorage
It's currently implemented as `Vec<Option<T>>`, thus blowing up the size by a machine word, with only one bit really used. This reduces cache efficiency. Instead, we can have a separate bit vector.

csherratt:
Resolves #13, #24 

This uses the Hierarchical BitSet mask anding that I have been
experimenting with to allow fast iteration over the key space.

When `joining` components we only visit the entity rows are shared
between the both components.

 Benchmark | Before | After
--- | --- | ---
parallel build | 428,938 ns/iter (+/- 4,011) | 429,104 ns/iter (+/- 5,833)
parallel update | 191,447 ns/iter (+/- 54,436) | 70,627 ns/iter (+/- 18,932)
pos_vel build | 239,626 ns/iter (+/- 2,882) | 276,367 ns/iter (+/- 13,210)
pos_vel build | 257,680 ns/iter (+/- 99,414) | 49,947 ns/iter (+/- 9,426)


src/lib.rs
```diff
@@ -126,14 +133,14 @@ impl Planner {
         }
     }
     /// Run a custom system.
-    pub fn run<F>(&mut self, functor: F) where
-        F: 'static + Send + FnOnce(RunArg)
+    pub fn run<'a, F>(&mut self, functor: F) where
```
csherratt: this looks like a merge fail, will fix

src/join.rs
```diff
@@ -0,0 +1,288 @@
+use tuple_utils::Split;
+use {BitSetAnd, BitSetLike};
+
+/// Join is a helper method to & bitsets togather resulting in a tree
+pub trait Join {
+    type Value;
+    fn join(self) -> Self::Value;
+}
+
+impl<A> Join for (A,)
```
kvark: These implementations beg to be generated by a macro

src/bitset.rs
```diff
@@ -0,0 +1,447 @@
...
+        if self.layer0[p0] & id.mask::<Shift0>() == 0 {
+            return false;
+        }
+
+        // if the bitmask was set we need to clear
+        // its bit from layer0 to 3. the layers abover only
+        // should be cleared if the bit cleared was the last bit
+        // in its set
+        self.layer0[p0] &= !id.mask::<Shift0>();
+        if self.layer0[p0] != 0 {
```
kvark: Since you've already removed this bit, how do you know if it was there in the first place (in order to `return true`)?

src/bitset.rs
```diff
@@ -0,0 +1,447 @@
...
+    #[inline(always)]
+    fn offset<S>(self) -> usize
+        where S: Shift
+    {
+        self as usize / (1 << S::shift())
+    }
+}
+
+#[derive(Copy, Clone)]
+pub struct Shift0;
```
kvark: Can these be just constant values? Why do you need `trait Shift`?

src/bitset.rs
```diff
@@ -0,0 +1,447 @@
...
+    pub fn contains(&self, bit: u32) -> bool {
+        let p0 = bit.offset::<Shift1>();
+        if p0 >= self.layer0.len() {
+            return false;
+        }
+        (self.layer0[p0] & bit.mask::<Shift0>()) != 0
+    }
+
+    /// Create an iterator over the the keyspace
+    pub fn iter<'a>(&'a self) -> Iter<'a, Self> {
```
kvark: I assume this is redundant since `BitSetLike` carries the implementation anyway

src/lib.rs
```diff
@@ -10,18 +10,23 @@ extern crate mopa;
 extern crate pulse;
 extern crate threadpool;
 extern crate fnv;
+extern crate tuple_utils;
```
kvark: What is it used for?

src/storage.rs
```diff
@@ -14,99 +16,333 @@ pub trait StorageBase {
...
     fn get(&self, Entity) -> Option<&T>;
     /// Try mutating the data associated with an entity.
     fn get_mut(&mut self, Entity) -> Option<&mut T>;
-    /// Insert a new data for a given entity.
-    fn insert(&mut self, Entity, T);
     /// Remove the data associated with an entity.
     fn remove(&mut self, Entity) -> Option<T>;
+    /// splits the bitset from the storage for use
+    /// by the join iterator.
+    fn entities(&self) -> (&BitSet, &Self::UnprotectedStorage);
```
kvark: Perhaps, we could find a better name for this method. `entities` here returns neither entities container, or an iterator.
Perhaps `open` and `open_mut`?

src/storage.rs
```diff
@@ -14,99 +16,333 @@ pub trait StorageBase {
...
+/// performance for the components mostly present in entities.
+pub struct VecStorage<T> {
+    set: BitSet,
+    values: InnerVec<T>,
+}
+
+impl<T> VecStorage<T> {
+    fn extend(&mut self, id: usize) {
+        unsafe {
+            self.values.0.reserve(id + 1);
```
kvark: `reserve` is used incorrectly here. See:
> Reserves capacity for at least additional more elements to be inserted in the given Vec<T>

src/storage.rs
```diff
@@ -14,99 +16,333 @@ pub trait StorageBase {
...
+        (&self.set, &self.values)
+    }
+    fn entities_mut(&mut self) -> (&BitSet, &mut InnerVec<T>) {
+        (&self.set, &mut self.values)
+    }
+}
+
+impl<T> super::UnprotectedStorage<T> for InnerVec<T> {
+    unsafe fn get(&self, e: u32) -> &T {
+        &self.0[e as usize].data
```
kvark: I suggest to use [get_unchecked](http://doc.rust-lang.org/nightly/std/primitive.slice.html#method.get_unchecked) here

src/storage.rs
```diff
@@ -14,99 +16,333 @@ pub trait StorageBase {
...
+    }
+    unsafe fn get_mut(&mut self, e: u32) -> &mut T {
+        &mut self.0[e as usize].data
```
kvark: and [get_unchecked_mut](http://doc.rust-lang.org/nightly/std/primitive.slice.html#method.get_unchecked_mut) here

kvark:
Ok, I reviewed the code and made some notes and suggestions. Very impressive work, as always, @csherratt , thank you!

My biggest concern is the complication of the `Storage` type, with the need to not only implement `UnsafeStorage` now, but also to expose the `BitSet`. For some storage types (like the hashmap that we already have), this is a little redundant, since they already know what they have and what not.
Another thing is the general increase in the code complexity... However, the following factors outweigh the concerns:
  - Storages are not implemented that often. Most users will be happy with built-in hashmaps and vectors. A few others can tolerate the increased complexity (when implementing custom storages) since they probably need high performance for their complex cases.
  - The performance win is tremendous

I suppose we should release v0.5 when this code gets cleaned up, merged, and tested. I'm sure some people may still find v0.4 appealing for its minimalistic design, but we'll move forward.

src/bitset.rs
csherratt: There is a check on Line 120-122 that checks to see if the bit was missing or not. At this point in the logic we know the bit was set.

src/lib.rs
csherratt: the `Split` trait used to implement `Join`

src/storage.rs
csherratt: O_O ooops.

src/join.rs
```diff
@@ -16,273 +17,36 @@ impl<A> Join for (A,)
...
+merge_impl!{A, B, C, D, E, F, G}
+merge_impl!{A, B, C, D, E, F, G, H}
+merge_impl!{A, B, C, D, E, F, G, H, I}
+merge_impl!{A, B, C, D, E, F, G, H, I, J}
+merge_impl!{A, B, C, D, E, F, G, H, I, J, K}
+merge_impl!{A, B, C, D, E, F, G, H, I, J, K, L}
+merge_impl!{A, B, C, D, E, F, G, H, I, J, K, L, M}
+merge_impl!{A, B, C, D, E, F, G, H, I, J, K, L, M, N}
+merge_impl!{A, B, C, D, E, F, G, H, I, J, K, L, M, N, O}
+merge_impl!{A, B, C, D, E, F, G, H, I, J, K, L, M, N, O, P}
```
kvark: muuuch better, thanks!

src/storage.rs
```diff
@@ -147,7 +147,8 @@ pub struct VecStorage<T> {
 impl<T> VecStorage<T> {
     fn extend(&mut self, id: usize) {
         unsafe {
-            self.values.0.reserve(id + 1);
+            let delta = (id + 1) - self.values.0.len();
```
kvark: should we `debug_assert!` if the `id` is already in range?

src/bitset.rs
```diff
@@ -75,16 +80,16 @@ impl BitSet {
     /// when the lowest layer was set from 0
     fn add_slow(&mut self, id: Index) {
         let (_, p1, p2) = offsets(id);
-        self.layer1[p1] |= id.mask::<Shift1>();
-        self.layer2[p2] |= id.mask::<Shift2>();
-        self.layer3 |= id.mask::<Shift3>();
+        self.layer1[p1] |= id.mask(SHIFT1);
```
kvark: much nicer, thanks!

src/storage.rs
```diff
@@ -14,99 +16,334 @@ pub trait StorageBase {
...
+/// performance for the components mostly present in entities.
+pub struct VecStorage<T> {
+    set: BitSet,
+    values: InnerVec<T>,
+}
+
+impl<T> VecStorage<T> {
+    fn extend(&mut self, id: usize) {
+        unsafe {
+            let delta = (id + 1) - self.values.0.len();
```
kvark: also, the `let delta` line as well as the `debug_assert!` should be moved up, outside of `unsafe{}`

kvark:
9..8..7..6..5..4..3..2..1.. MERGING
> Nuclear launch detected!

White-Oak:
Nice

Issue #34, opened 2016-04-15
Add is_alive(Entity) method to the World


kvark:
Could be useful, and should be trivial to implement.

futile:
Working on this now.

kvark:
Fixed by #35

PR #35, opened 2016-04-15
Add `World::is_alive(Entity)` method

futile:
Also fix a bug in `create_now()`, which would not update the stored
`Generation` in some cases.

Implements #34


kvark:
Impressive, thank you!

PR #36, opened 2016-04-15
Cleanup documenation at many places.

futile:
Also turns all comments of the form `Return value x` into the form
`Returns value x`, which is in line with official rust docs.


csherratt:
Thanks!

PR #38, opened 2016-04-16
replace two asserts with assert_eq!s

futile:
`assert_eq!` also prints the values of the two sides in case of panic,
so `assert_eq!(a, b)` should be preferred over `assert!(a == b)`.


futile:
these were all the instances of `assert!(a == b)` I found, btw.

PR #39, opened 2016-04-17
Add a `Join` trait

csherratt:
This is a follow-up to my changes that added the bitset masks. This gives a way for api users to do the bitset join in a much more friendly way then we were doing before.

Take a look at the updated example to see how it works.

### DIFF
--- a/examples/basic.rs
+++ b/examples/basic.rs
@@ -68,7 +68,7 @@ fn main() {
         // Instead of using the `entities` array you can
         // use the `Join` trait that is an optimized way of
         // doing the `get/get_mut` across entities.
-        for (a, b) in (&mut *sa, &*sb).join() {
+        for (a, b) in (&mut sa, &sb).join() {
             a.0 += if b.0 {1} else {0};
         }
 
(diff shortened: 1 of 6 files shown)


examples/basic.rs
```diff
@@ -61,11 +61,18 @@ fn main() {
...
-                a.0 = if b.0 {2} else {0};
+                a.0 = if b.0 {1} else {0};
             }
         }
 
-        // Dynamically creating and deleting entites
+        // Instead of using the `entities` array you can
+        // use the `Join` trait that is an optimized way of
+        // doing the `get/get_mut` across entities.
+        for (a, b) in (&mut *sa, &*sb).join() {
```
kvark: Looks good! I wonder if we can get rid of the `*` too

src/join.rs
```diff
@@ -1,52 +1,224 @@
...
+pub trait Open {
+    type Value;
+    type Mask: BitSetLike;
+    fn open(self) -> (Self::Mask, Self::Value);
+}
+
+impl<'a, T> Open for &'a T
+    where T: Storage
+{
+    type Value = GetRef<'a, <T as Storage>::UnprotectedStorage>;
```
kvark: Do you actually need `<T as Storage>` if `Storage` is the only constraint of `T`?

kvark:
I'm concerned about the increased complexity. Lots of traits and associated types...
I hope we'll be able to simplify this.
Given that you are still working on it, please merge when feeling ready.

PR #41, opened 2016-04-19
Added the System trait

### LINKED ISSUE #30: System data problem
Anything more than a simple demo will probably have some soft of a `System` trait, with objects implementing it being boxed and stored in a list to be executed automatically. It may look like this:
```rust
trait System {
    fn process(&mut self, &mut secs::Planner, TimeDelta);
}
```
Unfortunately, using `secs` under such abstraction is not elegant:
  1. The system execution is now split into 3 different chunks, executed at different moments:
      - `process()` entry, executed upon queuing the system
      - `fetch` phase, executed at the start of the system on another thread
      - `iter` phase, executed after all the storage locks are acquired
  2. Any data that system has (and wants to access in the execution closure) needs to be `Clone + Sync`. Imaging a mutable vector there - now you'd need to wrap it into `Arc<Mutex<X>>` , even though it should not be needed. This is inconvenient and bloats the code.

Optimally, I'd like to base ECS onto the system interface like this (edit: moved to #40):
```rust
trait System<C: Clone>: Send { //`C` stands for "Context"
    type Data<'a>; // some sort of HKT, which is not supported in current Rust
    fn fetch<'a>(&self, &'a secs::World) -> Self::Data<'a>;
    fn run(&mut self, Data<'a>, C);
}
```
Why this would be nice:
- this system is required to implement `Send`: each time it needs to be executed, it's moved onto the thread, run, and then moved back (or not, if we can track this).
- it allows any sort of data to be kept within the system without harsh restrictions or bloated wrappers
- it clearly separates executable parts and their contexts. This enforces a `fetch` at compile time.
- it allows the users to start using it right away (without higher abstractions) in complex systems, one just need to decide what `C` is (in the original example, `C = TimeDelta`)

kvark:
Closes #30 

The new code somehow introduces a deadlock in the tests. AFAIK, all I did was changing a pulse/barrier into a channel.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "specs"
-version = "0.5.2"
+version = "0.6.0"
 description = """
 Parallel Entity-Component System.
 
--- a/examples/basic.rs
+++ b/examples/basic.rs
@@ -14,6 +14,7 @@ impl specs::Component for CompBool {
     type Storage = specs::HashMapStorage<CompBool>;
 }
 
+
 fn main() {
     let (e, mut planner) = {
         let mut w = specs::World::new();
--- a/examples/basic.rs
+++ b/examples/basic.rs
@@ -28,7 +28,7 @@ fn main() {
         w.create_now().with(CompInt(127)).build();
         w.create_now().with(CompBool(false)).build();
         // Planner is used to run systems on the specified world with a specified number of threads
-        (e, specs::Planner::new(w, 4))
+        (e, specs::Planner::<()>::new(w, 4))
     };
 
     // Planner only runs closure on entites with specified components, for example:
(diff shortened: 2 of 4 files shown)


kvark:
Weird, if I move the same test code into the example, it works.
@csherratt do you see anything wrong here?

csherratt:
The test case that is failing is `test_panics`. I probably should have used something like [timebomb](https://crates.io/crates/timebomb) to check for tests like this.
The test case is checking to see if the task panicking propagates to the dispatch thread. With `pulse` a pulse would be dropped forcing it into an error state. This would propagate to anything waiting on the pulse.
My guess is this is not the behavior of `mpsc`, it probably only enters an error state if all produces have been hung up. Which can't happen since you hold your own sender side.

kvark:
@csherratt thanks for figuring this out! So what can be done to fix it nicely then?

csherratt:
Easiest way, add a drop guard to do the send in the case of a task panicking. This is the same technique the `threadpool` is doing internally. Well they spawn a thread, you would transmit the message.

https://github.com/frewsxcv/rust-threadpool/blob/master/src/lib.rs#L29-L60

kvark:
@csherratt the more I look at `threadpool` the more I realize that we are essentially building on top of it without actually hooking up to the same internal mechanisms. We'd benefit from tighter integration. so perhaps we should just hand-craft something similar manually.

csherratt:
@kvark I don't disagree on the `threadpool` Their implementation is pretty much wait on a mutex for work to arrive. It's simple, but obviously sub-optimal.

PR #43, opened 2016-04-21
Major Storage refactor

### LINKED ISSUE #32: Does Storage need to hold the generation?
Right now both storage types hold a generation with the data. This is used to validate if the value in the `get/get_mut` is the same one being referenced.I think the check is needed, the question is do we need to make every storage track this state?

Currently, entities are only created by systems using the dynamic create functionality. Deletes are done lazily via the dynamic delete. Because the deletes are lazy the `entity index` is never reused twice between a `planner.wait()`. And since the deletes are done by the `specs` itself, we are guaranteed that the value will be removed during a `.wait()`.

Because of this design I think it is reasonable to assume
1) no `Storage` will actually be storing a value that is out of date with the generation
2) `specs` only needs a single place that holds the `latest` generation.

What I purpose, is we remove the generation from `Storage` and return every `Storage` wrapped in a magic type that joins the shared generation count with the type. This will implement `get` ect but will validate that the object is both alive and that the generation is correct. The storage trait would be simplified a bit by referring to each `get/insert/remove` by Index rather then by Entity.

This will reduce the size of the components in the `VecStorage` `HashMapStorage` which should mean they will work faster since it is less cache pressure.

kvark:
Fixes #32 

Major changes:
- `BitSet` is exposed directly
- `VecStorage` and `HashMapStorage` implementations are much simpler, all the shared logic is moved up to the `MaskedStorage` and `Storage` types
- `Storage` is no longer a trait but a struct, which allows using its methods without any `use` items
- Storages no longer track the generations internally, but instead use the shared generations from the `World`
- `join` module is cleaned up by half

Performance:

| Benchmark  | specs-0.5.1 | specs-master | specs-pr |
| --------------- | --------------- | ----------------- | ----------- |
| pos_vel build | 320 us/iter (+/- 8) | 2,013 us/iter (+/- 139) |1,940 us/iter (+/- 75) |
| pos_vel update | 42 us/iter (+/- 8) | 42 us/iter (+/- 9) |42 us/iter (+/- 11) |
| parallel build | 472 us/iter (+/- 35) | 2,195 us/iter (+/- 89) |2,081 us/iter (+/- 101) |
| parallel update | 66 us/iter (+/- 11) | 64 us/iter (+/- 16) |64 us/iter (+/- 15) |

### DIFF
--- a/examples/basic.rs
+++ b/examples/basic.rs
@@ -83,17 +83,16 @@ fn main() {
         println!("Entity {} {}", a.0, b.0);
     });
     planner.run_custom(|arg| {
-        use specs::{Storage, Join};
         let (mut sa, sb, entities) = arg.fetch(|w| {
             (w.write::<CompFloat>(), w.read::<CompInt>(), w.entities())
         });
 
         // Insert a component for each entity in sb
-        for (eid, sb) in (&entities, &sb).join() {
+        for (eid, sb) in (&entities, &sb).iter() {
             sa.insert(eid, CompFloat(sb.0 as f32));
         }
 
-        for (eid, sa, sb) in (&entities, &mut sa, &sb).join() {
+        for (eid, sa, sb) in (&entities, &mut sa, &sb).iter() {
             assert_eq!(sa.0 as u32, sb.0 as u32);
             println!("eid[{:?}] = {:?} {:?}", eid, sa, sb);
         }
--- a/examples/basic.rs
+++ b/examples/basic.rs
@@ -1,4 +1,5 @@
 extern crate specs;
+use specs::Join;
 
 #[derive(Clone, Debug)]
 struct CompInt(i8);
@@ -66,7 +67,7 @@ fn main() {
         // Instead of using the `entities` array you can
         // use the `Join` trait that is an optimized way of
         // doing the `get/get_mut` across entities.
-        for (a, b) in specs::Join::from((&mut sa, &sb)) {
+        for (a, b) in (&mut sa, &sb).iter() {
             a.0 += if b.0 {2} else {0};
         }
 
--- a/examples/basic.rs
+++ b/examples/basic.rs
@@ -57,7 +57,6 @@ fn main() {
 
     // Instead of using macros you can use run_custom() to build a system precisely
     planner.run_custom(|arg| {
-        use specs::{Storage, Join};
         // fetch() borrows a world, so a system could lock neccessary storages
         // Can be called only once
         let (mut sa, sb) = arg.fetch(|w| {
@@ -67,7 +66,7 @@ fn main() {
         // Instead of using the `entities` array you can
         // use the `Join` trait that is an optimized way of
         // doing the `get/get_mut` across entities.
-        for (a, b) in (&mut sa, &sb).join() {
+        for (a, b) in specs::Join::from((&mut sa, &sb)) {
             a.0 += if b.0 {2} else {0};
         }
 
--- a/src/join.rs
+++ b/src/join.rs
@@ -1,9 +1,7 @@
 use std;
-use std::sync::{RwLockReadGuard, RwLockWriteGuard};
 use tuple_utils::Split;
-use bitset::{BitIter, BitSet, BitSetAnd, BitSetLike};
-use storage::{Storage, MaskedStorage, UnprotectedStorage};
-use {Index, Generation, Component};
+use bitset::{BitIter, BitSetAnd, BitSetLike};
+use Index;
 
 
 /// BitAnd is a helper method to & bitsets togather resulting in a tree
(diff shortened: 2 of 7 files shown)


csherratt:
This looks promising.

src/join.rs
```diff
@@ -144,51 +134,33 @@ impl<'a, T> Get for GetMut<'a, T>
...
-        if let Some(idx) = self.keys.next() {
-            unsafe { Some(self.values.get(idx)) }
-        } else {
-            None
-        }
+impl<O: Open> std::iter::Iterator for Joined<O> {
+    type Item = O::Type;
+    fn next(&mut self) -> Option<O::Type> {
+        self.keys.next().map(|idx| unsafe {
+            O::get(self.values, idx)
```
kvark: I believe I have to do the mutable transmute there, similarly to the original code...

csherratt: Yep. It's not possible to give multiple mutable access to anything in rust without a transmute of some type. The `&mut T` iterators for the &mut [T].

Everything is cheating to do this, `std::slice::IterMut` used this macro https://github.com/rust-lang/rust/blob/master/src/libcore/slice.rs#L186

The vec_Dequeue https://github.com/rust-lang/rust/blob/master/src/libcollections/vec_deque.rs#L1832

kvark:
Getting closer now. Everything compiles but half of the tests are failing.
Also, big merge ahead.

src/storage.rs
```diff
@@ -1,433 +1,480 @@
...
-        } else {
-            None
+    unsafe fn insert(&mut self, id: Index, v: T) {
+        let id = id as usize;
+        if self.0.len() <= id {
+            let delta = id + 1 - self.0.len();
+            self.0.reserve(delta);
+            self.0.set_len(id + 1);
         }
+        self.0[id] = v;
```
csherratt: You need to be careful here. if `T` supports drop this will drop the element before overwriting it, so it would drop an uninitialized value. Which is why I was using `ptr::write` before. of course if you use `ptr::write` and there was a value was suppose to be dropped you have now added a memory leak. If the value was set before you need to either use `mem::swap` to exchange it, or `ptr::read` and `ptr::write` to exchange it.

src/join.rs
```diff
@@ -54,141 +54,48 @@ bitset_and!{A, B, C, D, E, F, G, H, I, J, K, L, M, N}
...
-        } else {
-            None
-        }
+impl<O: Open> std::iter::Iterator for Join<O> {
+    type Item = O::Type;
+    fn next(&mut self) -> Option<O::Type> {
+        self.keys.next().map(|idx| unsafe {
+            // We only transmute and copy during iteration, which is safe and serves
+            // as a poor man's replacement for the missing re-borrowing semantic.
+            let values: O::Value = std::mem::transmute_copy(&self.values);
```
csherratt: This makes me nervous. I think it would be better to avoid the transmute here, and push it into the `Open` trait. While this is fine for our types, I think it will be a problem if anyone's `O::Value` is something other then a borrow.

kvark: Agreed. I was able to move this transmute much lower level, leaving the trait itself safe. Pushing in a moment.

kvark:
Note: currently, `Join` is built on top of the `Storage` types returned by the `World`. This means that we have multiple read locks of the G-vec (generations vector), each of which is checked for every component access.

Optimally, the join iterator should only have a single G-vec lock, and check it once of each entity. This should be possible to achieve by integrating the `Join` macro implementation with the `Storage` internals. This should speed it up a little.

TODO: move into a separate issue

examples/basic.rs
```diff
@@ -83,17 +83,16 @@ fn main() {
...
     });
     planner.run_custom(|arg| {
-        use specs::{Storage, Join};
         let (mut sa, sb, entities) = arg.fetch(|w| {
             (w.write::<CompFloat>(), w.read::<CompInt>(), w.entities())
         });
 
         // Insert a component for each entity in sb
-        for (eid, sb) in (&entities, &sb).join() {
+        for (eid, sb) in (&entities, &sb).iter() {
```
csherratt: ❤️

src/join.rs
```diff
@@ -31,8 +31,8 @@ macro_rules! bitset_and {
                 <<Self as Split>::Right as BitAnd>::Value
             >;
             fn and(self) -> Self::Value {
-              let (l, r) = self.split();
-              BitSetAnd(l.and(), r.and())
+                let (l, r) = self.split();
```
csherratt: Oops. I must have misconfigured my editor.

src/world.rs
```diff
@@ -174,17 +174,21 @@ impl<'a> Iterator for CreateEntities<'a> {
...
 mopafy!(StorageLock);
 
-impl<S: StorageBase + Any + Send + Sync> StorageLock for RwLock<S> {
+impl<T: Component> StorageLock for RwLock<MaskedStorage<T>> {
     fn del_slice(&self, entities: &[Entity]) {
+        use storage::PrivateStorage;
         let mut guard = self.write().unwrap();
         for &e in entities.iter() {
-            guard.del(e);
+            if guard.get_mask().contains(e.get_id()) {
```
csherratt: This is why the example is not working. This should be `remove()`, as right now this will leave the mask with a set bit that has been deleted.

csherratt:
LGTM. Is anything pending?

kvark:
@csherratt Have a look at the benchmark table I added. Other than that, this PR is ready to go in. You have the power to hit that green button, I'll much appreciate that ;)

src/world.rs
kvark: Genius, thank you!

PR #44, opened 2016-04-21
[WIP] Refactor how allocation/deletion works

csherratt:
- The DynamicEntityIter is removed
- Allocation / Deletion can be done without a write lock
- Add the ability to `Join` over the `entities()`

This removes the `add_list/sub_list` in-favour of atomic bit sets. Killing an Entity sets a bit atomically that can be looked up later during the merge. Raising a dead entity is done via an atomic bitset too. This allows systems to iterate over the entities without blocking system that need to create or delete entities.

The other big part of this change is that the functionality of the DynamicEntityIter & the EntitiyIter are now in one place. We merge the known alive entity ids, with any raised entities. The `Entities` can see entities that are created by systems in parallel, which might cause some unpredictable behaviour. But if you are masking with a component (that is locked upstream) it still reliably work.

### DIFF
--- a/examples/basic.rs
+++ b/examples/basic.rs
@@ -7,6 +7,7 @@ impl specs::Component for CompInt {
     // VecStorage is meant to be used for components that are in almost every entity
     type Storage = specs::VecStorage<CompInt>;
 }
+
 #[derive(Clone, Debug)]
 struct CompBool(bool);
 impl specs::Component for CompBool {
@@ -14,6 +15,12 @@ impl specs::Component for CompBool {
     type Storage = specs::HashMapStorage<CompBool>;
 }
 
+#[derive(Clone, Debug)]
+struct CompFloat(f32);
+impl specs::Component for CompFloat {
+    // HashMapStorage is better for componets that are met rarely
+    type Storage = specs::VecStorage<CompFloat>;
+}
 
 fn main() {
     let (e, mut planner) = {
@@ -21,6 +28,7 @@ fn main() {
         // All components types should be registered before working with them
         w.register::<CompInt>();
         w.register::<CompBool>();
+        w.register::<CompFloat>();
         // create_now() of World provides with an EntityBuilder to add components to an Entity
         w.create_now().with(CompInt(4)).with(CompBool(false)).build();
         // build() returns an entity, we will use it later to perform a deletion
@@ -53,24 +61,15 @@ fn main() {
         use specs::{Storage, Join};
         // fetch() borrows a world, so a system could lock neccessary storages
         // Can be called only once
-        let (mut sa, sb, entities) = arg.fetch(|w| {
-            (w.write::<CompInt>(),
-             w.read::<CompBool>(),
-             w.entities())
+        let (mut sa, sb) = arg.fetch(|w| {
+            (w.write::<CompInt>(), w.read::<CompBool>())
         });
-
-        for ent in entities {
-            // Will only run for entities with both components present
-            if let (Some(a), Some(b)) = (sa.get_mut(ent), sb.get(ent)) {
-                a.0 = if b.0 {1} else {0};
-            }
-        }
 
         // Instead of using the `entities` array you can
         // use the `Join` trait that is an optimized way of
         // doing the `get/get_mut` across entities.
         for (a, b) in (&mut sa, &sb).join() {
-            a.0 += if b.0 {1} else {0};
+            a.0 += if b.0 {2} else {0};
         }
 
         // Dynamically creating and deleting entities
@@ -84,14 +83,22 @@ fn main() {
     planner.run0w2r(|a: &CompInt, b: &CompBool| {
         println!("Entity {} {}", a.0, b.0);
     });
-    planner.wait();
-    if false {   // some debug output
-        let w = &planner.world;
-        //println!("Generations: {:?}", &*w.get_generations());
-        //println!("{:?}", &*w.read::<CompInt>());
-        //println!("{:?}", &*w.read::<CompBool>());
-        for e in w.entities() {
-            println!("{:?}", e);
+    planner.run_custom(|arg| {
+        use specs::{Storage, Join};
+        let (mut sa, sb, entities) = arg.fetch(|w| {
+            (w.write::<CompFloat>(), w.read::<CompInt>(), w.entities())
+        });
+
+        // Insert a component for each entity in sb
+        for (eid, sb) in (&entities, &sb).join() {
+            sa.insert(eid, CompFloat(sb.0 as f32));
         }
-    }
+
+        for (eid, sa, sb) in (&entities, &mut sa, &sb).join() {
+            assert_eq!(sa.0 as u32, sb.0 as u32);
+            println!("eid[{:?}] = {:?} {:?}", eid, sa, sb);
+        }
+    });
+
+    planner.wait();
 }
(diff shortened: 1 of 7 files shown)


examples/basic.rs
```diff
@@ -7,20 +7,28 @@ impl specs::Component for CompInt {
...
 impl specs::Component for CompBool {
     // HashMapStorage is better for componets that are met rarely
     type Storage = specs::HashMapStorage<CompBool>;
 }
 
+#[derive(Clone, Debug)]
+struct CompFloat(f32);
+impl specs::Component for CompFloat {
+    // HashMapStorage is better for componets that are met rarely
+    type Storage = specs::VecStorage<CompFloat>;
```
kvark: comment is wrong

src/bitset.rs
```diff
@@ -1,7 +1,11 @@
 use std::iter::repeat;
+use std::sync::atomic::{AtomicUsize, Ordering};
 
 use Index;
 
+#[cfg(target_pointer_width= "64")]
+pub const BITS: usize = 6;
+#[cfg(target_pointer_width= "32")]
```
kvark: oo, fancy :)

src/bitset.rs
```diff
@@ -20,10 +24,10 @@ pub const SHIFT3: usize = SHIFT2 + BITS;
 /// Adding beyond this limit will cause the `BitSet` to panic.
 #[derive(Clone)]
 pub struct BitSet {
-    layer3: u32,
-    layer2: Vec<u32>,
-    layer1: Vec<u32>,
-    layer0: Vec<u32>,
+    layer3: usize,
```
kvark: usize would make it easier for the storage implementations in my PR - no more casting to `u32`

kvark: I'm very interested to see if our performance is affected by this change.

src/bitset.rs
```diff
@@ -295,6 +312,161 @@ impl<T> Iterator for Iter<T>
...
+        }
+    }
+
+    /// Adds `id` to the `AtomicBitSet`. Returns `true` if the value was
+    /// already in the set.
+    ///
+    /// Because we cannot safely extend an AtomicBitSet without unique ownership
+    /// this will panic if the Index is out of range
+    #[inline]
+    pub fn add_atomic(&self, id: Index) -> bool {
```
kvark: does it have to be public?

src/bitset.rs
```diff
@@ -295,6 +312,161 @@ impl<T> Iterator for Iter<T>
...
+        }
+
+        self.add_atomic(id)
+    }
+
+    /// Removes `id` from the set, returns `true` if the value
+    /// was removed, and `false` if the value was not set
+    /// to begin with.
+    #[inline]
+    pub fn remove(&mut self, id: Index) -> bool {
```
kvark: why does it have to be `&mut self`?

src/bitset.rs
```diff
@@ -295,6 +312,161 @@ impl<T> Iterator for Iter<T>
...
+
+        self.layer3.fetch_and(!id.mask(SHIFT3), Ordering::Relaxed);
+        return true;
+    }
+
+    /// Returns `true` if `id` is in the set.
+    #[inline]
+    pub fn contains(&self, id: u32) -> bool {
+        let p0 = id.offset(SHIFT1);
+        if p0 >= self.layer0.len() {
```
kvark: could merge with the following condition for clarity, as in `p0 < self.layer0.len() && ...`

src/bitset.rs
```diff
@@ -295,6 +312,161 @@ impl<T> Iterator for Iter<T>
...
+            return false;
+        }
+        (self.layer0[p0].load(Ordering::Relaxed) & id.mask(SHIFT0)) != 0
+    }
+
+    /// Clear all bits in the set
+    pub fn clear(&mut self) {
```
kvark: again, why does it have to be `&mut self`?

src/bitset.rs
```diff
@@ -373,3 +545,80 @@ mod set_test {
     }
 }
 
+#[cfg(test)]
+mod atomic_set_test {
```
kvark: bonus points for the great test suite :+1:

src/bitset.rs
csherratt: It hurt a bit unfortunately, it was minor but measurable. `ecs_bench` went from 60->65us, no change on the build times.

src/bitset.rs
csherratt: So it is sound either atomically add or atomically delete, but not both. If it was a plane bitset it would be safe to do both and this could be `&self`.

But the hierarchy causes problems. The atomic add will set the layers above it, and the worst case is that two writers both set the same bit, which has no effect.

In the case of an atomic remove, whoever migrates the layer to 0 is required to clear the bit in the layer above. This is also sound, as long as the other writers are doing the same.

The issue is when these two are mixed. If Thread 1 sets a bit, then sets the bits in the layers above. And Thread 2 removes the same bit there is a race between the two threads that will cause the layers above to enter an inconsistent state.

So therefore, the only way I can make sure this is sound is to make sure the removes require unique ownership, which prevents the case when thread A is setting and Thread B is clearing.

src/bitset.rs
kvark: I was hoping that using larger bitfields would give better perf, but eh...

src/bitset.rs
kvark: I see, makes sense. Thanks for the explanation! Might be worth putting it in the comments.

src/world.rs
```diff
@@ -12,26 +15,34 @@ pub trait Component: Any + Sized {
...
+        (BitSetOr(&self.guard.alive, &self.guard.raised), self)
+    }
+}
+
+impl<'a> Get for &'a Entities<'a> {
+    type Value = Entity;
+    unsafe fn get(&self, idx: Index) -> Self::Value {
+        if let Some(&gen) = self.guard.generations.get(idx as usize) {
+            if gen.is_alive() {
+                Entity(idx, gen)
```
kvark: could `gen` computations can be moved inside

src/world.rs
```diff
@@ -12,26 +15,34 @@ pub trait Component: Any + Sized {
...
+            } else {
+                Entity(idx, gen.raised())
             }
+        } else {
+            Entity(idx, Generation(1))
```
kvark: same here, we can just move out the `Entity(xxx)` construction to the end

src/world.rs
```diff
@@ -52,65 +63,128 @@ impl<'a> EntityBuilder<'a> {
...
 
-fn find_next(gens: &[Generation], lowest_free_index: usize) -> Entity {
-    if let Some((id, gen)) = gens.iter().enumerate().skip(lowest_free_index).find(|&(_, g)| !g.is_alive()) {
-        return Entity(id as Index, gen.raised());
+impl Allocator {
+    fn new() -> Allocator {
+        Allocator {
+            generations: vec![],
+            alive: BitSet::new(),
+            raised: AtomicBitSet::with_capacity(1_000_000),
```
kvark: where does this come from?

kvark:
I'm happy with the code. Is it still being updated? Please merge when finished.

csherratt:
@kvark I want to add some more docs and maybe another look over.

There are a few places I would like to improve since they are suboptimal (the bitset clear for example).

kvark:
Warnings: performance of building the entities got affected. Need to have another look at this change.

PR #45, opened 2016-04-26
Performance Optimizations

csherratt:
This change also adds quite a few benchmarks, and minor performance tweaks.

- Rather then preallocating the array as a giant block of 1M EIDS, we build it sparsely. Memory is lazily added in a thread safe way.
 - The `world::world_new` benchmark is improved from ~1MS to 37us on my machine.

- `allocate()` used a compare-and-exchange when a store would have worked.
  - Reduced allocation times from `51us -> 41us`

This bring `specs` back up the performance it had prior to #44


Cargo.toml
```diff
@@ -26,3 +26,4 @@ pulse = "0.5"
 threadpool = "1.0"
 fnv = "1.0"
 tuple_utils="0.2"
+atom = "0.3"
```
kvark: new line missing

benches/benches.rs
```diff
@@ -0,0 +1,186 @@
...
+        let mut bitset = AtomicBitSet::new();
+        let mut range = (0..1_000_000).cycle();
+        for i in 0..500_000 {
+            // events are set, odds are to keep the branch
+            // prediction from getting to aggressive
+            bitset.add(i * 2);
+        }
+        b.iter(|| range.next().map(|i| bitset.contains(i)))
+    }
+}
```
kvark: new line

src/bitset.rs
```diff
@@ -309,6 +312,56 @@ impl<T> Iterator for BitIter<T>
...
+impl AtomicBlock {
+    fn new() -> AtomicBlock {
+        AtomicBlock {
+            mask: AtomicUsize::new(0),
+            atom: AtomSetOnce::empty()
+        }
+    }
+    fn add(&self, id: Index) -> bool {
+        if self.atom.is_none() {
+            let v = Box::new(unsafe { std::mem::zeroed() });
```
kvark: There is a chance that we double-allocate there, if two concurrent `add()` are called. Not a big deal though, assuming it's a rare case.

kvark:
Looks great! Special thanks for the benchmarks :+1: 
Please merge when ready.

src/bitset.rs
csherratt: Yes there is a chance, it's by design. The allocation that looses the race will get freed, so worst case is a performance penalty.

PR #47, opened 2016-04-28
Storage clear() and DummyStorage

kvark:


src/storage.rs
```diff
@@ -254,6 +263,21 @@ impl<T> UnprotectedStorage<T> for VecStorage<T> {
     }
 }
 
+/// A dummy storage type, used for cases where the component
+/// doesn't contain any data and instead works as a simple flag.
+pub struct DummyStorage<T>(T);
+
+impl<T: Clone + Default> UnprotectedStorage<T> for DummyStorage<T> {
```
csherratt: T should probably be a Copy type over a clone.

src/storage.rs
```diff
@@ -254,6 +263,21 @@ impl<T> UnprotectedStorage<T> for VecStorage<T> {
...
+    fn new() -> Self {
+        DummyStorage(Default::default())
+    }
+    unsafe fn clean<F>(&mut self, _: F) where F: Fn(Index) -> bool {}
+    unsafe fn get(&self, _: Index) -> &T { &self.0 }
+    unsafe fn get_mut(&mut self, _: Index) -> &mut T { &mut self.0 }
```
csherratt: This should probably just panic. Giving to `&mut T` out is dangerous. In the cases with the other types this is assumed that `get_mut` will not be called with multiple indexs. Plus it makes zero sense.

csherratt:
I think `MarkerStorage` or `FlagStorage` would be a more logical name. And do you know of a way to enforce the type is a Zero-Sized-Type?

Clear looks good.

src/storage.rs
kvark: Why? Clone is less restrictive and enough for my needs.

src/storage.rs
kvark: Agreed, no need for get_mut here.

kvark:
That's arguable. The unprotected storage, which it implements, doesn't carry either markers or flags. It does nothing, and that's what Dummy prefix is for. I'll thinkof a better name...

kvark:
How about NullStorage or FakeStorage?

csherratt:
null storage sounds good!

src/storage.rs
```diff
@@ -254,6 +263,21 @@ impl<T> UnprotectedStorage<T> for VecStorage<T> {
...
+/// doesn't contain any data and instead works as a simple flag.
+pub struct NullStorage<T>(T);
+
+impl<T: Clone + Default> UnprotectedStorage<T> for NullStorage<T> {
+    fn new() -> Self {
+        NullStorage(Default::default())
+    }
+    unsafe fn clean<F>(&mut self, _: F) where F: Fn(Index) -> bool {}
+    unsafe fn get(&self, _: Index) -> &T { &self.0 }
+    unsafe fn get_mut(&mut self, _: Index) -> &mut T { panic!("One does not simply modify a NullStorage") }
```
csherratt: :+1:

PR #52, opened 2016-05-08
Return the old value during a insert.

csherratt:
This makes the storage api closer to the rust api, and allows for a more efficient logging of changes. IE a changefeed that works on-top of the of specs.

This is an api breaking change.


kvark:
If we wanted to stick with standard return types, we could have `Result<Option<T>, T>`. Although that would be less descriptive, so I'm merging your approach.

PR #55, opened 2016-05-19
Exported all types need for the return types of World::read() and World::write()

Kimundi:
This is needed to write out the types in, eg, a struct.


kvark:
Thanks!

PR #56, opened 2016-05-25
Added resources to the world object

csherratt:
This is a pretty small change, it add a concept of `Resources` to `specs`. A Resource works like a component in how it is synchronised with systems, but itself does not contain any entities.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "specs"
-version = "0.6.1"
+version = "0.6.2"
 description = """
 Parallel Entity-Component System.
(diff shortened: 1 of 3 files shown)


Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "specs"
-version = "0.6.1"
+version = "0.7.0"
```
kvark: it's not a breaking change, as far as I see

csherratt:
@kvark fixed

kvark:
Thanks, :shipit: !

PR #58, opened 2016-07-05
Safer instant entity creation

### LINKED ISSUE #57: Deadlock when using create_iter()
The following program deadlocks under Mac OS 10.11:

```rust
extern crate specs;

use specs::{Component, VecStorage, World, Planner, System, RunArg};

struct Foo(i32);
impl Component for Foo { type Storage = VecStorage<Foo>; }

struct Bar(i32);
impl Component for Bar { type Storage = VecStorage<Bar>; }

type Ctx = ();

struct Spawner;

impl System<Ctx> for Spawner {
    fn run(&mut self, arg: RunArg, _: Ctx) {
        let (mut foo, mut bar, mut entities) = arg.fetch(|w| {
            let e = w.create_iter();
            let foo = w.write::<Foo>();
            let bar = w.write::<Bar>();
            (foo, bar, e)
        });

        for i in 0 .. 10 {
            let eid = entities.next().expect("Ran out of EIDs");
            foo.insert(eid, Foo(i));
            bar.insert(eid, Bar(i));
        }
    }
}

fn main() {
    let mut world = World::new();
    world.register::<Foo>();
    world.register::<Bar>();

    let mut planner = Planner::new(world, 3);
    planner.add_system(Spawner, "Spawner 1", 10);
    planner.add_system(Spawner, "Spawner 2", 9);

    for i in 0 .. 100 {
        planner.dispatch(());
        println!("Rendered frame {}", i);
    }
}
```

If run as written, Rust detects the deadlock and panics. If I move the `create_iter()` call after the calls to `write()`, the deadlock is not detected, and the program just freezes.

I haven't tested this under other operating systems.

kvark:
Fixes #57

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "specs"
-version = "0.6.2"
+version = "0.7.0"
 description = """
 Parallel Entity-Component System.
 
@@ -23,7 +23,7 @@ exclude = ["doc", ".travis.yml"]
 [dependencies]
 mopa = "0.2"
 pulse = "0.5"
-threadpool = "1.0"
+threadpool = "1.3"
 fnv = "1.0"
 tuple_utils="0.2"
 atom = "0.3"
--- a/examples/basic.rs
+++ b/examples/basic.rs
@@ -63,7 +63,7 @@ fn main() {
         b.0 = a.0 > 0;
     });
     // Deletes an entity instantly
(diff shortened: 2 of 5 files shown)


kvark:
Now that I think of it, we can't expose `Arc<World>` publicly at all, since there is code that assumes that no running systems means we own the unique world reference...

PR #62, opened 2016-07-06
Fixed has_gen() and World access

### LINKED ISSUE #60: Unable to populate components of the dynamically created entities
Our `has_gen` appears to be too strict. It accepts generations outside of the stored array, since it considers these new. But it doesn't understand the newly dynamically allocated entities at the old slots. This results in `EntityIsDead` error upon adding a component. We should extend `has_gen` to fix it.

cc @malleusinferni

### LINKED ISSUE #61: Hide `Arc<World>` in private
Our code assumes that if no systems are running, then we have an unique world reference.

kvark:
Fixes #61 , fixes #60

cc @malleusinferni

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -171,7 +171,7 @@ impl<C> Drop for SystemGuard<C> {
 /// distributes the load in parallel using a thread pool.
 pub struct Planner<C> {
     /// Shared `World`.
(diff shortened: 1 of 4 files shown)


csherratt:
@homu r+

PR #63, opened 2016-07-07
One more fix to has_gen() - to remove it

kvark:


csherratt:
Wonderful! I was bit by this bug just last night.

kvark:
Sorry, I've been messing around trying to fix it without seeing much of the big picture.
We are certainly lacking in tests for the entity deletion/creation.
Besides, @malleusinferni reports that some rare stuff is still missing, albeit no panics any more (yay!).

PR #65, opened 2016-07-07
Test for stillborn entities

malleusinferni:
This commit adds a **failing** test addressing dynamic creation and deletion of entities in different systems. The core intent is to check whether components can be inserted immediately after entity creation. Note that in a test which only creates entities (ie. without deleting any) I was not able to detect the problem this test is designed to highlight.

cc @kvark


csherratt:
Thanks for the test case! #66 should fix it...

kvark:
Great stuff, thank you @malleusinferni , you rock!

PR #69, opened 2016-08-06
SystemInfo: change name from String to Cow<'static, str>

serprex:
This optimizes the common case that a system's name is a string literal.
Cow could be bound to a non-static lifetime,
but I'm unsure if the type complexity is worth it.

https://github.com/search?l=rust&q=specs+add_system&type=Code&utf8=%E2%9C%93

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -14,6 +14,7 @@ extern crate fnv;
 extern crate tuple_utils;
 extern crate atom;
 
+use std::borrow::Cow;
 use std::cell::RefCell;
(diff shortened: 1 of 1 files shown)


kvark:
but systems are not added too often (typically just once, is your use case different?), I don't quite see that optimizing much...

serprex:
This avoids an allocation per call to run_custom

kvark:
Where exactly? I don't see `run_custom` allocating or cloning strings.

serprex:
In drop of SystemGuard

Issue #64, opened 2016-07-07
Support for dynamic component types


mrZalli:
This library seems promising and I would like to use it in a project of mine, but I really can't because I would like the ability to create custom component types at runtime.

The current implementation seems like it uses typeids to map component types into their relative data. It seems like it would be easy to change them from typeids into data of generic type that implements the necessary traits (like `PartialEq`, `Eq`, `Clone`, etc).

The problem is that I don't really know how to make one implementation that supports the current way of accessing components and this dynamic one at the same time. There could be two different `World` implementations with slightly different interfaces but that would be messy and unwanted.

I think I'll fork this project to try and make a dynamic version and see what I can do.

kvark:
Hi @mrZalli and thanks for filing this issue!
Would you mind sharing the details of your use-case, e.g. why and where do you need runtime-defined components?

mrZalli:
Thanks for the fast response!

The project I'm doing (or planning to do) is a sort of a game engine, where most of the content and some rules of the world are in external text files to allow easy editing and modding.

Of course I could allow only a pre-existing selection of components and it would work, but I feel like it would be an arbitrary restriction. Because of that I would like if new components could be introduced in the aforementioned text files when the program starts.

kvark:
> where most of the content and some rules of the world are in external text files to allow easy editing and modding

But how do you process these run-time entities then? There has to be some code doing stuff with them, and dealing with an untyped memory blob doesn't sound too Rusty for the lack of type safety.

mrZalli:
These components values can be enums with different variants for different datatypes or something like JSON strings.

The rules that refer to these components are also defined in text files and could be written in some script language, like Lua.

mrZalli:
Now when I think of this I have a solution of sorts. You can just add another (optional) identifying value alongside the the type-id and it should work. Then you could have additional functions that take these dynamic identifiers as their parameters.
I'll try to do something like this in a fork.

OvermindDL1:
This is the same thing I spoke of before.  Remember when I mentioned that in my old engine that you can define new component types in LUA (type names are a 64-bit integer) and they could define systems over them (or even from C++, or vice-versa, as Aspects work based on the 64-bit integer type IDs).  It is a fairly important feature for moddability.

mrZalli:
So I made a simple support for the dynamic components, but it's still kinda clunky and not tested yet.
https://github.com/mrZalli/specs

mrZalli:
[...]

PR #70, opened 2016-08-21
Dynamic component type feature

mrZalli:
Issue: #64 

I added dynamic identifiers for the components in addition of the original type identifiers, while keeping the interface of `World` and `EntityBuilder` backwards compatible.

The both identifiers are equal, so that there can be multiple components with the same dynamic id but different type id.

(I also added a simple test case for the feature)

### DIFF
--- a/src/world.rs
+++ b/src/world.rs
@@ -48,7 +48,7 @@ impl<'a, C> EntityBuilder<'a, C>
     where C: 'a + PartialEq + Eq + Hash
 {
     /// Adds a `Component` value to the new `Entity`.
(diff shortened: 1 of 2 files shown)


kvark:
@csherratt want to review it?
I've taken a look, fine with that as long as it doesn't break the existing API.

csherratt:
Looks ok to me. But this may need to be tweaked in a follow up change. The problem being that if you want to use string names for component_ids you will end up doing a `clone()` of the strings to fetch anything. Might not be the end of the world.

This is a breaking change. The break is pretty small, just `World` => `World<()>` but it is still a breaking change.

PR #71, opened 2016-08-24
Put planner with threadpool/pulse dependencies behind feature

serprex:
[My own use case has found that the sequencing of components is essentially linear](https://github.com/serprex/rg/commit/f6cb0bf8aa2797eef8758ec8caae44ce6605e966)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,8 @@ rust:
 script:
 - cargo build --verbose
 - cargo test --verbose
+- cargo build --verbose --no-default-features
+- cargo test --verbose --no-default-features
 
 after_success: |
   [ $TRAVIS_BRANCH = master ] &&
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,8 +22,12 @@ exclude = ["doc", ".travis.yml"]
 
 [dependencies]
 mopa = "0.2"
-pulse = "0.5"
-threadpool = "1.3"
+pulse = { version = "0.5", optional = true }
+threadpool = { version = "1.3", optional = true }
 fnv = "1.0"
 tuple_utils="0.2"
 atom = "0.3"
+
+[features]
+default = ["parallel"]
+parallel = ["threadpool", "pulse"]
--- a/examples/basic.rs
+++ b/examples/basic.rs
@@ -25,6 +25,11 @@ impl specs::Component for CompFloat {
 #[derive(Clone, Debug)]
 struct Sum(usize);
 
+#[cfg(not(feature="parallel"))]
+fn main() {
+}
+
+#[cfg(feature="parallel")]
 fn main() {
     let (e, mut planner) = {
         let mut w = specs::World::new();
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -8,16 +8,13 @@
 
 #[macro_use]
 extern crate mopa;
+#[cfg(feature="parallel")]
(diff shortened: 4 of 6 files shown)


acmcarther:
(a quiet user of specs dropping my two cents)

Sorry if I'm jumping the gun a bit on this; I know this PR just feature-gates the planner, but I wanted to get ahead of this change. I actually do use the planner, but I don't use it raw -- I've written a wrapper that automatically identifies the system priorities from explicit dependency annotations.

1. [In use](https://github.com/acmcarther/space_coop/blob/b29236888fedcd1a06c51caa865e111b8657be67/prototype2/client/src/engine/systems.rs)
2. [A sample annotation in a system](https://github.com/acmcarther/space_coop/blob/59186cd806f04ae1a53070dd1ab283a2a5054e7c/prototype2/client/core/synchronization/src/lib.rs#L38)
3. [Test demoing simple usage](https://github.com/acmcarther/space_coop/blob/b29236888fedcd1a06c51caa865e111b8657be67/prototype2/client/infra/automatic_system_installer/src/lib.rs#L183)

My janky wrapper isn't exactly ready for anyone to look at let alone incorporate into specs, but I wanted to emphasize that at least *one* person uses the planner and that it might not be great to hide this feature.

It seems that the planner in its current state is "too much" for linear hierarchies, but "too little" for complex non-linear hierarchies. Maybe we could use this discussion point to spurn more discussion about what the planner is supposed to do for the consumer?

serprex:
NB I've made the planner a default feature; it's opt-out, not opt-in

malleusinferni:
@serprex The weakness here isn't in Planner or the problem domain. It's in your analysis. Two systems can run at the same time if all shared components are accessed as read-only. If you break up your larger systems into smaller stages, you'll expose many more places throughout your code where this can happen. Some things you might try:

- Instead of a giant enum (`AiState::Random`, `AiState::Player`...) use separate components entirely (`struct AiStateRandom`, `struct AiStatePlayer`...). Write separate systems to handle each, and another system that executes AI state changes at the end of a turn when other stuff is happening.
- Use short-lived components for intermediate calculations. For example, if two different systems need to know the position of an entity's foe, create a lightweight AiFoePosition component and write a third system that does nothing but insert and delete it.
- Perform calculations early, and discard them later for entities that don't need them. For example, you might have an AiRandomDirection component that gets inserted on every frame for every single entity, but only used in the DoAiRandom system.

If this level of granularity seems like overkill for your game, you might not need ECS in the first place. That's fine, of course.

kvark:
(using stack-based responding)

@malleusinferni Thanks for a detailed answer!

@acmcarther looks interesting! My worry would be the redundancy of dependency information - first you declare it with your new annotation, but then you still implicitly depend on other systems by sharing the components. So we'd need to take measures to ensure these two dependency trackers are synced, and that would involve some complexity.

> My own use case has found that the sequencing of components is essentially linear

Current parallelization of tasks is far from optimal, sometimes producing sequential execution, but this is an implementation detail ought to change. For example, the most recent discussions about the ticketed locks (@amaranth) would allow further lookup for systems able to run in parallel, improving the efficiency.

kvark:
The idea to make planner optional though has been on the table for a while, and I'm glad to see it being requested here (with a nice PR!). @csherratt could you have a look?

csherratt:
LGTM

PR #74, opened 2016-09-15
World: use fnv

serprex:
TypeId is a u64 in reality. C is usually ()
It may be desirable to make the hasher a type parameter


kvark:
LGTM. @csherratt ?

csherratt:
Thank you @serprex!

PR #75, opened 2016-10-05
debug free of vec storage

### LINKED ISSUE #73: error: Process didn't exit successfully: invalid memory reference
it seems that an error can appear on drop.

I made a [quite minimal project](https://github.com/thiolliere/specs_error) that raises the error on the first test (at least on my computer today)
And if I change code a little, it sometimes raises the error sometimes not.

Do you have the same error ?

thiolliere:
as mentioned in the issue #73 the drain iterator doesn't iterate
correctly on this unsafe inserted vector.
I don't know if it is expected and I will post the question on forums soon.

this commit contains a test that failed in past and an
implementation that pass this test.

basically it cleans the vector like it fills it by swapping value to drop
them and then set it len to 0

closes #73

### DIFF
--- a/src/storage.rs
+++ b/src/storage.rs
@@ -262,10 +262,10 @@ impl<T> UnprotectedStorage<T> for VecStorage<T> {
         VecStorage(Vec::new())
     }
     unsafe fn clean<F>(&mut self, has: F) where F: Fn(Index) -> bool {
-        use std::mem;
+        use std::ptr;
         for (i, v) in self.0.iter_mut().enumerate() {
             if has(i as Index) {
-                mem::swap(v, &mut mem::uninitialized());
+                ptr::drop_in_place(v);
             }
         }
         self.0.set_len(0);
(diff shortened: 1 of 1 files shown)


kvark:
Swapping with some `unitialized` address doesn't sound good. Is this a documented behavior?

thiolliere:
I don't think uninitialized cause any issue.
But I propose another design that follow the truncate standard function of vec.

the truncate function :
```rust
pub fn truncate(&mut self, len: usize) {
    unsafe {
        // drop any extra elements
        while len < self.len {
            // decrement len before the drop_in_place(), so a panic on Drop
            // doesn't re-drop the just-failed value.
            self.len -= 1;
            let len = self.len;
            ptr::drop_in_place(self.get_unchecked_mut(len));
        }
    }
}
 ```

the commit i just made uses `ptr::drop_in_place` instead of `mem::swap` to be like truncate function

so know the only difference between clear() (i.e. truncate(0)) and
clean() is that it doesn't drop in place all elements but only just those who
are really in the vector.

kvark:
Looks great, thanks!

Jragonmiris:
I arrived at the same solution independently. If you're wondering, the issue is basically that the old method was attempting to move uninitialized memory, which is undefined. You need to flip it so that, instead, it doesn't ever TOUCH the uninitialized memory (hence `drop_in_place` on the valid elements rather than forgetting the invalid ones).

The "swap with uninitialized" thing should have worked, due to how swap works, but `drop_in_place` is standard (it's how `Vec` itself drops things).

PR #76, opened 2016-10-06
Fix uninitialized memory read in StorageVec::insert

Jragonmiris:
This is just something I saw while working on a fix for #73 (which I was beaten to). This is an extremely minor patch that makes the intent of `insert` more clear.


Jragonmiris:
Also, I'm not certain `mem::swap` is defined when one of the values is uninitialized. It's probably better for safety.

kvark:
Thanks for the PR!
I checked the `ptr::write` semantics and indeed this is perfect for our case.

PR #79, opened 2016-10-25
Allow storages to handle updating themselves

Aceeri:
Would be extremely useful for sorted storages.


kvark:
Could you provide some details on the example usage?

serprex:
My way of handling this is to make the storage act like NullStorage, don't use get/get_mut/insert. Only be a Storage to be part of tracking entity lifetime. Offer getter/setter like methods on storage which are used to mutate instead

PR #86, opened 2016-12-16
Added World::create_later_build

kvark:


kvark:
Actually, this seems to cause issues:
> Tried to perform an operation on component type that was not registered

Need to take my time to look deeper into that. I thought the storages allow future entity components specifically to make this case work.

kvark:
Ah, no, the error I get was from a different spot, phew!

msiglreith:
Do you have a code snippet showing the difficulty with creating entities currently? I assume it's somehow related to requiring `&mut World`.

kvark:
@msiglreith thanks for taking a look!
Here is the code in [yasteroids](https://github.com/kvark/yasteroids) using `create_now` inside the `fetch()` block:
https://github.com/kvark/yasteroids/blob/514e0af8e40992633371d31e3f95dcc759ec73db/src/sys/aster.rs#L40

Between 0.6 and 0.7, `create_now` turned to require `&mut self`, which makes it applications rather limited to the initialization step only. I'd be able to use the new `create_later_build` here safely now.

I consider the current API for creation/deletion rather obscure, and we need to revise it heavily for specs-0.8. But this would be a matter of a separate issue - for now I just want to have a backwards-compatible extension to get `yasteroids` running (and maybe others will benefit as well?).

So what I'm going to do is - merge this, publish 0.7.1, and then start a discussion for the new API. If you have any objections, now would be the best moment to express them ;)

PR #90, opened 2016-12-28
CheckStorage

Aceeri:
Somewhat similar to `AntiStorage`, allows iteration through entities that have components without returning the components in the iterator.


kvark:
Thanks!

PR #92, opened 2016-12-28
Implement Deref/Mut for Storage

Aceeri:
Allows access to `T::Storage`'s methods from the `Storage<T, A, D>`. Makes specs somewhat easier to use by removing the requirement to get the actual storage before calling any methods on the inner storage.

```rust
let (storage) = arg.fetch(|w| w.write::<Component>());
storage.method();
```
is the equivalent of
```rust
let (storage) = arg.fetch(|w| w.write::<Component>());
let actual_storage = storage.open().1;
actual_storage.method();
```

Pros:
- Easier to call methods on inner storages as well as getting values. (no need for opening the storage).

Cons:
- Introduces more ambiguity to the compiler leading to some cases requiring more explicit typing.

### DIFF
--- a/src/storage.rs
+++ b/src/storage.rs
@@ -136,9 +136,6 @@ impl<T, A, D> Storage<T, A, D> where
     }
 }
 
-
-
-
 /// the status of an insert operation
 pub enum InsertResult<T> {
     /// The value was inserted and there was no value before
--- a/src/storage.rs
+++ b/src/storage.rs
@@ -348,7 +348,7 @@ mod map_test {
 
     #[test]
     fn insert() {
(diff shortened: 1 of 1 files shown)


src/storage.rs
```diff
@@ -91,6 +91,27 @@ impl<'a, T, A, D> Not for &'a Storage<T, A, D> where
     }
 }
 
+impl<T, A, D> Deref for Storage<T, A, D> where
+    T: Component,
+    A: Deref<Target = Allocator>,
```
kvark: nit: you don't need to bound `A` trait here, since you are not using it

Aceeri:
After a bit more thought this would probably be less beneficial as it can increase complexity with abstractions since now you get some weird coercions sometimes.

Issue #89, opened 2016-12-17
TrackedStorage (flagged component storage)


Aceeri:
Having a `TrackedStorage` that would add/remove flags depending on if the component was modified would be a good bonus to performance in a lot of cases. For instance, if you only want to recalculate something once the component is modified.

One problem is if the user uses a mutable storage and iterates instead of joining then it will flag all of the components. This could be solved by comparing the two results on mutable iterations.

Aceeri:
Adding associated types to the `UnprotectedStorage` may make this easier and less painful to use.
e.g.
```rust
trait UnprotectedStorage<'a, T> {
    type Target: 'a;
    type TargetMut: 'a;
    ...
    unsafe fn get(&self, id: Index) -> Target;
    unsafe fn get(&mut self, id: Index) -> TargetMut;
    ...
}
```

This would allow for the storages to return components with a wrapper type for various purposes, in this case it would track if the component was modified through `Deref/DerefMut`.

Current issues with this means adding a lifetime to the UnprotectedStorage bubbles up:
`UnprotectedStorage<'a, T> + Component<'a> + MaskedStorage<'a, T>`

It isn't an issue until you reach the typemap of `TypeId -> Storage`, since the type id requires a `'static` bound and we require a variable/non-'static bound.

Issue for non-'static TypeId: https://github.com/rust-lang/rust/issues/41875

torkleyy:
From Gitter:

We decided to have a `FlaggedStorage` and a `ChangedStorage`.
The first one just flags components which were iterated over mutably (or fetched using `get_mut`), while the latter one does an equality check.

torkleyy:
As commented above, this issue has been split into two parts. `FlaggedStorage` was already implemented by #93. Closing this now and creating a new issue for the `ChangedStorage`.

PR #93, opened 2017-01-20
FlaggedStorage

Aceeri:
A basic implementation for a `TrackedStorage` (https://github.com/slide-rs/specs/issues/89).

This version of the `TrackedStorage` does not implement a way to compare the past version and the modifies version of the components. Usage requires unwrapping the storage to the base `TrackedStorage` which implements `Join` for flagged components.

### Alternative design choices:
**Implement `TrackedStorage` similar to how `MaskedStorage` works.**

_Pros:_
- Would make it possible to have the `Join` of the storage only work on the flagged components without calling any additional methods.
- Might open up a path for a generic storage wrapper like `UnprotectedStorage` but instead as a container for how the `Join` gets the components necessary for the loop (this could be interesting since it would allow for different methods other than bitsets, but would also be pretty hard to get right).

_Cons:_
- Difficult to get all the components instead of just the flagged ones as there is no current way to directly get access to the container storage.
- Might be unintuitive at a glance as you would expect to get all the components when joining the storage.

**Compare modifications for more accurate flags**

_Pros:_
- Removes a large concern for if the user iterates over a mutable tracked storage.

_Cons:_
- Performance benefits of the tracked storage decreases.
- Difficult to implement. Requires a way to directly compare the new "modified" version with the past version.

### DIFF
--- a/src/storage/storages.rs
+++ b/src/storage/storages.rs
@@ -194,3 +194,136 @@ impl<T: Default> UnprotectedStorage<T> for NullStorage<T> {
         Default::default()
     }
 }
+
+/// Wrapper storage that stores modifications to components in a bitset.
+///
+/// **Note: Never use `.iter()` on a mutable component storage that uses this.**
+///
+/// #Example Usage:
+/// 
+/// ```rust,ignore
+///# extern crate specs;
+///# use specs::{Planner, World, RunArg, Component, TrackedStorage, VecStorage, System, Join};
+/// 
+/// pub struct Comp(u32);
+/// impl Component for Comp {
+///     // `TrackedStorage` acts as a wrapper around another storage.
+///     // You can put any store inside of here (e.g. HashMapStorage, VecStorage, etc.)
+///     type Storage = TrackedStorage<Comp, VecStorage<Comp>>;
+/// }
+/// 
+/// pub struct CompSystem;
+/// impl System<()> for CompSystem {
+///     fn run(&mut self, arg: RunArg, _: ()) {
+///         let (entities, mut comps) = arg.fetch(|w| {
+///             (w.entities(), w.write::<Comp>()) 
+///         });
+/// 
+///         // Iterates over all components like normal.
+///         for (entity, comp) in (&entities, &comps).iter() {
+///             // ...
+///         }
+/// 
+///         // **Never do this**
+///         // This will flag all components as modified regardless of whether the inner loop
+///         // did modify their data.
+///         for (entity, comp) in (&entities, &mut comps).iter() {
+///             // ...
+///         }
+/// 
+///         // Instead do something like:
+///         for (entity, comp) in (&entities, &comps.check()).iter() {
+///             if true { // check whether you should modify this component or not.
+///                 let mut comp = comps.get_mut(entity);
+///                 // ...
+///             }
+///         }
+/// 
+///         // To iterate over the flagged/modified components:
+///         for (entity, flagged_comp) in (&entities, (&comps).open().1).iter() {
+///             // ...
+///         }
+/// 
+///         // Clears the tracked storage every frame with this system.
+///         (&mut comps).open().1.clear();
+///     }
+/// }
+///# fn main() { }
+/// ```
+pub struct TrackedStorage<C: Component, T: UnprotectedStorage<C>> {
+    mask: BitSet,
+    storage: T,
+    phantom: PhantomData<C>,
+}
+
+impl<C: Component, T: UnprotectedStorage<C>> UnprotectedStorage<C> for TrackedStorage<C, T> {
+    fn new() -> Self {
+        TrackedStorage {
+            mask: BitSet::new(),
+            storage: T::new(),
+            phantom: PhantomData,
+        }
+    }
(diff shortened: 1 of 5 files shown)


src/storage.rs
```diff
@@ -334,6 +334,138 @@ impl<T: Default> UnprotectedStorage<T> for NullStorage<T> {
...
+///         // This will flag all components as modified regardless of whether the inner loop
+///         // did modify their data.
+///         for (entity, comp) in (&entities, &mut comps).iter() {
+///             // ...
+///         }
+/// 
+///         // Instead do something like:
+///         for (entity, comp) in (&entities, &comps.check()).iter() {
+///             if true { // check whether you should modify this component or not.
+///                 let mut comp = comps.get_mut(entity);
```
kvark: Hmm, we should be able to make this nicer. Asking for `get_mut` on an entity that we are currently iterating on would be slower than just giving the value right away. Perhaps, the `check()` could iterate over some nice wrapper aka `std::hashmap::Entry`, so that actually accessing the data becomes straightforward (in terms of performance)?

src/storage.rs
```diff
@@ -334,6 +334,138 @@ impl<T: Default> UnprotectedStorage<T> for NullStorage<T> {
...
+    }
+    unsafe fn remove(&mut self, id: Index) -> C {
+        self.mask.remove(id);
+        self.storage.remove(id)
+    }
+}
+
+impl<C: Component, T: UnprotectedStorage<C>> TrackedStorage<C, T> {
+    /// All components will be cleared of being flagged.
+    pub fn clear(&mut self) {
```
kvark: needs to be renamed to `clear_flags`

src/storage.rs
Aceeri: I'm assuming the main problem is just that calling `get_mut` means it is checking again whether the entity has the component, even though we already know?

Aceeri:
Also, this runs into the same problem as the `Join` implementation for a `&'a mut Storage`. Maybe it might be better to have something like a `Join` and `JoinMut`, or possibly have `Join` with it automatically implementing it for `&'a T/&'a mut T where T: Join`?

kvark:
> I'm assuming the main problem is just that calling get_mut means it is checking again whether the entity has the component, even though we already know?

Yes.

> Maybe it might be better to have something like a Join and JoinMut?

I don't think `JoinMut` would solve that. It would require all the iterated things to be mutable, while we need to mix and match them.

src/storage.rs
```diff
@@ -334,6 +334,138 @@ impl<T: Default> UnprotectedStorage<T> for NullStorage<T> {
...
+    unsafe fn clean<F>(&mut self, has: F) where F: Fn(Index) -> bool {
+        self.mask.clear();
+        self.storage.clean(has);
+    }
+    unsafe fn get(&self, id: Index) -> &C {
+        self.storage.get(id)
+    }
+    unsafe fn get_mut(&mut self, id: Index) -> &mut C {
+        // calling `.iter()` on a mutable reference to the storage will flag everything
+        self.mask.add(id);
```
torkleyy: I imagine that such a `TrackedStorage` would be best used in the following way:

```rust
// pos is a component storage with a `TrackedStorage`
for (vel, pos) in (&vel, &mut pos).join() {
    *pos += vel * delta_time; 
}
```

In another system:

```rust
for (pos_handle, matrix) in (pos.check(), &mut mvp).join() {
    if pos.flagged(pos_handle) {
        // Recalc
    } else {
        // Use cached version
    }
}
```

It's inconvenient to do whatever checks and often too complicated, instead we might want a method on `UnprotectedStorage` which is called after every iterated component, where we check for equality. This would also solve the lifetime problem of a guard.

Aceeri: What would a cached version be? Wouldn't the matrix just stay the same?

torkleyy: @Aceeri Yeah, the matrix would stay just the same. The comment above is just not very clear about that.

src/storage/storages.rs
```diff
@@ -194,3 +197,137 @@ impl<T: Default> UnprotectedStorage<T> for NullStorage<T> {
         Default::default()
     }
 }
+
+/// Wrapper storage that stores modifications to components in a bitset.
+///
+/// **Note: Never use `.iter()` on a mutable component storage that uses this.**
+///
+///# Example Usage:
```
torkleyy: Nice docs!

torkleyy:
In case you don't want to add or change anything, I'd say you can merge it after squashing.

bors delegate=Aceeri

src/storage/storages.rs
```diff
@@ -194,3 +197,138 @@ impl<T: Default> UnprotectedStorage<T> for NullStorage<T> {
...
+///             // ...
+///         }
+///
+///         // Clears the tracked storage every frame with this system.
+///         (&mut comps).open().1.clear_flags();
+///     }
+/// }
+///# fn main() { }
+/// ```
+pub struct TrackedStorage<C, T> {
```
torkleyy: I just had an idea:

Given that both this storage and `ChangeStorage` track the mutation of their components, maybe we should rename this one to `FlaggedStorage`?

torkleyy:
Nice we finally have a `FlaggedStorage` :tada: 
Thanks @Aceeri!

bors r+

PR #97, opened 2017-02-06
Cleanup `Planner::run*` implementations

murarth:
Places all `impl_run!` macro invocations inside a single `impl` block.

The result is functionally identical, but makes the documentation more
readable.


kvark:
Nicely done, thanks!

Issue #96, opened 2017-02-06
Make registering components idempotent


jeffparsons:
I'd like to be able to have systems in my library automatically register any components they use upon initialisation, so that clients don't need to do that separately. The consequence of this is that multiple systems would attempt to register the same component type.

[`World::register`](https://slide-rs.github.io/specs/specs/struct.World.html#method.register) can't be used in that way directly, by the look of it because it doesn't keep track of what components are registered before blindly "re-initialising" the storage for the component type.

I thought of checking this myself using something like [TypeMap](https://github.com/reem/rust-typemap), then it occurred to me that you're effectively tracking the information necessary for this internally in Specs anyway, so it wouldn't be much of a stretch to make `register` idempotent`.

Any objection to this, or reason I've missed that it shouldn't be idempotent? I'm happy to implement it and PR if no objections.

kvark:
@jeffparsons It sounds sensible to me. How about a separate `register_maybe` variant that returns a `bool`?

jeffparsons:
@kvark How's this? https://github.com/slide-rs/specs/pull/98

jeffparsons:
Replaced ~~#98~~ with https://github.com/slide-rs/specs/pull/99.

jeffparsons:
I think we can call this one done. :)

PR #98, opened 2017-02-07
Add World::register_*_maybe (#96)

jeffparsons:
Added `World::{ register_w_comp_id_maybe, register_maybe }` functions.

The implementation of `register_w_comp_id_maybe` doesn't just
call through to `register_w_comp_id` because I'd need to be able
to clone `comp_id`, which isn't possible without requiring the
component ID type to implement `Clone`.

My feeling is that this small redundancy in implementation is worthwhile
to avoid placing additional bounds on the component ID type.

### DIFF
--- a/src/world.rs
+++ b/src/world.rs
@@ -1,5 +1,6 @@
 use std::any::TypeId;
 use std::collections::HashMap;
+use std::collections::hash_map::Entry;
 use std::hash::{BuildHasherDefault, Hash};
 use std::sync::{RwLock, RwLockReadGuard, RwLockWriteGuard};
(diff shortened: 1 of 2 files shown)


kvark:
@jeffparsons uh, now that I think of it, this should be the default behavior. Since otherwise adding the same type twice would mess things up and the old one would be lost forever. Sorry for changing my mind!

jeffparsons:
Closing in favour of https://github.com/slide-rs/specs/pull/99.

jeffparsons:
@kvark No worries. Let's try again! 😁

PR #99, opened 2017-02-08
Make World::register* functions idempotent (#96)

jeffparsons:
Previously repeated calls to `register` (and friends) would stomp over
the existing storage for the component type.

Now repeated calls will silently do nothing.

### DIFF
--- a/src/world.rs
+++ b/src/world.rs
@@ -249,9 +249,14 @@ impl<C> World<C>
         }
     }
     /// Registers a new component type and id pair.
+    ///
+    /// Does nothing if the type and id pair was already registered.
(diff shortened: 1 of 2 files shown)


jeffparsons:
Replaces https://github.com/slide-rs/specs/pull/98.

jeffparsons:
I haven't included any way to _ask_ whether a component is already registered; I didn't want to bend the existing API (by adding a return value) without a really good reason, and don't personally have a use case for checking instead of just blindly saying "register it if you need to please!".

Happy to add it if anyone wants that.

kvark:
Perfect, thanks!

Issue #101, opened 2017-03-03
Consider using futures


torkleyy:
Could you try [futures](https://github.com/alexcrichton/futures-rs) instead of `threadpool`? Probably make it an optional dependency? I think it would also be interesting in terms of performance.

Additionally, `amethyst` could reuse the `CpuPool` then (Probably make it an `Arc`?).

torkleyy:
If you give me some hints / explanation / links how this parallel model works with `Pulse` and the channel, I'd try it myself.

cjhowe:
Is this a duplicate of #42?

torkleyy:
@cjhowe Oh you are right, it is covered by that, but anyways, I tried to implement that, but I don't think there is really an advantage (it would have been just an internal thing, anyways).

PR #102, opened 2017-03-03
Use futures and cpu pool to run systems

torkleyy:
Addresses issue #101 

It's not fully working, one test (`wait`) is failing and the benchmarks are behaving really weird (sometimes crashing with `index out of bounds: the len is 4096 but the index is 4096`).

Could somebody help me to fix this issue?

### DIFF
--- a/src/planner.rs
+++ b/src/planner.rs
@@ -2,7 +2,7 @@ use std::cell::RefCell;
 use std::sync::{mpsc, Arc};
 
 use pulse::{Pulse, Signal};
-use threadpool::ThreadPool;
+use futures_cpupool::{CpuFuture, CpuPool};
 
 use super::{Component, JoinIter, World, Entity};
 
@@ -99,10 +99,10 @@ pub struct Planner<C> {
     world: Arc<World>,
     /// Permanent systems in the planner.
     pub systems: Vec<SystemInfo<C>>,
-    wait_count: usize,
+    waiting: Vec<CpuFuture<(), ()>>,
     chan_out: mpsc::Sender<SystemInfo<C>>,
     chan_in: mpsc::Receiver<SystemInfo<C>>,
-    threader: ThreadPool,
+    cpu_pool: CpuPool,
 }
 
 impl<C: 'static> Planner<C> {
@@ -112,10 +112,10 @@ impl<C: 'static> Planner<C> {
         Planner {
             world: Arc::new(world),
             systems: Vec::new(),
-            wait_count: 0,
+            waiting: Vec::new(),
             chan_out: sout,
             chan_in: sin,
-            threader: ThreadPool::new(num_threads),
+            cpu_pool: CpuPool::new(num_threads),
         }
     }
     /// Add a system to the dispatched list.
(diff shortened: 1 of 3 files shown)


src/planner.rs
```diff
@@ -36,7 +30,9 @@ impl RunArg {
         // It's only receiving `&mut self` to prevent deadlocks, and these are not possible in
         // the pre-fetch phase we are in right now.
         let entities = unsafe { transmute::<&World, &mut World>(&self.world) }
```
torkleyy: Not sure if this is still safe

src/planner.rs
```diff
@@ -55,10 +51,18 @@ impl RunArg {
 pub trait System<C>: Send {
     /// Run the system, given its context.
     fn run(&mut self, RunArg, C);
+
+    /// Clone this system as a trait
+    /// object
+    fn box_clone(&self) -> Box<System<C>>;
```
torkleyy: Is there a better way to solve cloning the systems?

src/planner.rs
```diff
@@ -231,7 +196,7 @@ impl<C: 'static> Planner<C> {
     impl_run!( run0w3r [] [R0, R1, R2] );
     impl_run!( run0w4r [] [R0, R1, R2, R3] );
     impl_run!( run1w0r [W0] [] );
-    impl_run!( run1w1r [W0] [R0] );
+    impl_run!(run1w1r[W0][R0]);
```
kvark: no need to touch this line

src/planner.rs
```diff
@@ -18,12 +15,9 @@ impl RunArg {
     pub fn fetch<'a, U, F>(&'a self, f: F) -> U
         where F: FnOnce(&'a World) -> U
     {
-        let pulse = self.pulse.borrow_mut().take()
-                        .expect("fetch may only be called once.");
-        let u = f(&self.world);
-        pulse.pulse();
-        u
+        f(&self.world)
```
kvark: You seem to be missing the point of the whole `fetch` phase here. If the code could just call `f`, we wouldn't need that phase in the first place! Here is why it's needed:
  - we don't want multiple systems to lock the storages they need to work on, since that is a recipe for deadlocks
  - thus, we separate the system execution into 2 phases: fetch and processing. The `fetch` phase execution is serialized
  - once a system finishes the `fetch` phase, it "detaches" from the planner and continues execution (the actual data processing) on its own, in parallel with all other things

I don't see how simply switching from threadpools to future cpu pools solves that.

src/planner.rs
torkleyy: It seems `rustfmt` formats exactly this line...

src/planner.rs
torkleyy: @kvark Yes you are right, I didn't really understand how it works.

> The fetch phase execution is serialized

Do you mean sequentialized?

kvark: > Do you mean sequentialized?

Yes. Basically, the planner fires up the systems one by one. It waits for a system to finish the `fetch` state before proceeding to the next one.

torkleyy: Thanks for the explanation, I'll try to fix it

torkleyy:
@kvark It seems the `bitset` module is never exported, but the benchmarks are using it. Are the benchmarks just broken or am I missing something?

torkleyy:
@kvark I think the replacement I did only makes it slower... I found another way to fix my PR in amethyst and I'm not sure if integrating futures is a benefit for this library. If it is, I'm sure you (or somebody else knowing how this library works) can do it much better.

torkleyy:
I think it works better with threadpool and asset loading is also easier that way.

PR #104, opened 2017-03-04
Add a PlannerBuilder

torkleyy:
This allows sharing the thread pool, which is pretty important for highly parallel applications in order to allow a good performance.

Sorry I've done so bad in my previous PR, I hope this one is better.

### DIFF
--- a/src/planner.rs
+++ b/src/planner.rs
@@ -3,6 +3,7 @@ use std::sync::{mpsc, Arc};
 
 use pulse::{Pulse, Signal};
 use threadpool::ThreadPool;
+use num_cpus::get as get_num_cpus;
 
 use super::{Component, JoinIter, World, Entity};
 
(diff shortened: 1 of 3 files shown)


src/planner.rs
```diff
@@ -92,6 +93,113 @@ impl<C> Drop for SystemGuard<C> {
...
+/// Also optional, use this if you want to
+/// share the thread pool with other parts of
+/// your crate (you should do this if you use
+/// threads somwhere else).
+///
+/// If you didn't specify a thread pool nor a number of
+/// threads, a new thread pool with the number of virtual
+/// cpus will be created.
+pub struct PlannerBuilder {
+    world: Option<World>,
```
kvark: what is the point of having an `Option` here in the first place?

torkleyy: Yes, I thought about using `with_thread_pool`, but I thought it's not descriptive enough because you still have another parameter. Yes, you are right I could have asked for world already in the builder constructor, but I think it is nicer to use. Of course it's not compile-time enforced then.

torkleyy: Plus, now the pool size defaults to the number of cpus, which should be pretty common. I do have another solution in mind, though:

```rust
enum PoolOptions {
    Pool(Arc<ThreadPool>),
    NumThreads(usize),
    NumCpus,
}

impl Planner {
    pub fn from_options(options: PoolOptions) -> Self {
        
    }
}
```

kvark:
How about we have just `new` going the `NumCpu` path and `from_pool` accepting a thread pool? that would be simple enough and covers all use-cases.

torkleyy:
I found a non-breaking way, using a trait `PoolSize`.

kvark:
@torkleyy I don't think it's worth it. Let's not have a fancy API, let's have a simple one, with `new()` and `from_pool(...)` constructors.

torkleyy:
So you mean I should just take a plain `usize`?

kvark:
No, I mean that we don't even need a constructor that takes `usize`. A most common case would be to use the number cpus, and that's what I propose for the `new()` constructor to do. A less common case would be to receive the `Arc<ThreadPool>`, which covers a case where you want a specific number of threads.

nchashch:
Maybe add a `Planner::new`, taking a `usize` and `Planner::new_num_cpus`. There are use cases where you don't need an `Arc<ThreadPool>` and you need to set the number of threads (for example to benchmark performance with and without multithreading).

And it isn't obvious that the way you get a `Planner` with a set number of threads is call `Planner::from_pool`.

torkleyy:
Closing because title does not match the content anymore. #106 replaces this PR

Issue #72, opened 2016-08-25
Discussing alternative ways of specifying system priority


acmcarther:
Expanding upon from #71. Sorry for the wall of text, I've got a ton of thoughts about this issue and didn't have a lot of time to refine them down to a bite sized blurb. Thanks for reading, anyway!

My thoughts, mostly boil down to two things: 
- That data dependencies aren't _necessarily_ the same as runtime dependencies, 
- That maintaining the "knowledge" of who goes before who in more than one place, at least with respect to a given system, is a bit of a pain. 
### Data dependency is not _always_ the same as run order dependency

It turns out in at least my case that a data dependency is not the same as the run ordering. If that were the case, I'd have cycles in my execution strategy. One example (from my client)

**[network::AdapterSystem]**
 -> **[some interpreting system]** Via incoming network events
 -> **[player::InputSystem]** Via some interpreted state
 -> **[network::AdapterSystem]** Via outbound player input event

I resolve this in my particular case by just not having the network system depend on things that supply outgoing events. It emits them first thing on the next tick.

I've also got cases where I enforce a "dependency" that's not directly visible via the data dependencies. One example from my client.

**[network::AdapterSystem]**
-> **[synchronization::System]** reconstructs state emitted by server, writing (LOTS of) component state
~> **[renderer::System]** uses that component state.

There are alternate approaches to solve both problems. It would be possible to break the adapter system into two systems, or to have the renderer system depend on some sentinel value. In my case, it was more difficult at an implementation level to do the former due to my underlying socket, and the latter is actually more complicated due to a Send issue with my renderer.
### Centralizing the "knowledge" of system priority.

A different, and for me more compelling, reason is that maintaining the explicit priorities is just a pain in the butt when specifying them the current way.

I've found, at least in my case, that keeping the knowledge of dependency ordering localized, even if its often redundant, is less overhead than the alternatives. Those alternatives as I see them are either maintaining a huge set of constants with numerical priorities, or making systems responsible for knowing their priority in the raw. The former gets pretty busy and doesn't separate concerns well, and the latter IMO isn't viable since that priority value is intrinsically dependent on the values of its dependencies, and the number of other systems -- it becomes a maintenance headache.

Thus, at least in my swing at the problem, I put off actually numbering the systems, and just resolve those values at the last possible second when every system, and their explicit dependencies, is available.
### Why now?

[...]

PR #106, opened 2017-03-06
Add more constructors

torkleyy:
* Use the number of cpus as pool size
* Allows sharing the thread pool

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,10 +24,11 @@ exclude = ["doc", ".travis.yml"]
 mopa = "0.2"
 pulse = { version = "0.5", optional = true }
 threadpool = { version = "1.3", optional = true }
+num_cpus = { version = "1.0", optional = true }
 fnv = "1.0"
 tuple_utils="0.2"
 atom = "0.3"
 
 [features]
 default = ["parallel"]
-parallel = ["threadpool", "pulse"]
+parallel = ["threadpool", "pulse", "num_cpus"]
--- a/examples/basic.rs
+++ b/examples/basic.rs
@@ -51,7 +51,7 @@ fn main() {
         w.add_resource(Sum(0xdeadbeef));
 
         // Planner is used to run systems on the specified world with a specified number of threads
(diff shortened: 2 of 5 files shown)


kvark:
Why don't you want to simply remove the constructor that accepts a number of threads? I don't see many consumers of this one, and they can work around by creating `ThreadPool` themselves.

torkleyy:
As @nchash wrote

> Maybe add a Planner::new, taking a usize and Planner::new_num_cpus. There are use cases where you don't need an Arc<ThreadPool> and you need to set the number of threads (for example to benchmark performance with and without multithreading).

> And from the user's perspective it isn't obvious that the way you get a Planner with a set number of threads is call Planner::from_pool.

Additionally, doing it that way is non breaking and easier to understand what it does from the function name. 

And there's not really a disadvantage of having one more function.

kvark:
Well, I don't agree with @nchash proposal here. I think `new()` should be dumb simple and good for everyone, so having no parameters is great.

> There are use cases where you don't need an Arc and you need to set the number of threads (for example to benchmark performance with and without multithreading).

The non-multithreaded mod is a bigger task - #72 , we don't need to cater for it since we don't yet know how it is going to work.

> And from the user's perspective it isn't obvious that the way you get a Planner with a set number of threads is call Planner::from_pool.

As I said, a regular user would not care. An advanced user would figure `from_pool` out.

> Additionally, doing it that way is non breaking ...

Don't worry about breaking too much. We are on the verge of 0.8, which is fine to break API a little.

> ... and easier to understand what it does from the function name.

The easiest thing to understand is `new()` without parameters, and that's what I'm arguing for.

> And there's not really a disadvantage of having one more function.

Sure thing, as long as you keep the most used case unencumbered. So, it's fine to have an extra `with_num_threads` constructor, but not fine to have `usize` parameter for `new()`.

torkleyy:
@kvark Good idea, did that

kvark:
Thanks @torkleyy ! Last things - make CI happy and squash the commits please.

PR #107, opened 2017-03-07
clarify usage of convenience-functions

minecrawler:
They were a very mysterious concept at first and I had to read the source in order to understand them. Writing just a few sentences should clarify the usage for everyone else with the same problem.


kvark:
Gorgeous, thank you!

Issue #87, opened 2016-12-16
Entity creation/deletion APIs


kvark:
This concern came up in #86.

We currently have:
 - `create_now` - needs `&mut self` and thus only applicable to early initialization
 - `create_later` - doesn't allow components building conveniently, so the usage is rather limited
 - `create_later_build` - much usable, but the name is long and confusing

I suggest we revise these APIs for specs-0.8. Basic proposal:
 - leave `create_now` as is, given that it's sufficiently expressive
 - rename `create_later_build` -> `create`, assuming it's what users actually want to use
 - rename `create_later` to `create_pure`, reflecting the fact only the `Entity` is created with no building.

Can we do better? cc @msiglreith @csherratt

msiglreith:
I'm in favor of the renaming, definitely better than the current ones.
The new names should also be updated for `RunArg` (https://github.com/slide-rs/specs/blob/master/src/planner.rs#L43-L45)

sinesc:
Would it be possible to get a more convenient API to create entities+components from within a system?
Unless I'm overlooking something, we're currently required to fetch() the component storages, create() an entity and then insert() the components into the storages.
There's a nice comment in https://github.com/slide-rs/specs/issues/48#issuecomment-262008082 on that issue as well

kvark:
@sinesc there is another option, a simpler one:
https://github.com/kvark/yasteroids/blob/master/src/sys/bullet.rs#L82

sinesc:
But can you do that after the system has iterated over the entities given that fetch() may only be used once?
Unless you're constantly shooting, you'll probably want to trigger spawn based on whatever happened during the iteration. I could not find a way to get a mutable world (for the entity builder) at that point in the system.

kvark:
Right.
> you'll probably want to trigger spawn based on whatever happened during the iteration

Care to give an example?

sinesc:
Something like spawning an entity with some property based on the one that spawned it, e.g. same directional vector, position, ...

vitvakatu:
I'll do it

kvark:
Fixed in #113

PR #113, opened 2017-03-07
Rename funcs according to #87

vitvakatu:
- rename `create_later_build` to `create`, assuming it's what users actually want to use
- rename `create_later` to `create_pure`, reflecting the fact only the Entity is created with no building.
Also rename `RunArg::create` to `RunArg::create_pure` for the same reason.


kvark:
Thanks!

PR #114, opened 2017-03-09
Re-export GatedStorage as public

msiglreith:


src/storage.rs
```diff
@@ -240,6 +240,7 @@ impl<'a, T, A, D> Join for &'a mut Storage<T, A, D> where
 
 
 #[cfg(feature="ticket")]
+///
```
msiglreith: Suggestions for documentation are welcome! Didn't know how to properly describe the use-case /:

PR #116, opened 2017-03-10
Made benchmarks runnable

WaDelma:
I tried to run benchmarks and noticed that they don't work:
* bitsets were moved to different place, were private and didn't have documentations
* There were mutability problems with bunch of benchmarks.

After fixing these, benchmarks work, but `create_now` test panics:
````
test world::create_now              ... thread 'main' panicked at 'index out of bounds: the len is 1024 but the index is 1024', src\libcollections\vec.rs:1401
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
stack backtrace:
   0: specs::world::Allocator::allocate
   1: <specs::world::World<C>>::create_now
   2: benches::world::create_now
   3: test::run_test
   4: test::run_tests
   5: test::run_tests_console
   6: test::test_main
   7: test::test_main_static
   8: benches::__test::main
   9: std::panicking::try::do_call
  10: __rust_maybe_catch_panic
  11: std::rt::lang_start
  12: main
  13: __tmainCRTStartup
  14: unit_addrs_search
  15: unit_addrs_search
````

### DIFF
--- a/benches/benches.rs
+++ b/benches/benches.rs
@@ -31,13 +31,13 @@ mod world {
 
     #[bench]
     fn create_now(b: &mut test::Bencher) {
-        let w = specs::World::new();
+        let mut w = specs::World::new();
         b.iter(|| w.create_now().build());
     }
 
     #[bench]
     fn create_now_with_storage(b: &mut test::Bencher) {
-        let w = create_world();
+        let mut w = create_world();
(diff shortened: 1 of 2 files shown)


kvark:
Thank you! We've got to figure out why it panics.

Issue #54, opened 2016-05-10
Scoped parallel execution within a single System


kvark:
@Kimundi:

> Would it be possible to support scoped-parallel iteration over entities in a System? As in, having a build-in way to distribute work over the backing threadpool of specs rather than "working around" that with multiple systems and/or launching your own threadpool

Suggestions: use `crossbeam` or `rayon`
A nice library for our consideration - https://github.com/reem/rust-shared-mutex

WaDelma:
I'll try to switch threadpool for rayon. (Wanted to test speed difference and stumbled into #116)

WaDelma:
I have been looking at how to integrate `rayon` to `specs`.
I currently have just switched thread pool to use `rayon` one which seemed to need `rayon-core` which is currently unstable, but switching it was easy enough.

Bigger task is to implement parallel iterator for the `join` stuff as it required touching the `bitset`.
Is there any sources that explain it other than the code? Only thing that I found searching 'Hierarchical bitset' was [this](http://www.cs.loyola.edu/~binkley/papers/tcsrt08-hbit-vectors.pdf) pdf. It seems at cursory clance to be similar. Is it accurate?

I am also currently little lost in `rayons` internals even though I have pretty much watched all the video presentations and read blogs on it. Have to just put time in to familiarise myself with the code base.

WaDelma:
Also crossbeam is currently in flux:
https://www.reddit.com/r/rust/comments/5ygbwb/crossbeam_request_for_help/
https://www.reddit.com/r/rust/comments/5yp2b2/initial_release_candidate_100_by_ticki_pull/

torkleyy:
Implemented by PR #120 by @WaDelma.

PR #120, opened 2017-03-12
Integrating rayon and scoped threading

WaDelma:
This is meant to address #54 
NOTE: This PR is highly experimental and not meant to be merged as is. I am also not even sure if any of the parallel iteration stuff is correct.    
This PR is currently meant for gathering feedback and discussion,


src/bitset.rs
```diff
@@ -22,7 +26,7 @@ pub const SHIFT3: usize = SHIFT2 + BITS;
 /// a certain component exists. It does not track the `Generation` of the
 /// entities that it contains.
 ///
-/// Note, a `BitSet` is limited by design to only 1,048,576 indices.
+/// Note, a `BitSet` is limited by design to only 1,048,576 indices for 32 bit and 16,777,216 indices for 64 bit architectures.
```
vitvakatu: It's really cool that you added comments about `BitSet`. But I prefer if it will be separate PR.

WaDelma: I added the comments mostly for my own understanding.
Can remove.

src/planner.rs
```diff
@@ -121,7 +121,9 @@ impl<C: 'static> Planner<C> {
     /// Creates a new planner with a thread pool that has
     /// `num_threads` threads.
     pub fn with_num_threads(world: World, num_threads: usize) -> Planner<C> {
-        let config = Configuration::new().set_num_threads(num_threads);
+        let config = Configuration::new()
+            .set_num_threads(num_threads)
+            .set_panic_handler(Arc::new(|e| println!("{:?}", e)));
```
vitvakatu: Perhaps we can make this message more readable?
Just notice, it's pretty good for now.

WaDelma:
I started implementing parallel join and encountered few problems.

Because `Join::get` requires mutable reference and we are splitting the joining there needs to be multiple mutable references. It currently uses mutable raw pointer for this and then creates mutable reference out of it. This is dubious at best and most propably incorrect, but I am not sure how to fix it.

Other problem is that, because `Allocator` is wrapped `RwLock` and thus the `Storage` doesn't implement `Send` it doesn't work with rayons parallel iterator. I am not sure what to do with this.

WaDelma:
Because `RwLockReadGuard` I tried to switch it to use `Lock` which if I understand right is either `RwLock` or `TicketedLock` depending feature.
This caused tests to deadlock:
````
test join_two_components has been running for over 60 seconds
test stillborn_entities has been running for over 60 seconds
test wait has been running for over 60 seconds
````
How it should be used?

**Comment by kvark** [comment]
@WaDelma sorry, I haven't had a chance to get a good look at your code.

**Comment by WaDelma** [comment]
I rebased this PR and pushed experimentaion where I switched methods to return tickets instead of guards.
This doesn't work with `Storage` as they don't `Deref` to their contents.

**Comment by WaDelma** [comment]
So I got it to build and run by implementing `Send` and `Sync` for `Storage` (Horribly unsafe, I know) and I fixed brokenness of `BitProducer`...

But I don't have any idea how to make it work without doing those `Send` and `Sync` implementations.

File: Cargo.toml [hunk 3]
```diff
@@ -37,9 +37,14 @@ version = "0.7"
 optional = true
 features = ["unstable"]
 
+[dependencies.rayon-core]
```
torkleyy: No need to add `rayon-core` as dependency, you are only using `rayon`.

WaDelma: Thanks. I forgot there from earlier when it was needed.

kvark:
@WaDelma could you describe in some detail the problem you are trying to solve, the approach you took, and the difficulties that are left unresolved please?

WaDelma:
The basic idea was to implement `ParallelIterator` for `BitSet` first. It basically uses the `BitIter` for folding and adds extra splitting logic. The idea for splitting is to look at the highest level that has more than one bit left and split it into two by either masking all bits of the mask under or over avarage. This doesn't give perfect splits, but like with quicksort it should give good splits if the data is random in the set. (I could have improved this if I had made bitset to record how many bits are set on each branch).

After that I started to implement `ParallelIterator` for `Join`. This is where I have had the most of the problems. First problem was how I would share the `Join::Value` inside the `JoinProducer`. I finally settled on `Arc<UnsafeCell<Join::Value>>` as when splitting each of the `Producer`s should have mutable access to (their non-overlapping subset) of `Join::Value`. I am not that happy with the unsafety of it afaik it produces multiple mutable references which is undefined behaviour even with `UnsafeCell`.

Biggest issue is that `Join` has been implemented for `&/&mut Storage` and immutable references are `Send` only for `Sync` types (`Send` is needed for moving it to thread pool). Other issues include: `World::read/write` returning `Storage` with guards which are also non-`Send`. I currently just circumvent these last two problems by implementing `Sync` and `Send` for `Storage` which is blatant lie.

I pretty certain that that `Send` issue cannot be solved without redoing much of the API which is what I tried to avoid. I tried to go around the second `Send` issue by making `World::read/write` return tickets instead of guards, but that didn't work as `Deref` is not and cannot be implemented for them
nd` issue cannot be solved without redoing much of the API which is what I tried to avoid. I tried to go around the second `Send` issue by making `World::read/write` return tickets instead of guards, but that didn't work as `Deref` is not and cannot be implemented for them (and other API needs that implementation). Also making `Allocator` inside `World` to use ticketed lock made tests to deadlock‽

src/join.rs
```diff
@@ -77,7 +78,13 @@ pub trait Join {
     {
         JoinIter::new(self)
     }
-
+    /// Create a joined parallel iterator over the contents.
+    #[cfg(feature="parallel")]
+    fn par_join(self) -> JoinParIter<Self>
```
torkleyy: I'd say there's no guarantee that a storage supports joining in parallel.

WaDelma: Hmmm.... So I should really be creating new trait that has that? That would propably resolve some problems I have had. Have to look into it.

Currently all storages need to support random indexing (Join::get) and only time splitting then be impossible would be if the structure needed to touch mutably other indices when indexing or if distinct indices would return mutable references to same index.

torkleyy: @WaDelma Yeah, that's what I thought. Just a quick idea, I thought it might be easier then.

src/bitset.rs
```diff
@@ -233,6 +232,14 @@ pub trait BitSetLike {
             set: self,
         }
     }
+
```
torkleyy: The bitset stuff moved to https://github.com/slide-rs/hibitset btw.

WaDelma: I noticed that when I tried to rebase it. Have to move my changes there.

src/bitset.rs
```diff
@@ -366,42 +373,154 @@ impl<T> Iterator for BitIter<T>
...
+        }
+    }
+
+    pub struct BitProducer<'a, T: 'a + Send + Sync>(pub BitIter<&'a T>);
+
+    impl<'a, T: 'a + Send + Sync> UnindexedProducer for BitProducer<'a, T>
+        where T: BitSetLike
+    {
+        type Item = Index;
+        fn split(mut self) -> (Self, Option<Self>) {
```
torkleyy: You documented `BitIter` very nicely, could you do that for your own code, too?

src/join.rs
```diff
@@ -114,6 +121,106 @@ impl<J: Join> std::iter::Iterator for JoinIter<J> {
...
+
+    struct JoinProducer<'a, 'b, J>
+      where
+        J: Join + Send,
+        J::Type: Send,
+        J::Value: 'a + Send,
+        J::Mask: 'b + Send + Sync,
+    {
+        keys: BitProducer<'b, J::Mask>,
+        values: Arc<UnsafeCell<J::Value>>,
```
torkleyy: Yeah, this is unfortunate; we have to make the guarantee that every possible safe code, even if it's wrong (or unusual), can not cause undefined behavior. However, consider the following:

```rust
pub struct MyUnusualCollection<T> {
    access_counter: usize, // incremented on every access
    inner: Vec<T>,
}
```

What I wanted to show is that even though it is usually the case that you *could* access the storage with multiple mutable references given they don't access the same access, we are not allowed to make such an assumption.

I didn't give it a try, but if I were you I'd try to go with a separate trait, as already proposed earlier, and optionally allow a storage to provide a `split` method, which would be just the [`Vec::chunks`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.chunks) method for vectors, for example.

ebkalderon:
@WaDelma Is this pull request still being worked on? I really like the introduction of parallel iterators into the API.

WaDelma:
Yes it's still worked on. Currently I am focusing on [splitting](https://github.com/slide-rs/hibitset/pull/4) the bitset part to the hibitset crate. There is currently stack overflow error in travis and I can't reproduce it locally. After that is resolved and that pr is merged, I can continue with this pr.

src/world.rs
```diff
@@ -295,6 +297,10 @@ impl<'a> Join for &'a Entities {
     }
 }
 
+// TODO: Is this correct?
+#[cfg(feature="parallel")]
+impl<'a> ParJoin for &'a Entities {}
```
torkleyy: I didn't look at the rest of the code yet, but sharing an immutable struct always is safe.

src/join.rs
```diff
@@ -139,6 +263,11 @@ macro_rules! define_open {
                 ($($from::get($from, i),)*)
             }
         }
+        #[cfg(feature="parallel")]
```
torkleyy: BTW, we removed the feature flag.

src/world.rs
WaDelma: Yeah, it should be fine. I just put that there to remind me to recheck it as I stopped coding pretty late.

src/join.rs
WaDelma: Ah I didn't actually realise that. Will remove it from my PR too then.

torkleyy:
@WaDelma Please comment here as soon as you consider your PR ready for review.

torkleyy:
Oops, just noticed that I forgot to release a new `shred` version. It'll be on crates.io any second ;)

WaDelma:
I think it should be ready to review. There is some TODOs, but I am not sure what to do about them.

Cargo.toml
```diff
@@ -32,6 +32,7 @@ mopa = "0.2"
 shred = "0.4"
 shred-derive = "0.3"
 tuple_utils = "0.2"
+rayon = { version = "0.7" }
```
torkleyy: Given that we don't use any unstable features anymore, let's not use an inline table but just a string.

src/join.rs
```diff
@@ -113,6 +131,104 @@ impl<J: Join> std::iter::Iterator for JoinIter<J> {
     }
 }
 
+/// `JoinParIter` is an `ParallelIterator` over a group of `Storages`.
```
torkleyy: is **a** `ParallelIterator`

src/join.rs
```diff
@@ -113,6 +131,104 @@ impl<J: Join> std::iter::Iterator for JoinIter<J> {
...
+        let (keys, values) = self.0.open();
+        let producer = BitProducer((&keys).iter());
+        let values = Arc::new(UnsafeCell::new(values));
+        bridge_unindexed(JoinProducer::<J>::new(producer, values), consumer)
+    }
+}
+
+struct JoinProducer<'a, 'b, J>
+  where
+    J: Join + Send,
```
torkleyy: It's usually recommended to not have any bounds on the type parameters unless necessary. Maybe use two type parameters instead?

src/join.rs
```diff
@@ -113,6 +131,104 @@ impl<J: Join> std::iter::Iterator for JoinIter<J> {
...
+    J::Type: Send,
+    J::Value: 'a + Send,
+    J::Mask: 'b + Send + Sync,
+{
+    keys: BitProducer<'b, J::Mask>,
+    values: Arc<UnsafeCell<J::Value>>,
```
torkleyy: Are you sure we need an `Arc` here? Parallel iterators should be scoped, so no need for `'static`; just use clone the `&UnsafeCell`s.

src/join.rs
```diff
@@ -113,6 +131,104 @@ impl<J: Join> std::iter::Iterator for JoinIter<J> {
...
+        JoinProducer {
+            keys: keys,
+            values: values,
+            _marker: PhantomData,
+        }
+    }
+}
+
+// TODO: This is required so that there isn't lock for values (which would make the whole parallel iterator pointless).
+// Is this nessary/correct? The way the splits happen should guarantee that the values are accessed only in one thread.
```
torkleyy: Still have to review this in depth, but given that you use `UnsafeCell` we'll have such an `impl` somewhere.

src/join.rs
```diff
@@ -1,6 +1,13 @@
 use std;
+use std::marker::PhantomData;
+use std::sync::Arc;
+use std::cell::UnsafeCell;
+
+use hibitset::{BitSetAnd, BitIter, BitSetLike, BitProducer};
```
torkleyy: Nicely ordered imports. Great ;)

src/join.rs
WaDelma: What you mean by two type parameters?

src/join.rs
WaDelma: Hmm... Yeah that should actually work now that I think about it.

src/join.rs
WaDelma: Yeah. The implementation is nessary. So the only question is about it's correctness.

src/join.rs
WaDelma: I could actually combine the two different lifetimes together (dunno why I didn't do this earlier) and remove the `PhantomData`.

src/join.rs
```diff
@@ -113,6 +129,97 @@ impl<J: Join> std::iter::Iterator for JoinIter<J> {
...
+  where
+    J: Join + Send,
+    J::Type: Send,
+    J::Value: 'a + Send,
+    J::Mask: 'a + Send + Sync,
+{
+    type Item = J::Type;
+    fn split(self) -> (Self, Option<Self>) {
+        let (cur, other) = self.keys.split();
+        let prod = JoinProducer::new;
```
torkleyy: This is a little confusing, could you just use `JoinProducer::new()` down there?

tests/tests.rs
```diff
@@ -308,3 +309,125 @@ fn register_idempotency() {
...
+        }
+    }
+    let mut dispatcher = DispatcherBuilder::new()
+        .add(Iter(flags.clone()), "iter", &[])
+        .build();
+    dispatcher.dispatch(&mut world.res);
+
+    let (ref first, ref second, ref error) = *flags;
+    if let Some((int, boolean)) = *error.lock().unwrap() {
+        panic!("Entity that shouldn't be in the join: CompInt({:?}) CompBool({:?})", int, boolean);
```
torkleyy: Could simplify this:

```rust
assert_eq!("Entity shouldn't be in the join", *error.lock.unwrap(), None);
```

tests/tests.rs
```diff
@@ -308,3 +309,125 @@ fn register_idempotency() {
...
+                int.0 += 1;
+            });
+        }
+    }
+    let mut builder = DispatcherBuilder::new();
+    for i in 0..255 {
+        // TODO: Remove allocation by using ""
+        builder = builder.add(Incr, &i.to_string(), &[]);
+    }
+    let failed = Arc::new(Mutex::new(vec![]));
```
torkleyy: Same here, no `Arc` needed.

src/join.rs
```diff
@@ -113,6 +129,97 @@ impl<J: Join> std::iter::Iterator for JoinIter<J> {
     }
 }
 
+/// `JoinParIter` is a `ParallelIterator` over a group of `Storages`.
+#[must_use]
+pub struct JoinParIter<J: Join>(J);
```
torkleyy: No need for any bounds here.

src/join.rs
WaDelma: Okay. I have been on edge on this one. When I change it to the `JoinProducer::new` it becomes harder to read, but it's bit odd this way too.

src/join.rs
```diff
@@ -1,6 +1,11 @@
 use std;
+use std::cell::UnsafeCell;
+
+use hibitset::{BitSetAnd, BitIter, BitSetLike, BitProducer};
+
+use rayon::iter::ParallelIterator;
+use rayon::iter::internal::{UnindexedProducer, UnindexedConsumer, Folder, bridge_unindexed};
```
WaDelma: Do I put the function import still as last here?

WaDelma: Also is the `ParallelIterator` after `internal` module?

torkleyy: Doesn't really matter because it's readable both ways; I usually do place the imports which access the deeper hierarchy at the end.

But no need to spend too much time on that, I'll do a cleanup PR after this one is merged anyways.

kvark:
Nice review @torkleyy , and awesome work @WaDelma !

src/join.rs
```diff
@@ -196,9 +196,10 @@ impl<'a, J> UnindexedProducer for JoinProducer<'a, J>
     type Item = J::Type;
     fn split(self) -> (Self, Option<Self>) {
         let (cur, other) = self.keys.split();
-        let prod = JoinProducer::new;
         let values = self.values;
-        (prod(cur, self.values), other.map(|o| prod(o, values)))
+        let first = JoinProducer::new(cur, values);
```
torkleyy: Nice!

src/join.rs
```diff
@@ -113,6 +129,98 @@ impl<J: Join> std::iter::Iterator for JoinIter<J> {
...
+        let (keys, values) = self.0.open();
+        let producer = BitProducer((&keys).iter());
+        let values = UnsafeCell::new(values);
+        bridge_unindexed(JoinProducer::<J>::new(producer, &values), consumer)
+    }
+}
+
+struct JoinProducer<'a, J>
+  where
+    J: Join + Send,
```
torkleyy: It would be nice to cut down bounds on the struct to a minimum. You could probably just do `JoinProducer<'a, M, V>` or something. The reason for this convention is that it would force you to add these bounds everywhere, infecting all the types using it.

Consider this struct:

```rust
struct BoxMap<K, V, H> {
    map: HashMap<K, V, H>,
}
```

So no bounds necessary in the struct declaration.

src/join.rs
```diff
@@ -113,6 +129,98 @@ impl<J: Join> std::iter::Iterator for JoinIter<J> {
...
+    {
+        let JoinProducer {values, keys, ..} = self;
+        let iter = keys.0.map(|idx| unsafe {
+            // This unsafe block should be safe if the `J::get`
+            // can be safely called from different threads with distinct indices.
+
+            // The `J::get` gets called for distinct indices because splitting happens by `BitProducer::split`
+            // which should ensure that all splitted parts are distinct.
+
+            // TODO: There can be multiple mutable references to storage and thus this is not safe, right?
```
torkleyy: Err..is this comment outdated? It's basically the whole cheat in here.

src/join.rs
```diff
@@ -113,6 +129,98 @@ impl<J: Join> std::iter::Iterator for JoinIter<J> {
...
+    J: Join + Send,
+    J::Type: Send,
+    J::Value: 'a + Send,
+    J::Mask: 'a + Send + Sync,
+{
+    type Item = J::Type;
+    fn split(self) -> (Self, Option<Self>) {
+        let (cur, other) = self.keys.split();
+        let values = self.values;
+        let first = JoinProducer::new(cur, values);
```
torkleyy: Nice!

src/join.rs
WaDelma: I am still not sure if it's UB or not. It's true that each index is only used only once, but there is still mutable reference constructed on the whole storage each time.  And AFAIK that would still break the contract of mutable references in rust. This could be fine, but as the unsafety quidelines are in the works and what can and can't be done with it up in air, I am not 100% confident on it.

src/join.rs
WaDelma: So M is mask here? And V is value? How am going to call `Join::get` then?

src/join.rs
torkleyy: I'm pretty sure it's correct due to the fact that `UnsafeCell` prohibits optimization based on aliasing rules.

WaDelma: "However, it is up to the abstraction designer to ensure that no two mutable references obtained this way are active at the same time, and that there are no active mutable references or mutations when an immutable reference is obtained from the cell. This is often done via runtime checks.", says `UnsafeCell` [documentation](https://doc.rust-lang.org/std/cell/struct.UnsafeCell.html).

src/join.rs
torkleyy: You'd still do

```rust
impl<'a, J>
```

but for

```rust
JoinProducer<'a, J::Mask, J::Value>
```

> How am going to call `Join::get` then? And how I am going to get `Join::Type`?

By adding these bounds only where they're necessary.

src/join.rs
torkleyy: @WaDelma I'm aware of the wording in the documentation. Let's see what @kvark thinks about it. I don't see a reason why it shouldn't work.

src/join.rs
WaDelma: `impl<'a, J> UnindexedProducer for JoinProducer<'a, J::Mask, J::Value>` doesn't work due to `J` being unconstrained.

src/join.rs
kvark: Don't take my word as final, but I read this line:
> it is up to the abstraction designer to ensure that no two mutable references obtained this way are active at the same time

And then see our code doing `&mut *values.get()` in multiple threads, and see a clear conflict.
Thus, it appears unsafe to me.

src/join.rs
torkleyy: I see. Just leave it as is then.

src/join.rs
WaDelma: That's what I was reading from it too, but I don't know what to do about it. Best way would just keep it as raw pointer, but I don't see way to do it.

torkleyy: > Best way would just keep it as raw mutable pointer, but I don't see way to do it.

You should always prefer `UnsafeCell` over raw pointers.

WaDelma: Only way that I can think how to make it work would be put everything inside storage to `UnsafeCell` and index them immutably.

src/storage/mod.rs
```diff
@@ -20,6 +20,9 @@ pub mod storages;
 #[cfg(test)]
 mod tests;
 
+/// A trait that guarantees that mutable storage getter can be used on distinct indices from different threads.
```
torkleyy: I think this needs a clearer warning in which cases one may implement it, but I don't see any reason we couldn't do that after merging.

torkleyy:
Let's merge this. I think this will totally boost our benchmarks :wink: 

bors r+

torkleyy:
bors retry

torkleyy:
What? Shouldn't that work?

bors r+

WaDelma:
Where did you get the info about legality of it? It would be interesting to read more about it.

torkleyy:
@WaDelma Mainly discussed with @kvark. Docs are lacking here unfortunately :(

PR #121, opened 2017-03-12
Remove Sync requirement on resources

cjhowe:
I'm not sure why this requirement is here. First of all, `Mutex` implements `Sync` for any type that implements `Send`, and [`TicketedLock` implements `Send` and `Sync` for everything](https://github.com/kvark/ticketed_lock/blob/master/src/lib.rs#L158). For this reason, the `ResourceLock` should always be `Send` and `Sync` (since it is just a wrapped `Mutex<TicketedLock<T>>`). So it seems like any type should be fine for resources.

EDIT: Also, the thing which seems to cause `Component` to require `Sync` is the `Storage` that wraps it. I don't think this applies to resources.

### DIFF
--- a/src/world.rs
+++ b/src/world.rs
@@ -300,7 +300,7 @@ trait ResourceLock: Any + Send + Sync {}
 
 mopafy!(ResourceLock);
 
-impl<T:Any+Send+Sync> ResourceLock for Lock<T> {}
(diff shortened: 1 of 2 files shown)


cjhowe:
This is necessary.

Issue #124, opened 2017-04-07
Use rayon for parallel execution


torkleyy:
Would be great if Rust libraries would use one common threadpool and because [`rayon`](https://github.com/nikomatsakis/rayon) supports [parallel iterators, futures, scoped and static parallelism as of 0.7](http://smallcultfollowing.com/babysteps/blog/2017/04/06/rayon-0-7-released/), it is the ideal choice!

If there are no concerns, I'll implement it.

WaDelma:
I actually started implementing this in #120, but got stuck plumbing it through specs. It also was before 0.7 and relied on unstable internal API which now is stable.

I should really look again into it, but if you want to you can do it too either using what I did as basis or discarding it if it's bad (I am not that confident on how good it's approach is).

torkleyy:
@WaDelma I've already implemented it, but only as a thread pool. So it doesn't yet address https://github.com/slide-rs/specs/issues/54, which can still be done later.

PR #125, opened 2017-04-07
Use rayon instead of threadpool

torkleyy:
This PR does not implement scoped parallel execution, but can be used as a basis for that.

Fixes https://github.com/slide-rs/specs/issues/124


torkleyy:
@kvark seems rayon is propagating the panics differently. Any clue how to solve this?

WaDelma:
You can see fix to the panic problem in my [pr](https://github.com/WaDelma/specs/blob/e5263113da6c7d3a58b0c6d46c1b8406159945d6/src/planner.rs#L126)

torkleyy:
@kvark would it be okay to just have rayon as a thread pool for now?

kvark:
@torkleyy I think so. Eventually, we'd like `rayon` to at least often be used with `specs`, if not by `specs`.

PR #127, opened 2017-04-12
Remove ticket feature

torkleyy:
The ticket feature was non-additive. This way, the library only provides a version with tickets.

I'll do some crate polishing after this is merged.


kvark:
Looks reasonable, thank you!
Waiting for CI and merging.

PR #128, opened 2017-04-12
Crate polishing

torkleyy:
* rustfmt
* reorder imports
* newlines between methods
* replace `*` import from planner


kvark:
Looks great, thank you!

PR #129, opened 2017-04-15
Added larger tupleic gates and joins.

WaDelma:
We ran into a problem with `Gate` not being implemented for tuples with 9 fields and this PR increases the size of tuples the traits are implemented for.


kvark:
Oh my :)
THanks!

PR #130, opened 2017-04-17
Basic serialization and deserialization for storages

Aceeri:
This PR adds some basic serialization and deserialization into specs storages. I'm not sure how to go about having a single function call for multiple storages since I would like to avoid dynamic dispatch with serde. This could probably be solved with a macro of some sort, but I'm not entirely sure on how to design it.

In detail:
`serde::Serialize` is implemented for `Storage<T, A, D>` so you can just call any method that accepts a `serde::Serialize` like:
```rust
let serialized = serde_json::to_string(&storage);
```

Deserialization is a bit more complicated since there are a couple of cases where you might want to deserialize into a world:
- Adding a scene to the world ontop of the other elements.
- Loading the startup scene of the world from data.
- Adding an entity prefab of some sort, so only partial loading into the world.

So instead of implementing `serde::Deserialize` on the `Storage<T, A, D>` (which would return a new storage and not have a way to make sure the world has the correct entities that correspond to the storage), there is a `PackedData<T>` struct. The `PackedData` struct implements `serde::Deserialize` and contains a list of offsets and components. Offsets correspond to the component list along with a "base". 

So if you have a PackedData like so:
```rust
PackedData<Comp> {
  offsets: [1, 5, 7],
  components: [ Comp(1), Comp(5), Comp(0) ],
}
```

Then the offsets mean that the `components[0]` belongs to the entity `base + offsets[0]` (or base + 1) and `components[1]` belongs to the entity `base + offsets[2]` (or base + 5). By default the base is 0 and can be shifted by the `rebase` method for `PackedData`. (This just shifts all the offsets down by the base).

You use this struct to merge into an existing specs storage with the `merge` method implemented on storages by passing in a list of entities and the packed data.

For instance:
```rust
let entity_list = vec![1, 5, 7, 8, 9, 10, 11, 15];
let packed_data = PackedData {
  offsets: [3, 4, 6, 7],
  components: [Comp(1), Comp(2), Comp(3), Comp(4)],
};
storage.merge(&entity_list, packed_data);
```
This merges the `Comp(....)` into the entity in the `entity_list` using the offset as the index. So `Comp(1)` would be inserted into the storage under the entity `8` and `Comp(3)` would be inserted into the storage under the entity `11`.
If you called `rebase(3)` on the `PackedData` then the entities it would merge into would be entity `1` and entity `8` since the offsets would be shifted down by 3.

### DIFF
--- a/examples/serialize.rs
+++ b/examples/serialize.rs
@@ -44,10 +44,7 @@ mod s {
             println!("Serialized storage: {}", serialized);
 
             // Get a list of all entities in the world
-            let mut entity_list = Vec::new();
-            for entity in (&entities).join() {
-                entity_list.push(entity);
-            }
+            let mut entity_list: Vec<specs::Entity> = (&entities).join().collect();
 
             // Remove all components
             for (entity, _) in (&entities, &components.check()).join() {
@@ -57,7 +54,7 @@ mod s {
             // Deserialize with entity list
             let mut list = serde_json::from_str::<PackedData<CompSerialize>>(&serialized);
             println!("list: {:?}", list);
-            let created = components.merge(&mut entity_list, list.unwrap());
+            let created = components.merge(entity_list.as_slice(), list.unwrap());
 
             for (entity, _) in (&entities, &components.check()).join() {
                 println!("Has: {:?}", entity);
--- a/examples/serialize.rs
+++ b/examples/serialize.rs
@@ -1,0 +1,101 @@
+
+extern crate specs;
+#[cfg(feature="serialize")]
+extern crate serde;
+#[cfg(feature="serialize")]
+#[macro_use]
+extern crate serde_derive;
+extern crate serde_json;
+
+use std::fmt;
+
+#[cfg(feature="serialize")]
+mod s {
+    use serde::{self, Serialize};
+    use serde_json;
+    use specs::{self, Join, PackedData, Gate};
+
+    #[derive(Debug, Serialize, Deserialize)]
+    struct CompSerialize {
+        field: u32,
+        other: bool,
+    }
+    impl specs::Component for CompSerialize {
+        type Storage = specs::VecStorage<CompSerialize>;
+    }
+
+    struct SerializeSystem;
+    impl specs::System<()> for SerializeSystem {
+        fn run(&mut self, arg: specs::RunArg, _: ()) {
+            use fmt::Display;
+
+            let (entities, mut components) = arg.fetch(|w| {
+                let entities = w.entities();
+                let mut components = w.write::<CompSerialize>();
+
+                (entities, components)
+            });
+
+            // Serialize the storage into JSON
+            let mut buffer: Vec<u8> = Vec::new();
+            let mut serializer = serde_json::Serializer::pretty(buffer);
+            let result = components.serialize(&mut serializer);
+            let serialized = serializer.into_inner().iter().map(|b| *b as char).collect::<String>(); 
+            println!("Serialized storage: {}", serialized);
+
+            // Get a list of all entities in the world
+            let mut entity_list = Vec::new();
+            for entity in (&entities).join() {
+                entity_list.push(entity);
+            }
+
+            // Remove all components
+            for (entity, _) in (&entities, &components.check()).join() {
+                components.remove(entity);
+            }
+
+            // Deserialize with entity list
+            let mut list = serde_json::from_str::<PackedData<CompSerialize>>(&serialized);
+            println!("list: {:?}", list);
+            let created = components.merge(&mut entity_list, list.unwrap());
+
+            for (entity, _) in (&entities, &components.check()).join() {
+                println!("Has: {:?}", entity);
+            }
+        }
+    }
+
+    pub fn main_redirect() {
+        let mut world = specs::World::<()>::new();
+        world.register::<CompSerialize>();
+
+        world.create_pure();
+        world.create_pure();
+        world.create_now().with(CompSerialize { field: 5, other: true }).build();
+        world.create_pure();
+        world.create_pure();
+        world.create_now().with(CompSerialize { field: 5, other: true }).build();
+        world.create_now().with(CompSerialize { field: 10, other: false }).build();
+        world.create_pure();
+        world.create_now().with(CompSerialize { field: 0, other: false }).build();
+
+        let mut planner = specs::Planner::<()>::new(world);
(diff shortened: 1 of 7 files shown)


examples/serialize.rs
```diff
@@ -0,0 +1,101 @@
...
+
+            // Serialize the storage into JSON
+            let mut buffer: Vec<u8> = Vec::new();
+            let mut serializer = serde_json::Serializer::pretty(buffer);
+            let result = components.serialize(&mut serializer);
+            let serialized = serializer.into_inner().iter().map(|b| *b as char).collect::<String>(); 
+            println!("Serialized storage: {}", serialized);
+
+            // Get a list of all entities in the world
+            let mut entity_list = Vec::new();
```
kvark: could do `let entity_list: Vec<_> = (&entities).join().collect()`

src/storage.rs
```diff
@@ -220,10 +223,38 @@ impl<T, A, D> Storage<T, A, D>
...
+    ///
+    /// e.g.
+    /// ```rust,ignore
+    ///let list = vec![Entity(0, 1), Entity(1, 1), Entity(2, 1)];
+    ///let packed = PackedData { offsets: [0, 2], components: [ ... ] };
+    ///storage.merge(&list, packed);
+    /// ```
+    /// Would merge the components at offset 0 and 2, which would be `Entity(0, 1)` and `Entity(2, 1)` while ignoring
+    /// `Entity(1, 1)`.
+    pub fn merge<'a>(&'a mut self, entities: &'a Vec<Entity>, mut packed: PackedData<T>) {
```
kvark: any reason to have `&'a Vec<Entity>` instead of `&'a [Entity]`?

src/storage.rs
```diff
@@ -220,10 +223,38 @@ impl<T, A, D> Storage<T, A, D>
...
+        for (component, offset) in packed.components.drain(..).zip(packed.offsets.iter()) {
+            match entities.get(*offset as usize) {
+                Some(entity) => { self.insert(*entity, component); },
+                None => { println!("No entity at offset {:?}", *offset as usize) }
```
kvark: should be an `error!`

src/storage.rs
```diff
@@ -262,7 +293,69 @@ impl<'a, T, A, D> Join for &'a mut Storage<T, A, D>
...
+        use serde::ser::SerializeStruct;
+
+        // Serializes the storage in a format of PackedData<T>
+        let (bitset, storage) = self.open();
+        let mut structure = serializer.serialize_struct("PackedData", 2)?;
+        let mut components: Vec<&T> = Vec::new();
+        let mut offsets: Vec<u32> = Vec::new();
+        for index in bitset.iter() {
+            offsets.push(index);
+            unsafe {
```
kvark: nit: reduce the `unsafe` scope to `storage.get(index)`

src/storage.rs
```diff
@@ -262,7 +293,69 @@ impl<'a, T, A, D> Join for &'a mut Storage<T, A, D>
...
+    A: Deref<Target = Allocator>,
+    D: Deref<Target = MaskedStorage<T>>,
+{
+    fn serialize<S: serde::Serializer>(&self, serializer: S) -> Result<S::Ok, S::Error> {
+        use bitset::BitSetLike;
+        use serde::ser::SerializeStruct;
+
+        // Serializes the storage in a format of PackedData<T>
+        let (bitset, storage) = self.open();
+        let mut structure = serializer.serialize_struct("PackedData", 2)?;
```
kvark: what is `2` magic number here?

src/storage.rs
```diff
@@ -262,7 +293,69 @@ impl<'a, T, A, D> Join for &'a mut Storage<T, A, D>
...
+/// Offsets define which entities the components correspond to, based on a list of entities
+/// the packed data is sent in with.
 ///
+/// If the list of entities is all entities in the world, then the offsets in the 
+/// packed data are the indices of the entities.
+pub struct PackedData<T> {
+    /// List of components.
+    pub components: Vec<T>,
+    /// Offsets used to get entities which correspond to the components.
+    pub offsets: Vec<u32>,
```
kvark: would it be more appropriate to use `Index` here instead of `u32`?

src/storage.rs
```diff
@@ -262,7 +293,69 @@ impl<'a, T, A, D> Join for &'a mut Storage<T, A, D>
...
+#[cfg(feature="serialize")]
+impl<T> PackedData<T> {
+    /// Shifts all offsets in the packed data to match a new base.
+    ///
+    /// Useful if you want to merge some components ontop of the current world:
+    /// `packed.rebase( ... /* amount of entities in world */ );`
+    ///
+    /// If the base is higher than the offset then the function will return the index of that offset.
+    pub fn rebase(&mut self, base: u32) -> Option<usize> {
+        for index in 0..self.offsets.len() {
```
kvark: would be cleaner to do `for (index, offset) in self.offsets.iter_mut().enumerate() {...}`

src/storage.rs
Aceeri: Number of fields in the PackedData, I'll add a comment.

src/storage.rs
```diff
@@ -262,7 +293,69 @@ impl<'a, T, A, D> Join for &'a mut Storage<T, A, D>
...
+            if self.offsets[index] < base {
+                return Some(index);
+            }
+            self.offsets[index] -= base;
```
torkleyy: Isn't this method for merging them ontop of the current world? So why don't you add base but subtract it?

torkleyy: I think I understood somehow, but we probably need a little bit better documentation for this method. Also, I'm not sure how you would use it in an actual application.

msiglreith:
Cool!
Regarding merging, I'm wondering what would be the optimal strategy to retrieve the amount of required entities to create before merging. It might be useful to add a field containing the number of entities to avoid having to iterate over all offset to find the maximum, or is it required that offsets is ordered?

src/storage.rs
Aceeri: I'm thinking I might just remove the rebase stuff since you could just use the entity list instead (and it is less weird because it won't get wrapping weirdness if the base is higher than the offsets).

Aceeri:
@msiglreith If you deserialize from a storage the offsets should always be ordered since it just goes through the `BitIter`. If we want I could add an insertion-sort-esque thing to the deserialization.

kvark:
Thank you! 🎆 🎆 🎆

PR #131, opened 2017-04-18
Dense vector storage

kvark:
Inspired by calx-ecs ;)

### DIFF
--- a/examples/basic.rs
+++ b/examples/basic.rs
@@ -21,7 +21,7 @@ impl specs::Component for CompBool {
 #[derive(Clone, Debug)]
 struct CompFloat(f32);
 impl specs::Component for CompFloat {
-    type Storage = specs::VecStorage<CompFloat>;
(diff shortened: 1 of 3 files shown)


OvermindDL1:
I'd think this would be more for memory savings rather than for performance though, yes?  I do exactly mappings like this in my sparse component tables in my ECS in C++, though I name it SparseArray since the components are sparse into a dense array.  ^.^;

kvark:
> I'd think this would be more for memory savings rather than for performance though, yes?

I think it's both. Scanning a dense vector would give you more data cache hits, thus less RAM access, thus higher performance.

> Unfortunately, this does not help the performance.

That is expected and totally fine. Those benchmarks need to be updated (created https://github.com/lschmierer/ecs_bench/issues/20).

As for this PR - thanks for having a look! I'm merging it. In the near future, we'll probably want to move the exact storage implementations into separate modules for cleaner code.

PR #133, opened 2017-04-19
[WIP] ExternalSystem type

kvark:
The implementation is not correct yet, and this is a little bigger problem than just this PR. It's about `RunArg` getting sent while the system guard considers it done. We need to:
  1. prevent sending of `RunArg` or storing it. Perhaps, `run` should return it?
  2. make the system guard also wait for the external system, somehow


torkleyy:
I think I do have an idea, I'm trying it out atm.

torkleyy:
Implemented as "thread-local systems" as of #150

PR #139, opened 2017-04-26
CheckStorage Entry

Aceeri:
`CheckStorage` will return an `Entry` which can be used to get components from a storage without the unnecessary checks.


kvark:
Nice optimization!

PR #141, opened 2017-04-30
Remove bounds from world struct and entity builder

Aceeri:
Removes bounds on the struct, but leaving them in the implementations removes a lot of boilerplate when using the `World`. Since not all methods related to the `World` require those bounds.


kvark:
Sweet!
@homu r+

kvark:
@homu retry

kvark:
Sadly, it doesn't seem to respond.

PR #143, opened 2017-05-01
Better ergonomics for component id worlds

Aceeri:
I think this is a better design for worlds with component ids. I'm assuming most components in a world would not want a component id, only a couple. So if a person mid-project wanted to change it out to use a component id they wouldn't need to change a lot of trivial `register` -> `register_w_comp_id`, `read` -> `read_w_comp_id`, etc.

### DIFF
--- a/src/world.rs
+++ b/src/world.rs
@@ -117,7 +117,7 @@ impl<'a, C> EntityBuilder<'a, C>
     }
 
     /// Adds a `Component` value to the new `Entity`.
-    pub fn with_w_comp_id<T: Component>(self, comp_id: C, value: T) -> EntityBuilder<'a, C> {
+    pub fn with_w_comp_id<T: Component>(self, comp_id: Option<C>, value: T) -> EntityBuilder<'a, C> {
         self.1
             .write_w_comp_id::<T>(comp_id)
             .insert(self.0, value);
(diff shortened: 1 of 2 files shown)


kvark:
I'm not sure this is strictly a positive change. The use case has not been proven to be a problem yet, and `Option<C>` is more verbose.

Aceeri:
I think the verbosity is worth not having to modify any function related to the world when switching from a `World<()>` to a `World<C>`.

If you use `read` in your systems when using a `World<()>` and then change the world to be a `World<C>` then you have to change *all* the methods from `read` to `read_w_comp_id`.

kvark:
> If you use read in your systems when using a World<()> and then change the world to be a World<C> then you have to change all the methods from read to read_w_comp_id.

Yes, but it's a one time operation, which is not going to be needed for most users.

How about, instead of turning it into `Option<C>`, we implement `register` and friends for `C: Default` ?

torkleyy:
@kvark I guess this may make transition harder than it is, because you may forget to change that one invocation.

I'm fine with how it is, also it is not too hard to replace the same method name. But I still don't understand - if we want this - why an `Option`? `()` implements `Hash`.

The title should probably say "Easier adoption for component ids".

@Aceeri Hm, I first liked the idea, but I guess for the user side a find & replace is already easier than watching out where you are passing a `None` and changing that. That way, the compiler can help out.

Aceeri:
@torkleyy Because you can't pass a `()` into a world that isn't a `World<()>`.

torkleyy:
@Aceeri 

#150 changes the `World` so that each component may or may not have a component id (of any type). It's not that type-safe, but on the other hand you almost never want none or all of them to have an id. So you can just do

```rust
world.register::<Pos>();
```

and

```rust
world.register_with_id::<Custom, _>("foo");
```

with the same world.

Do you consider that practical enough?

PR #151, opened 2017-05-24
Add SmallVec storage.

Object905:
SmallVec storage from [smallvec](https://github.com/servo/rust-smallvec) crate, for optimized storage of small number of components.
Should it be behind feature flag, since in brings new dependency?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,6 +27,7 @@ hibitset = "0.1"
 mopa = "0.2"
 ticketed_lock = "0.1"
 tuple_utils = "0.2"
+smallvec = "0.4"
 
 pulse = { version = "0.5", optional = true }
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -13,6 +13,7 @@ extern crate hibitset;
 extern crate mopa;
 extern crate ticketed_lock;
 extern crate tuple_utils;
+extern crate smallvec;
 
 #[cfg(feature="parallel")]
 extern crate pulse;
@@ -29,7 +30,7 @@ pub use gate::Gate;
 pub use join::{Join, JoinIter};
 pub use storage::{AntiStorage, BTreeStorage, GatedStorage, HashMapStorage, InsertResult,
                   MaskedStorage, NullStorage, Storage, UnprotectedStorage, VecStorage,
-                  DenseVecStorage};
+                  DenseVecStorage, SmallVecStorage};
 pub use world::{Allocator, Component, CreateEntities, Entities, World};
 
 #[cfg(feature="parallel")]
--- a/src/storage.rs
+++ b/src/storage.rs
@@ -7,6 +7,7 @@ use std::ops::{Deref, DerefMut, Not};
 
 use fnv::FnvHasher;
 use hibitset::{BitSet, BitSetNot};
+use smallvec::{SmallVec, Array};
 
 use gate::Gate;
 use join::Join;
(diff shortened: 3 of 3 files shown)


torkleyy:
Cannot review properly because I'm using my phone, but isn't this unpractical because as soon as you have more entities it isn't useful anymore? Probably it should rather be some redirection storage.On May 24, 2017 4:04 PM, Dzmitry Malyshau <notifications@github.com> wrote:Thank you for the PR!
Anyone familiar with smallvec wants to review the unsafe code?

Object905:
@torkleyy , yes when placed more items it allocates. Also I can replace smallvec for [arrayvec](https://github.com/bluss/arrayvec) and make it panic instead of allocating when it exceeds inline size limit.

Object905:
@kvark , smallvec almost exactly repeats standart `Vec`, so I basically just copied implementation from `VecStorage` and it seems to work.

kvark:
@Object905 My main concern here is that I don't see it practical, like @torkleyy said. Our `VecStorage` may not be using lots of entries, but it surely allocates a lot, so there is no way `SmallVec` could help replacing it.

A different situation is with `DenseVecStorage`, where the actual data is packed. I think it would make sense to have `smallvec` backing it (actual data vector only, while keeping the redirection tables to be regular vectors). Hiding the whole thing (new `DenseVecStorage`) behind a feature gate also sounds good.

Would you want to rewrite the PR for `DenseVecStorage` (replacing the current one) instead of `Vec`?

torkleyy:
@Object905 To explain this a bit more:

On insertion, your storage tries to insert a component at an index `id`. Even though you don't need many components, you need a pretty big `Vec`, because there's an uninitialized element for every entity without that component.

This means that a `SmallVec` with 16 elements does not support a component for an entity with id `16` or greater.

The `DenseVecStorage` works by using two collections; the first one for an entity -> index relation and the second one gives you the component from a given index (the one you got from the first collection).

That being said, I'm not sure if `SmallVec` does provide any advantage here:

* fixed-size
* needs more space than `Vec`
* even though it's stored on stack usually, it ends up on the heap because we need to `Box` it.

One advantage might be that it is more cache-efficient (because it's more plain), but I'm not sure what impact this has in practice.

kvark:
Right, @torkleyy, so even if we port `DenseVecStorage`, we'll not get much speed-up from `smallvec`...

Object905:
@torkleyy thanks for explaining this, I just didn't know how's Vec used with indices in VecStorage, i thought that VecStorage behaves like DenseVecStorage and didn't quite understand it. @kvark yes I'd like to rewrite it to use smallvec, but i think that DenseVecStorage is ok as it is now, and SmallVecStorage just needs to be rewritten to use  redirection tables like in DenseVecStorage, and also hidden behind feature flag. I'll work on it soon.

torkleyy:
@Object905 Which benefits do you expect when using a `SmallVec`?

Object905:
@torkleyy I get the idea after reading [this article](https://gist.github.com/jFransham/369a86eff00e5f280ed25121454acec1#use-stack-based-variable-length-datatypes). Benefits are that elements are stored on the stack. Also there is similar to smallvec data structure [from Qt](http://doc.qt.io/qt-4.8/qvarlengtharray.html#details).

torkleyy:
@Object905 Okay, let's try it out then. They are not really stored on the stack in this situation, but they allow a flat structures, which is what I meant with

> One advantage might be that it is more cache-efficient

PR #152, opened 2017-05-28
Update to serde 1.0

Aceeri:
Updates to the latest serde version.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,8 +29,8 @@ shred = { git = "https://github.com/torkleyy/shred" }
 shred-derive = "0.1.2"
 tuple_utils = "0.2"
 
-serde = { version = "0.9.11", optional = true }
-serde_derive = { version = "0.9.11", optional = true }
+serde = { version = "1.0", optional = true }
+serde_derive = { version = "1.0", optional = true }
 
 [features]
 serialize = ["serde", "serde_derive"]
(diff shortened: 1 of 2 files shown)


kvark:
Please rebase

src/storage/mod.rs
```diff
@@ -323,7 +323,7 @@ pub enum MergeError {
 
 #[cfg(feature="serialize")]
 impl<'e, T, D> Storage<'e, T, D>
-    where T: Component + serde::Deserialize,
+    where T: Component + serde::Deserialize<'e>,
```
torkleyy: I think we have to lessen the lifetime here and do an `impl<'a, 'e, T, D> &'a Storage<'e, T, D>` because you would have to feed the [`Visitor`](https://docs.serde.rs/serde/de/trait.Visitor.html) with slices that live as long as the borrowed entities.

Cargo.toml
```diff
@@ -29,13 +29,13 @@ shred = { git = "https://github.com/torkleyy/shred" }
 shred-derive = "0.1.2"
 tuple_utils = "0.2"
 
-serde = { version = "0.9.11", optional = true }
-serde_derive = { version = "0.9.11", optional = true }
+serde = { version = "1.0.8", optional = true }
```
kvark: what's the reason for specifying "1.0.8" as opposed to just "1.0" here and for `serde_derive`?

Aceeri: Ah true, forgot that those were patches in semver.

PR #153, opened 2017-05-29
Expose generation id/Follow rust guidelines for getters

### LINKED ISSUE #148: Unique entity Id
Hi!

It would be great, to have something like 

```
impl Entity {
...
    /// Returns the unique id of the `Entity`.
    #[inline]
    pub fn get_uid(&self) -> i64 {
        (self.1).0 as i64 | ((self.0 as i64) << 32)
    }
...
```

Or:

```
impl Entity {
...
    /// Returns the generation of the `Entity`.
    #[inline]
    pub fn get_gen_id(&self) -> i32 {
        (self.1).0
    }
...
```

Because now, generation id field is private, and create component for storing unique id of every entity looks like overhead.

Aceeri:
Just adds a getter to the `Generation` for getting the id.

Also changes the getters to follow the rust guidelines.

fixes #148

### DIFF
--- a/src/world.rs
+++ b/src/world.rs
@@ -24,9 +24,9 @@ pub struct Allocator {
 impl Allocator {
     fn kill(&mut self, delete: &[Entity]) {
         for entity in delete {
-            self.alive.remove(entity.get_id());
-            self.raised.remove(entity.get_id());
-            let id = entity.get_id() as usize;
+            self.alive.remove(entity.id());
+            self.raised.remove(entity.id());
+            let id = entity.id() as usize;
             self.generations[id].die();
             if id < self.start_from.load(Ordering::Relaxed) {
                 self.start_from.store(id, Ordering::Relaxed);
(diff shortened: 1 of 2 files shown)


src/world.rs
```diff
@@ -297,6 +297,12 @@ impl Generation {
         Generation(v)
     }
 
+    /// Returns the id of the generation.
+    #[inline]
+    pub fn get_id(&self) -> i32 {
```
torkleyy: Getters in Rust shouldn't start with `get_` unless necessary. Would recommend renaming it to `generation()`.

torkleyy: Hold on, the title says it's about generations, but this is the entity id, right?

minecrawler: yes, it's what was discussed in #148

torkleyy: Oh, I'm sorry, I couldn't read the code properly first because I was using my phone. So the correct getter would be `.id()`. Additionally, the `Entity` getters should be renamed.

Aceeri: Ah true, Im too used to writing C getters.

PR #156, opened 2017-06-01
Add The Specs Book

torkleyy:
I'm aware of the fact that there are several locations where more information might be nice (especially the storages chapter), but it's a start. Also, it would be good if some native speaker could check this PR, I'm sure there are many, many grammar mistakes. Feel free to just report all of them (also stylistic issues). Thanks!


book/src/01_intro.md
```diff
@@ -0,0 +1,83 @@
+# Introduction
+
+Welcome to The Specs Book, an introduction into ECS and the Specs API.
+This book is targeted to beginners, taking you through all the difficulties
```
kvark: targeted *at*

book/src/01_intro.md
```diff
@@ -0,0 +1,83 @@
...
+
+The term **ECS** is a shorthand for Entity-component system. These are the three
+core concepts. Each **entity** is associated with some **components**. Those entities and
+components are processed by **systems**. This way, you have your data (components)
+completely separated from the behaviour (systems). The entity just logically
+groups components, so a `Velocity` is applied to the `Position` of the
+same entity.
+
+ECS is sometimes seen as a counterpart to Object-Oriented Programming. I wouldn't
+say that's a hundred percent true, but let me give you some comparision here.
```
kvark: "comparision" typo. Might be worth running through a cheker, e.g. in LibreOffice?

book/src/01_intro.md
```diff
@@ -0,0 +1,83 @@
...
+
+In OOP, your player might look like this (I've used Java for the example):
+
+```java,ignore
```
kvark: I don't think you need `ignore` here, since it's not the in-code documentation that is touched by rustdoc

book/src/01_intro.md
```diff
@@ -0,0 +1,83 @@
...
+public class Player extends Character {
+    private final Transform transform;
+    private final Inventory inventory;
+}
+```
+
+There are several limitations here:
+
+* There is no multiple inheritance, therefore OOP programming is limited to
```
kvark: multiple inheritance will raise an eyebrow of C++ programmers, who would (rightfully) argue that their OOP is different

book/src/01_intro.md
```diff
@@ -0,0 +1,83 @@
...
+  extending just `Character`; moreover, you have to think about "*Is* the player
```
kvark: "*is*"

book/src/01_intro.md
```diff
@@ -0,0 +1,83 @@
...
+  a collider or *has* it a collider?"
+* You cannot easily extend the player with modding; all the
```
kvark: do we really need line breaks in the book? any editor will do them automatically, and better

book/src/01_intro.md
```diff
@@ -0,0 +1,83 @@
...
+  attributes are hardcoded
```
kvark: missing "." at the end

book/src/01_intro.md
```diff
@@ -0,0 +1,83 @@
...
+* Imagine you want to add a NPC, which looks like this:
```
kvark: "an NPC" sounds better to me

book/src/01_intro.md
```diff
@@ -0,0 +1,83 @@
...
+```
+
+Now you have the stuff duplicated; you would have to write mostly identical code for
+your player and the NPC, even though e.g. they both share a transform.
+
+This is where the ECS comes into play: Components are *associated* to entities;
+you just insert some component, whenever you like. Also, you just define the functionality
+once: One System, taking a `Force` and a `Mass` - it produces a `Velocity`.
+
+In fact, an entity does not even own the components; it's just
```
kvark: It would be great to define an entity, conceptually, as a set of components

book/src/01_intro.md
```diff
@@ -0,0 +1,83 @@
...
+
+```rust,ignore
+struct Entity(u32, Generation);
+```
+
+where the first field is the id and the second one a generation, used to validate
```
kvark: "second one is the generation"

book/src/01_intro.md
```diff
@@ -0,0 +1,83 @@
...
+if the entity hasn't been deleted. The components are stored separately, which
+also has the advantage of being more cache efficient.
+
+## Where to use an ECS?
+
+In case you were looking for a general-purpose library for doing it
+the data-oriented way, I have to disappoint you; it's not.
```
kvark: "it's not" -> "there is none"

book/src/02_hello_world.md
```diff
@@ -0,0 +1,195 @@
...
+        use specs::Join;
+    
+        for position in position.join() {
+            println!("Hello, {:?}", &position);
+        }
+    }
+}
+```
+
+> There are many other types you can use as system
```
kvark: that ">" is driving GH highlighting crazy

book/src/02_hello_world.md
```diff
@@ -0,0 +1,195 @@
...
+```rust,ignore
+use specs::run_now;
+
+let hello_world = HelloWorld;
+run_now(hello_world, &mut world.res);
+```
+
+## Full example code
+
+Here the complete example of this chapter:
```
kvark: I think these should just be in `/examples/chapterXX/*`, checked by CI, and simply referenced from here

book/src/03_dispatcher.md
```diff
@@ -0,0 +1,187 @@
+# Dispatcher
+
+## When to use a `Dispatcher`
+
+The `Dispatcher` allows you to automatically parallelize
+system execution where possible. It requires a bit more planning
```
kvark: perhaps, add that it's "fork-join model"

book/src/03_dispatcher.md
```diff
@@ -0,0 +1,187 @@
...
+and may have a little bit more overhead, but it's pretty convenient,
+especially when you're building a pretty big game where you don't
```
kvark: double use of "pretty"

book/src/03_dispatcher.md
```diff
@@ -0,0 +1,187 @@
...
+        }
+    }
+```
+
+Now the `.join()` method also makes sense: it joins the two component
+storages, so that you either get no new element or a new element with
+both components, meaning that entities with only a `Position`, only
+a `Velocity` or none of them will be skipped. The `0.05` fakes the
+so called **delta time** which is the time needed for one frame.
+We have to hardcode it right now, because it's no component (it's the
```
kvark: "not a component"

book/src/03_dispatcher.md
```diff
@@ -0,0 +1,187 @@
...
+}
+
+struct UpdatePos;
+
+impl<'a> System<'a> for UpdatePos {
+    type SystemData = (ReadStorage<'a, Velocity>, 
+                       WriteStorage<'a, Position>);
+                       
+    fn run(&mut self, data: Self::SystemData) {
+            use specs::Join;
```
kvark: weird indentation

book/src/03_dispatcher.md
```diff
@@ -0,0 +1,187 @@
...
+        .build();
+        
+    dispatcher.dispatch(&mut world.res);
+}
+```
+
+---
+
+[The next chapter][c4] will be a really short chapter about `Resource`s,
+a way to share data between systems which do only exist once
```
kvark: "exist once" -> "exist independent of entities", or "not tied to entities"

book/src/04_resources.md
```diff
@@ -0,0 +1,73 @@
+# Resources
+
+This (short) chapter will explain the concept of resources, data
+which is shared between systems.
+
+First of all, when would you need resources? There's actually a great
+example in chapter 3, where we just faked the delta time when applying
```
kvark: make "chapter 3" a link?

book/src/05_storages.md
```diff
@@ -0,0 +1,78 @@
...
+
+Here a list of the storages with a short description and a link
+to the corresponding heading.
+
+|Storage Type         |Description               |Optimized for                 |
+|:-------------------:|--------------------------|------------------------------|
+| [`BTreeStorage`]    | Works with a `BTreeMap`  | no particular case           |
+| [`DenseVecStorage`] | Uses a redirection table | fairly often used components |
+| [`HashMapStorage`]  | Uses a `HashMap`         | rare components              |
+| [`NullStorage`]     | Can flag entities        | universally applicable       |
```
kvark: I find "universally applicable" a bit confusing, given that it's not really an alternative to others

book/src/05_storages.md
```diff
@@ -0,0 +1,78 @@
...
+## `HashMapStorage`
+
+This should be used for rare components, because it provides a lower insertion
+cost and is better packed together. You should not use it for frequently used
+components, because the hashing cost would definitely be noticeable.
+
+## `NullStorage`
+
+As already described in the overview, the `NullStorage` does itself
+only contain a user-defined ZST. Because a mask is added when iterating,
```
kvark: clarify "ZST"

book/src/06_system_data.md
```diff
@@ -0,0 +1,19 @@
...
+You want to create/delete entities from a system? There is
+good news for you. You can just use `specs::data::Entities`
+(which also is in prelude) to do that. It implements
+`SystemData` so just put it 
+
+> Don't confuse `specs::data::Entities` with `specs::entity::Entities`.
+  While the latter one is the actual resource, the first one is a type
+  definition for `Fetch<entity::Entities>`.
+
+Please not that you may never write to these `Entities`, so only
```
kvark: "not" -> "note"

book/src/01_intro.md
torkleyy: Okay, you're right. Should I add a note about the diamond problem then or just remove that one?

book/src/01_intro.md
torkleyy: Did run a checker, but apparently missed that one.

book/src/02_hello_world.md
torkleyy: We'd need, however, be careful to always point to the file locked on a specific commit because they won't match otherwise.

book/src/01_intro.md
torkleyy: Mine does not - at least not with the standard config. I often find it just more convenient.

book/src/01_intro.md
kvark: I don't think we want to go there, so let's just not mention it ;)

book/src/02_hello_world.md
kvark: Agreed. It would be great to also CI-test all the snippets in this book, but that's not strictly required, and we can do it as a follow-up.

torkleyy: Yeah, it would be really nice, but `rust-skeptic` is unusable at the moment.

PR #159, opened 2017-06-02
Add parameter names to anonymous function parameters

torkleyy:
Anonymous function parameters are deprecated as of RFC 1685.

### DIFF
--- a/src/storage/mod.rs
+++ b/src/storage/mod.rs
@@ -461,7 +461,7 @@ impl<T> PackedData<T> {
     }
 }
 
-/// Used by the framework to quickly join componets
+/// Used by the framework to quickly join components.
 pub trait UnprotectedStorage<T>: Debug + Sized {
     /// Creates a new `Storage<T>`. This is called when you register a new
     /// component type within the world.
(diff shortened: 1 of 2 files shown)


src/storage/mod.rs
```diff
@@ -461,15 +461,15 @@ impl<T> PackedData<T> {
...
+/// Used by the framework to quickly join components.
 pub trait UnprotectedStorage<T>: Debug + Sized {
     /// Creates a new `Storage<T>`. This is called when you register a new
     /// component type within the world.
     fn new() -> Self;
 
     /// Clean the storage given a check to figure out if an index
     /// is valid or not. Allows us to safely drop the storage.
-    unsafe fn clean<F>(&mut self, F) where F: Fn(Index) -> bool;
+    unsafe fn clean<F>(&mut self, f: F) where F: Fn(Index) -> bool;
```
Aceeri: Why not call it `function` to describe it better?

torkleyy: I personally use `f` to accept closures very often. I also saw it in other libraries, and in case somebody doesn't know, the trait bound isn't far

PR #169, opened 2017-06-08
ChangedStorage

Aceeri:
Extension on the discussions in gitter and https://github.com/slide-rs/specs/issues/89.

### DIFF
--- a/src/storage/storages.rs
+++ b/src/storage/storages.rs
@@ -4,7 +4,7 @@ use std::collections::BTreeMap;
 use std::marker::PhantomData;
 
 use fnv::FnvHashMap;
-use hibitset::BitSet;
+use hibitset::{BitSet, BitSetAnd, BitSetNot};
 
 use world::EntityIndex;
 use {DistinctStorage, Index, Join, UnprotectedStorage};
@@ -350,14 +350,11 @@ unsafe impl<T> DistinctStorage for VecStorage<T> {}
 /// Wrapper storage that stores modifications to components with a bitset along with
 /// comparing equality to a cached version.
 ///
-/// `ChangedStorage::maintain()` should be called after every modification of the storage
-/// to *maintain* the cache and flagging correctly. 
-///
 ///# Example Usage:
 /// 
 /// ```rust
 /// extern crate specs;
-/// use specs::prelude::*;
+/// use specs::{Component, ChangedStorage, Join, System, VecStorage, WriteStorage};
 /// 
 /// #[derive(PartialEq, Clone)]
 /// pub struct Comp(u32);
@@ -385,9 +382,6 @@ unsafe impl<T> DistinctStorage for VecStorage<T> {}
 ///         for comp in (&mut comps).join() {
 ///             // ...
 ///         }
-///         // `maintain` checks for equality between the cache and the actual components 
-///         // and filters out the components that did not change from the flagged ones.
-///         (&mut comps).open().1.maintain();
 ///
 ///         // Clears the changed storage every frame with this system.
 ///         (&mut comps).open().1.clear_flags();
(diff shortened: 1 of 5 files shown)


src/storage/tests.rs
```diff
@@ -451,19 +467,34 @@ mod test {
...
         w.register_with_id::<FlaggedCvec>(1);
         w.register_with_id::<FlaggedCvec>(2);
-        let mut s1: Storage<FlaggedCvec, _> = w.write_with_id(1);
-        let mut s2: Storage<FlaggedCvec, _> = w.write_with_id(2);
 
-        for i in 0..15 {
-            // Test insertion flagging
-            s1.insert(Entity::new(i, Generation::new(1)), i.into());
-            assert!(s1.open().1.flagged(Entity::new(i, Generation::new(1))));
+        let mut entities = Vec::new();
```
torkleyy: Why are you doing this complicated loop here? Wouldn't a simple `.collect()` be sufficient? Or given that you only use these entities once, no allocation at all?

src/storage/storages.rs
```diff
@@ -332,3 +334,191 @@ impl<'a, C, T: UnprotectedStorage<C>> Join for &'a mut FlaggedStorage<C, T> {
...
+        }
+    }
+    // Forwarding methods to `FlaggedStorage`.
+    /// Whether the component related to the entity was flagged or not.
+    pub fn flagged<E: EntityIndex>(&self, entity: E) -> bool {
+        self.inner.flagged(entity)
+    }
+    /// Clears the bitset for flagged components.
+    ///
+    /// Should be called at least once depending on where you want to reset the flags.
```
torkleyy: I think it should be always called at the beginning of the system which actually sets these components.

src/storage/storages.rs
```diff
@@ -332,3 +334,191 @@ impl<'a, C, T: UnprotectedStorage<C>> Join for &'a mut FlaggedStorage<C, T> {
...
+          U: UnprotectedStorage<C>,
+{
+    /// Maintains the flagging and cache of the `ChangedStorage`.
+    /// 
+    /// Checks for equality between the cache and the actual components data
+    /// to filter out the components that did not change, leaving only the modified ones
+    /// in the flagged bitset.
+    ///
+    /// Should be called whenever you care about flagged components or alternatively
+    /// whenever the component storage is modified.
```
torkleyy: This is very unclear; unless you know the implementation details, this line won't help users. Could you maybe include a usage example?

torkleyy:
Given that this is a rather big feature and needs to be carefully reviewed, we decided to wait for this until `0.9` is published.

src/storage/storages.rs
Aceeri: This method doesn't matter anymore.

src/storage/storages.rs
Aceeri: There are no right times to clear the flags, it depends on the context. If every system that sets these components to something clears them then the flags are completely useless to other systems for things like transforms. In other circumstances if you only want it to be flagged from a certain iteration then you can clear it before that iteration.

torkleyy:
@Aceeri I think you implemented a `TrackedStorage` in the past. Do you think we'd need both (`ChangedStorage` and `TrackedStorage`)?

Aceeri:
@torkleyy The `TrackedStorage` stuff I implemented was just `FlaggedStorage` with insertion/removal of components tracking. The `ChangedStorage` is still kind of a more "correct" flagged storage since it only flags when something was truly modified, but I'm unsure if that level of accuracy is worth the complexity still (considering the whole issue I had with component wrapper storages vs addon storage).

Aceeri:
This storage seems kind of niche, so I'm probably not going to be working on it for a while. If anyone else wants to pick this up and use it themselves it should work.

PR #171, opened 2017-06-09
Remove `Split` bound from parallel iterator impl

torkleyy:
This allows to iterate over a single component storage in parallel.

r? @WaDelma


torkleyy:
I'm sure I can proceed here without a review.

bors r+

WaDelma:
I am actually not sure why there was `Split` requirement to begin with...

PR #172, opened 2017-06-09
Formatting, improve comments and fix minor issues

torkleyy:


src/storage/mod.rs
```diff
@@ -260,11 +279,13 @@ impl<'e, T, D> Storage<'e, T, D>
                    "Attempt to get an unchecked entry from a storage: {:?} {:?}",
                    entry.original,
                    self as *const Storage<'e, T, D>);
+
         unsafe { self.data.inner.get(entry.id) }
     }
 }
 
-/// The status of an insert operation
+/// The status of an `insert()`ion into a storage operation.
```
WaDelma: In the documentations I have seen (including standard library) function are written without parentheses (in this case \`insert\`).

src/join.rs
```diff
@@ -179,44 +177,43 @@ impl<'a, J> JoinProducer<'a, J>
...
+                     // can be safely called from different threads with distinct indices.
 
-            // The `J::get` gets called for distinct indices because splitting happens by `BitProducer::split`
-            // which should ensure that all splitted parts are distinct.
+                     // The indices here are guaranteed to be distinct because of the fact
+                     // that the bit set is split.
+                     J::get(&mut *values.get(), idx)
+                 });
 
-            // TODO: There can be multiple mutable references to storage and thus this is not safe, right?
```
WaDelma: I think the reasoning why creating `&mut` here is legal should be added.

src/storage/mod.rs
```diff
@@ -20,7 +20,26 @@ pub mod storages;
...
-/// A trait that guarantees that mutable storage getter can be used on distinct indices from different threads.
+/// This is a marker trait which requires you to uphold the following guarantee:
+///
+/// > Multiple threads may call `get_mut()` with distinct indices without causing
+/// > undefined behavior.
+///
+/// This for example valid for `Vec`:
+///
+/// ```rust
+/// vec![1, 2, 3];
```
WaDelma: Maybe add example using `VecStorage`?

tests/tests.rs
```diff
@@ -362,24 +380,28 @@ fn par_join_two_components() {
...
-                } else if !second.load(Ordering::SeqCst) && int.0 == 2 && boolean.0 {
-                    second.store(true, Ordering::SeqCst);
-                } else {
-                    *error.lock().unwrap() = Some((int.0, boolean.0));
-                }
-            });
+            (&int, &boolean)
+                .par_join()
+                .for_each(|(int, boolean)| if !first.load(Ordering::SeqCst) && int.0 == 1 &&
+                                              !boolean.0 {
```
WaDelma: This formating looks odd to me...

src/storage/mod.rs
torkleyy: I's okay both ways I guess.

src/storage/mod.rs
torkleyy: I don't think it's necessary for the explanation.

WaDelma: Okay. I just thought it bit jarring to see the example just contain creation of `Vec`

PR #181, opened 2017-06-10
Restructure type reexports

torkleyy:
* No crate root types, all grouped into modules
* `prelude` module reexports only `Join` and `ParJoin`

### DIFF
--- a/.rustfmt.toml
+++ b/.rustfmt.toml
@@ -1,0 +1,1 @@
+reorder_imported_names = true
--- a/README.md
+++ b/README.md
@@ -33,13 +33,8 @@ Unlike most other ECS libraries out there, it provides
 ## Example
 
 ```rust
-extern crate specs;
-
-use specs::prelude::*;
-
 // A component contains data
 // which is associated with an entity.
-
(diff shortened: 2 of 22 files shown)


kvark:
I like the approach, but it's hard to close the eyes on degraded ergonomics here. E.g. given a type the user might need, it's hard to tell which submodule it comes from. I think submodule includes make sense in very few cases, like
  - when there are name collisions between modules
  - when module itself gives unique meaning to the included type, e.g. `specs::storages::Vec` as opposed to `specs::VecStorage`.

I don't think we fall under any of these cases, although my list is unlikely complete:)
Anyhow, do you see any issues with re-exporting all the types from root and making the modules private?

torkleyy:
> do you see any issues with re-exporting all the types from root and making the modules private?

There's one name collision (`data::Entities` and `entity::Entities`), but otherwise I'd like to have all of them in the root.

kvark:
Let's do it then, and rename one of those (which is used less?) to avoid collision.

kvark:
Thanks!
bors r+

PR #191, opened 2017-06-12
Missing Send / Sync and ParJoin for CheckStorage

### LINKED ISSUE #190: CheckedStorage cannot be used with par_iter
Because `<CheckedStorage as Join>::Value` is `*const Storage<'e, T, D>` it cannot be used with `par_iter` as it isn't `Send`.

There is same problem with `Entry`, but you can implement `Send` for it.

torkleyy:
@WaDelma Sorry for the inconvenience!

Fixes #190

### DIFF
--- a/tests/tests.rs
+++ b/tests/tests.rs
@@ -451,3 +451,25 @@ fn par_join_many_entities_and_systems() {
                n);
     }
 }
+
+fn par_join_check_storage() {
+    use rayon::iter::ParallelIterator;
+
+    struct Sys;
+
+    let mut world = create_world();
+
+    impl<'a> System<'a> for Sys {
+        type SystemData = (WriteStorage<'a, CompInt>);
+
+        fn run(&mut self, data: Self::SystemData) {
+            let mut ints = data;
+            (ints.check())
+                .par_join()
+                .for_each(|c_i| ints.get_mut_unchecked(&mut c_i).0 += 1);
+        }
+    }
+
+    let mut dispatcher = DispatcherBuilder::new().add(Sys, "", &[]).build();
+    dispatcher.dispatch(&mut world.res);
+}
--- a/src/storage/check.rs
+++ b/src/storage/check.rs
@@ -1,4 +1,5 @@
-
+use std::cmp::PartialEq;
+use std::fmt::{Debug, Formatter, Result as FmtResult};
 use std::marker::PhantomData;
 use std::ops::{Deref, DerefMut};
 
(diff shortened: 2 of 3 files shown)


WaDelma:
I don't think that's enough: Doesn't the `CheckedStorage` still need `ParJoin` implementation? Also the `Value` is still un-`Send`.

WaDelma:
Should also probably add test for it... Should we consider [compiletest](https://github.com/laumann/compiletest-rs)?

torkleyy:
Why should we need compiletest? I think a simple test would be sufficient.

WaDelma:
Yeah. It just feels odd to have normal test that basically just tests that something compiles.

Aceeri:
Just know that the `CheckStorage` never dereferences the pointer, only uses it for an equality check so it is completely safe to implement Sync/Send as long as that never occurs.

tests/tests.rs
```diff
@@ -451,3 +451,25 @@ fn par_join_many_entities_and_systems() {
...
+    let mut world = create_world();
+
+    impl<'a> System<'a> for Sys {
+        type SystemData = (WriteStorage<'a, CompInt>);
+
+        fn run(&mut self, data: Self::SystemData) {
+            let mut ints = data;
+            (ints.check())
+                .par_join()
+                .for_each(|c_i| ints.get_mut_unchecked(&mut c_i).0 += 1);
```
torkleyy: There seems to be some lifetime issue going on here, I'll try to track it down later.

WaDelma: If you switch `ints.get_mut_unchecked(&mut c_i).0 += 1` to `ints.get_unchecked(&c_i);` it works fine.
If you change it to `&mut ints;` then it errors: 'cannot borrow data mutably in a captured outer variable in an `Fn` closure', because `for_each` takes `Fn` closure as parameter.
If you assign `&mut ints` to variable and try to capture it, it will 'cannot borrow `ints` as immutable because it is also borrowed as mutable' (as it should).

I think the `get_mut_unchecked` should be able to take `&self` as parameter as `&mut Entry` and it's lifetime should ensure correctness.

This would force to wrap `data` field inside `Storage` to `UnsafeCell` (or `inner` inside `MaskedStorage`)...

WaDelma: Actually it wouldn't be safe, right? Because it would allow to write through read only storage...

Aceeri: Well, the only way to write to a storage is to have the *mutable* storage and then ask for the component with the `Entry`. So no, calling `get_mut_unchecked` on an immutable storage would give a compiler error.

WaDelma: Well that's how it's currently. What I was proposing was to make `get_mut_unchecked` take `&self` instead of `&mut self` which would make it possible to use it on read only storage.

torkleyy:
Closing in favor of #203

PR #192, opened 2017-06-12
Custom derive for Component trait

ebkalderon:
This PR introduces an optional `specs_derive` crate containing a custom derive macro for defining new components in a less verbose way. The ergonomics loosely resemble the [`derivative`](https://github.com/mcarton/rust-derivative) crate.

### Before

```rust
#[derive(Debug)]
struct Pos(f32, f32, f32);

impl Component for Pos {
    type Storage = VecStorage<Pos>;
}
```

### After

```rust
#[derive(Component, Debug)]
struct Pos(f32, f32, f32);
```

The macro will store components in `VecStorage`s by default. To specify a different storage type, you may use the `#[component]` attribute.

```rust
#[derive(Component, Debug)]
#[component(HashMapStorage)]
struct Pos(f32, f32, f32);
```

I also included a revised `basic.rs` example called `basic_derive.rs` to demonstrate its usage, but this can probably be omitted from the PR if people ultimately prefer implementing `Component` the explicit way.

**EDIT**: Use `DenseVecStorage` by default.


specs_derive/src/lib.rs
```diff
@@ -0,0 +1,46 @@
...
+        .and_then(|attr| match *attr {
+            NestedMetaItem::MetaItem(ref item) => Some(item),
+            _ => None,
+        })
+        .and_then(|attr| match *attr {
+            MetaItem::Word(ref ident) => Some(ident),
+            _ => None,
+        })
+        .cloned()
+        .unwrap_or(Ident::new("::specs::VecStorage"));
```
torkleyy: I'm not sure if `VecStorage` should be the default. Maybe `DenseVecStorage` would be better here?

specs_derive/src/lib.rs
```diff
@@ -0,0 +1,46 @@
...
+pub fn component(input: TokenStream) -> TokenStream {
+    let s = input.to_string();
+    let ast = syn::parse_derive_input(&s).unwrap();
+    let gen = impl_component(&ast);
+    gen.parse().unwrap()
+}
+
+fn impl_component(ast: &MacroInput) -> Tokens {
+    let name = &ast.ident;
+    let (impl_generics, ty_generics, where_clause) = ast.generics.split_for_impl();
```
torkleyy: Didn't even know that sync has this method. Nice!

src/join.rs
```diff
@@ -1,7 +1,7 @@
 use std;
 use std::cell::UnsafeCell;
 
-use hibitset::{BitSetAnd, BitIter, BitSetLike, BitProducer};
+use hibitset::{BitIter, BitProducer, BitSetAnd, BitSetLike};
```
torkleyy: Whoops, forgot to reorder this one.

specs_derive/src/lib.rs
```diff
@@ -0,0 +1,46 @@
...
+#[macro_use]
+extern crate quote;
+extern crate syn;
+
+use proc_macro::TokenStream;
+use syn::{Ident, MacroInput, MetaItem, NestedMetaItem};
+use quote::Tokens;
+
+#[proc_macro_derive(Component, attributes(component))]
+pub fn component(input: TokenStream) -> TokenStream {
```
ebkalderon: Okay, will do.

specs_derive/src/lib.rs
ebkalderon: It's very nice! I learned about it from the [Procedural Macros chapter](https://doc.rust-lang.org/beta/book/first-edition/procedural-macros.html) of the Rust book.

specs_derive/src/lib.rs
ebkalderon: If that's indeed the case, I'll switch it to use `DenseVecStorage`.

specs_derive/src/lib.rs
ebkalderon: Will do!

ebkalderon:
@torkleyy Sure, I would be happy to update the README! But if we're using `#[derive(Component)]` by default, should I remove the `basic_derive.rs` example and update the existing examples to use the macro?

torkleyy:
@ebkalderon 

> should I remove the basic_derive.rs example and update the existing examples to use the macro?

Yes, I think we should do that. And we could just show both ways of implementing `Component` in the `basic` example to explain what it desugars to and then use the `#[derive(Component)]` everywhere we want.

ebkalderon:
@torkleyy Another question: would you like me to re-export these macros under the top-level `specs` crate, or would you prefer users explicitly specify `#[macro_use] extern crate specs_derive` in their projects?

torkleyy:
@ebkalderon Is it possible to reexport custom-derive macros?

ebkalderon:
@torkleyy I'm pretty sure you can. I can quickly check on my end to verify, once I'm done revising the PR.

ebkalderon:
@torkleyy I successfully re-exported `specs_derive` and compiled the examples with only the `specs` crate, but I can't do the same with `shred_derive` since `#[derive(SystemData)]` has a line which expands to `::shred::SystemData`. Oh well, at least `specs_derive` re-exports fine.

README.md
```diff
@@ -33,20 +33,19 @@ Unlike most other ECS libraries out there, it provides
 ## Example
 
 ```rust
+#[macro_use]
+extern crate specs;
```
torkleyy: Let's not have any `extern crate`s or imports in the README, it's irrelevant for the API usage.

ebkalderon:
Finally done! I have updated the README, book, examples, and and as many doc comments as I could find. Let me know if there are any nits to take care of.

specs_derive/src/lib.rs
```diff
@@ -0,0 +1,65 @@
...
+/// ```
+///
+/// The macro will store components in `DenseVecStorage`s by default. To specify
+/// a different storage type, you may use the `#[component]` attribute.
+///
+/// ```
+/// #[derive(Component, Debug)]
+/// #[component(HashMapStorage)]
+/// struct Pos(f32, f32, f32);
+/// ```
```
torkleyy: That's great, thanks :+1:

README.md
ebkalderon: I understand, I'll fix that.

examples/full.rs
```diff
@@ -2,39 +2,32 @@ extern crate rayon;
...
+use specs::{DispatcherBuilder, Entities, Entity, HashMapStorage, Join, ParJoin, ReadStorage,
+            System, VecStorage, World, WriteStorage};
 
 // -- Components --
 // A component exists for 0..n
 // entities.
 
-#[derive(Clone, Debug)]
+// `VecStorage` is meant to be used for components that are in almost every
+// entity tend to be highly contested.
```
ebkalderon: Whoops, forgot the word "and" here.

kvark:
I'm sorry but I don't find this useful enough. It's saving just a few lines of user code, at the cost of more lines in our code (`specs_derive`) but, more importantly, another layer of indirection to think about. I'd rather want people to implement `Component` and understand the storage types, which is arguably easier than applying magic attributes.

ebkalderon:
@kvark Regardless, people will need to understand the storage types in order to use the attribute properly; it's just presented more concisely from the user's side. But still, you raise some good points. In the end, it's your call! 😀

malleusinferni:
I'm using specs in a codebase with nearly 50 different component types (and climbing). I currently implement `Component` via a macro, just to cut down on the number of places where I can make a mistake. While I can't speak for other users, I would personally prefer to use `#[derive]` for this purpose. If this change is rejected, I'll roll my own implementation.

kvark:
@malleusinferni thanks for your input! What's wrong with the macro you are using now?
Also, do @csherratt, @Aceeri and @WaDelma have strong opinions?

csherratt:
My first instinct was exactly what @kvark said. It saves a few lines per component, and comes at a cost of some indirection which can confuse new users. That being said, I'm not strong in my position.

I think we could probably strike a balance, if @malleusinferni finds it useful. There are bound to be other users who want to use the more terse form. Since this must be published as a secondary crate anyhow, maybe we can show the old form in the examples. We can include a quick passage in the book/readme so that people can find this crate.

torkleyy:
Please don't forget: while it's great to be beginner-friendly (which we are; we got a book, nice examples and a Gitter channel to ask for help), the people dealing with these components and the Specs API will do that for many, many hours. The `basic` example explains what is implemented and AFAIK also tells which storage is used.

Using the right storage for the right situation isn't any different: now, beginners would just copy the example code, then they'd copy `#[derive(Component)] #[component(VecStorage)]`.

ebkalderon:
Just FYI, for all participants, `specs_derive` doesn't need to be published to crates.io given that it's a hard dependency and the macro is re-exported from within the `specs` crate.

Also, if you haven't noticed, the crate itself is pretty darn small (just one source file, with only two functions inside). Nearly all of the changes included in this PR are updating docs, examples, doc comments, the relevant book chapters, and the README. Both the book and the basic example clearly explain what the desugared API boils down to, and I've done my best to document all that I could.

Personally, I think that the reduced user-side boilerplate and the lessening of possible copy-paste errors is worth the effort, but those are just my two cents.

Aceeri:
Im impartial about this, but I have one concern: Does this support wrapper storages like `FlaggedStorage`?

torkleyy:
@Aceeri Maybe we could use `DenseVecStorage` as a default there?

WaDelma:
Doesn't crates.io demand that all dependencies are in crates.io?

I would like to have derive as an option as it basically doesn't remove any relevant information and just reduces boilerplate.

It's shame that it cannot have syntax like this though: `#[derive(Component(VecStorage))]`.

specs_derive/src/lib.rs
```diff
@@ -0,0 +1,65 @@
...
+    let ast = syn::parse_derive_input(&s).unwrap();
+    let gen = impl_component(&ast);
+    gen.parse().unwrap()
+}
+
+fn impl_component(ast: &MacroInput) -> Tokens {
+    let name = &ast.ident;
+    let (impl_generics, ty_generics, where_clause) = ast.generics.split_for_impl();
+
+    let storage = ast.attrs.first()
```
WaDelma: Doesn't this fail if there is more that one attribute and the `component` attribute isn't first?

kvark:
So, let me recap:
Positive: @ebkalderon, @WaDelma (?) 
Negative: @kvark
Balanced (needs partial code revert): @csherratt , @torkleyy
Abstained: @Aceeri

torkleyy:
Proposing compromise:

* Merge this crate and leave it in-tree
* Point to it in README, full example and book; code examples should implement it explicitly, only the book contains a single and short example of how to use `#[derive(Component)]`

torkleyy:
Okay, I think we found a consensus here (I proposed, @kvark upvoted and @csherratt suggested pretty much the same earlier). @ebkalderon Would it be okay for you to move on like this?

ebkalderon:
Yes! That sounds reasonable enough.

specs_derive/src/lib.rs
ebkalderon: Whoops, you're right! I forgot to change that while I was refactoring.

ebkalderon:
Okay, I've finally had some spare time to rebase and update this PR to address the concerns raised. Please feel free to review!

benches/parallel.rs
```diff
@@ -2,15 +2,15 @@
...
 extern crate specs;
 
 extern crate test;
 
 use cgmath::Vector2;
 use rand::thread_rng;
-use specs::{Component, DenseVecStorage, DispatcherBuilder, Entities, Entity, Fetch,
-            HashMapStorage, Join, NullStorage, ReadStorage, RunningTime, System, VecStorage,
-            World, WriteStorage};
+use specs::{DenseVecStorage, DispatcherBuilder, Entities, Entity, Fetch, HashMapStorage, Join,
```
torkleyy: We still need that export.

benches/world.rs
```diff
@@ -3,11 +3,12 @@
 extern crate cgmath;
 extern crate rand;
 extern crate rayon;
+#[macro_use]
```
torkleyy: Not necessary anymore

benches/world.rs
```diff
@@ -3,11 +3,12 @@
...
 extern crate specs;
 
 extern crate test;
 
-use specs::{Component, HashMapStorage, Join, ParJoin, VecStorage, World};
+use specs::{HashMapStorage, Join, ParJoin, VecStorage, World};
```
torkleyy: Missing import

book/src/02_hello_world.md
```diff
@@ -20,7 +21,7 @@ extern crate specs;
 Let's start by creating some data:
 
 ```rust,ignore
-use specs::{Component, VecStorage};
+use specs::VecStorage;
```
torkleyy: Missing import

book/src/02_hello_world.md
```diff
@@ -43,10 +44,30 @@ impl Component for Velocity {
 }
 ```
 
-These will be our components, stored in a `VecStorage`
-(see [the storages chapter][sc] for more details), so we can associate some data
-with an entity. Before doing that, we need to create a world, because this is
-where the storage for all the components is located.
+These will be our two component types. Optionally, `specs` includes a convenient
```
torkleyy: I'm not sure. I think we agreed we wouldn't want `#[derive]` shown in the book - except in one dedicated chapter.

book/src/02_hello_world.md
```diff
@@ -137,6 +157,7 @@ hello_world.run_now(&world.res);
 Here the complete example of this chapter:
 
 ```rust,ignore
+<<<<<<< HEAD
```
torkleyy: I think this was left when rebasing.

book/src/03_dispatcher.md
```diff
@@ -102,6 +102,7 @@ dispatcher.dispatch(&mut world.res);
 Here the code for this chapter:
 
 ```rust,ignore
+<<<<<<< HEAD
```
torkleyy: Same here.

src/lib.rs
```diff
@@ -167,17 +177,22 @@
 //!
 //!
 
+#[allow(unused_imports)]
+#[macro_use]
+extern crate specs_derive;
```
torkleyy: I don't think we need to include it here.

src/lib.rs
```diff
@@ -187,16 +202,17 @@ extern crate serde_derive;
...
-pub use shred::{AsyncDispatcher};
-
 pub use storage::{BTreeStorage, CheckStorage, DenseVecStorage, DistinctStorage, FlaggedStorage,
                   HashMapStorage, InsertResult, NullStorage, ReadStorage, Storage,
                   UnprotectedStorage, VecStorage, WriteStorage};
 pub use world::{Component, CreateIter, CreateIterAtomic, EntitiesRes, Entity, EntityBuilder,
                 Generation, LazyUpdate, World};
 
+#[cfg(not(target_os = "emscripten"))]
+pub use shred::{AsyncDispatcher};
```
torkleyy: Why move the re-export?

benches/world.rs
ebkalderon: Thanks, I missed that.

benches/parallel.rs
ebkalderon: Good catch, thank you.

benches/world.rs
ebkalderon: Sorry, forgot to add it back in after rebasing.

book/src/02_hello_world.md
ebkalderon: Thank you, I'll fix that.

book/src/02_hello_world.md
ebkalderon: Whoops, you're right. I'll go remove that.

book/src/03_dispatcher.md
ebkalderon: Sorry, I messed up on that one.

src/lib.rs
ebkalderon: I was hoping to allow `specs_derive` to re-export the procedural macro to the user, meaning they don't need to add `extern crate specs_derive;` to their project. If you prefer that people add it explicitly as a separate crate to their projects, I can remove it.

src/lib.rs
ebkalderon: Sorry, I must have made a mistake while copy-pasting sections around during the rebase.

book/src/02_hello_world.md
ebkalderon: Do we need a dedicated chapter for this macro, though? It is mentioned here once as an optional thing, and then never used again in the book.

src/lib.rs
torkleyy: Not sure. cc @kvark

book/src/02_hello_world.md
torkleyy: No, we don't need to. If somebody wants it, it's not hard to do so.

ebkalderon: Sounds good.

src/lib.rs
torkleyy: Just discussed with @kvark, we decided not to reexport it here so it doesn't get included if it's unnecessary for the user.

ebkalderon: Okay, I'll take it out.

book/src/02_hello_world.md
```diff
@@ -12,6 +12,7 @@ specs = "0.9"
 And add this to your crate root (`main.rs` or `lib.rs`):
 
 ```rust,ignore
+#[macro_use]
```
torkleyy: No need for `#[macro_use]`.

torkleyy:
Hold on, please add an entry to the changelog.

src/lib.rs
```diff
@@ -191,6 +202,7 @@ pub use shred::{Dispatcher, DispatcherBuilder, Fetch, FetchId, FetchIdMut,
 #[cfg(not(target_os = "emscripten"))]
 pub use shred::{AsyncDispatcher};
 
+pub use specs_derive::*;
```
torkleyy: This line causes the build to fail.

book/src/02_hello_world.md
ebkalderon: Thanks, removed.

ebkalderon:
@torkleyy Done! It should build correctly now, and I've updated the changelog.

PR #199, opened 2017-06-17
Panic with a better message when a component hasn't been registered

### LINKED ISSUE #196: More helpful error message when forgotten to register a component
I added a new component to my project, but forgot to register it with ``World::register``. Then when I attempted to create an entity with that component, I was given the following error: "No resource with the given id"

I imagine this would be a common mistake for new users so perhaps it might be nice for usability to add a hint into that error message (Maybe something like: "No resource with the given id. Did you forget to register a component with `World::register<ComponentName>()`?").

pengowen123:
Fixes #196. Requires slide-rs/shred#42, so I think this should wait on that.


torkleyy:
New version (finally) published, please update to `shred = 0.4.3`.

pengowen123:
I updated `read_resource_with_id` and `write_resource_with_id` to panic with a better message (and the `shred` version. I tested this change with the following code:

```rust
extern crate specs;

use specs::Join;

struct FooSystem;
struct Resource(i32);

impl<'a> specs::System<'a> for FooSystem {
    type SystemData = specs::Fetch<'a, Resource>;

    fn run(&mut self, data: Self::SystemData) {}
}

fn main() {
    let mut world = specs::World::new();

    let mut dispatcher = specs::DispatcherBuilder::new().add(FooSystem, "foo", &[]).build();

    loop {
        dispatcher.dispatch(&mut world.res);
    }
}
```

This panics with the `shred` message instead of the new one. The backtrace shows the panic happening in `<shred::Fetch<'a, T> as shred::SystemData<'a>>::fetch`. While the methods I updated on `World` will have a better message, I think the most likely situation is the one in this example, where the new message won't help.

My first thought would be to add `try` variants to the `SystemData` trait. Have any other ideas?

torkleyy:
I'm not sure about it. The approach you're describing would mean that we have to make many changes. Let's figure out how to do it before you do any work.

torkleyy:
The `try_run` wouldn't really work out well, so I think we cannot have a better error message right now. Let's just improve the docs (state that it panics) and maybe add a note to the book.

pengowen123:
I didn't mean adding a method to `specs::System`, if that is what you are saying won't work. I was talking about adding a `try_fetch` method to the [SystemData trait][0] from `shred`. Maybe `fetch` could become a provided method that just calls `self.try_fetch().expect()` with a default panic message (if this won't work, where do you want those docs?).

[0]: https://docs.rs/shred/0.4.3/shred/trait.SystemData.html#tymethod.fetch

pengowen123:
I tried to add `try_fetch` to `SystemData` and switch to using it as many places as possible, but I don't think it will work well. `specs` simply uses the dispatcher from `shred`, so there is no place to use `try_fetch().expect()` with a `specs`-specific panic message like I did with components.

An option consistent with the `try_fetch` methods would be a `Dispatcher::try_dispatch`, but there are two problems with it. First, it feels like a large addition only for better error messages. Second, due to using `Option` rather than `Result`, a lot of information is lost by the type the `None` propagates to the caller of `try_dispatch`. Unless proper error handling is added as well, error messages won't be helped much because `try_dispatch` returning `None` doesn't say whether it was a missing resource or an unregistered component.

I think adding `try_dispatch` is a large enough feature that it should be considered separately if someone has a use for it. If it happens, then proper error handling may be worth considering. For this pr though, I will just make a note in docs and book like you asked.

pengowen123:
I added some notes to the book and docs about systems panicking when a resource or component doesn't exist. Could you look over these notes and see whether they look okay?

src/world.rs
```diff
@@ -369,11 +374,16 @@ impl Generation {
 /// the borrowing rules of Rust (multiple reads xor one write),
 /// you will get a panic.
 ///
-/// # Component / Resource ids
+/// # Component / Resources
```
torkleyy: I think we should leave this as "Component / Resource ids"

src/world.rs
```diff
@@ -369,11 +374,16 @@ impl Generation {
...
 ///
 /// Components and resources may, in addition to their type, be identified
 /// by an id of type `usize`. The convenience methods dealing
 /// with components assume that it's `0`.
+///
+/// If a system attempts to access a component/resource that has not been
+/// registered/added, it will panic when run. Add all components before
+/// running any systems with `World::register`. Also add all resources
```
torkleyy: Maybe move the "with World::register" after "Add all components"?

Issue #155, opened 2017-06-01
Component/System/Resource Grouping


Aceeri:
Grouping systems together through a procedural macro that would allow calling a method with each type of component/system/resource. This would allow for static dispatch of serialization or anything dealing with groups of components, systems, or resources (like configurations).

### Reasons
**Ergonomic registering of components, systems, and resources**
While registering components by hand takes dozens of lines depending on the size of the project and its dependencies, registering a group would take a single line of code:
```rust
world.register_components::<ComponentsGroup>();
world.register_systems::<SystemsGroup>();
world.register_resources::<ResourcesGroup>();
```

**Modularization of dependencies**
External users of specs will likely not be using barebones specs and only their own components, especially in the future. For example, Amethyst provides `Transform`, `Renderable`, `Parent`, and more. Currently either the third party would have to make a special registering method to register all their components together or leave it up to the user to register those themselves, which can be a very tedious task when given a large amount of components.

Similarly without a component group, serialization would require manually plucking component storages and serializing them manually (as it is currently not possible to get a trait object of a `Serialize/Serializer/Deserialize/Deserializer` from `serde`.

It is very likely that a third party would make a macro for similar tasks to avoid a lot of boilerplate and centralizing it to a singular macro would be extremely beneficial rather than having the user juggle different implementations of macros/methods for registering and serialization.

### Implementation Details
Currently the groups are just defined by two traits:
```rust
/// Group of components. Can be subgrouped into other component groups.
pub trait ComponentGroup {
    /// Components defined in this group, not a subgroup.
    fn local_components() -> Vec<&'static str>;
    /// Components defined in this group along with subgroups.
    fn components() -> Vec<&'static str>;
    /// Subgroups included in this group.
    fn subgroups() -> Vec<&'static str>;
    /// Registers the components into the world.
    fn register(&mut World);
}

/// Group of serializable components.
#[cfg(feature="serialize")]
pub trait SerializeGroup: ComponentGroup {
    /// Serializes the group of components from the world.
    fn serialize_group<S: Serializer>(world: &World, serializer: S) -> Result<S::Ok, S::Error>;
    /// Helper method for serializing the world.
    fn serialize_subgroup<S: Serializer>(world: &World, map: &mut S::SerializeMap) -> Result<(), S::Error>;
    /// Deserializes the group of components into the world.
    fn deserialize_group<D: Deserializer>(world: &mut World, entities: &[Entity], deserializer: D) -> Result<(), D::Error>;
[...]

PR #201, opened 2017-06-18
Component Grouping

Aceeri:
fixes https://github.com/slide-rs/specs/issues/155

Restrutures into a workspace of `specs` and `specs_derive`.


specs/README.md
```diff
@@ -0,0 +1,71 @@
+# specs
```
torkleyy: It seems your rebase wasn't quite correct, this is the old README.

specs_derive/src/component_group.rs
```diff
@@ -0,0 +1,385 @@
...
+    // Normal group methods.
+    let default = quote! {
+        fn local_components() -> Vec<&'static str> {
+            vec![ #( stringify!(#comp_ty), )* ]
+        }
+        fn components() -> Vec<&'static str> {
+            #[allow(unused_mut)]
+            let mut list = #name::local_components();
+            #(
+                for component in #subgroup_ty::components() {
```
WaDelma: Can't this use `extend` instead of manual loop?

specs/src/group.rs
```diff
@@ -0,0 +1,150 @@
...
+    /// Registers the components into the world.
+    fn register(&mut World);
+}
+
+/// Group of serializable components.
+#[cfg(feature="serialize")]
+pub trait SerializeGroup: ComponentGroup {
+    /// Serializes the group of components from the world.
+    fn serialize_group<S: Serializer>(world: &World, serializer: S) -> Result<S::Ok, S::Error>;
+    /// Helper method for serializing the world.
```
WaDelma: Do we need to expose these helpers in the trait or are they only called in trait implementation? If they are not needed anywhere else then these helpers are just implementation detail and should probably removed.

kvark:
Can we have the restructurization PR separate from the new functionality please?

WaDelma:
You have to use workspace to include the custom derive as it needs to be it's own crate.

Aceeri:
I mean, I could have another PR for it if we want, but it would either just be a workspace of "specs" and "specs_derive" (without functionality).

torkleyy:
@Aceeri Maybe you could structure it the same way @ebkalderon did in their PR?

Aceeri:
@torkleyy I feel it makes more sense to use a workspace given this is exactly the usecase for it (multiple crates in one repo).

torkleyy:
@Aceeri They also implemented it as a workspace, see https://github.com/slide-rs/specs/pull/192/files

specs_derive/README.md
```diff
@@ -0,0 +1,107 @@
+# Specs
```
torkleyy: I don't think we need a second readme here (or at least, don't copy the specs README and just show a small example of component groups).

src/world.rs
```diff
@@ -646,3 +660,67 @@ impl Default for World {
         }
     }
 }
+
+#[cfg(feature="serialize")]
+/// Structure used to serialize a world using a component group.
+pub struct WorldSerializer<'a, G> {
```
torkleyy: Do we really need a struct for this?

Aceeri: Yeah, if I implemented `Serialize` or `Deserialize` on the `World` itself, then it would make a new copy of the world.

torkleyy: I was talking about just a function actually.

Aceeri: Well, it ties in better with `serde`s traits and such this way rather than just have some special methods for it.

Aceeri: I don't really mind either way since it would still be possible if it was a function instead of a struct.

Aceeri:
Well uh, I have made a mistake so please ignore this for the next couple hours while I go get my backup.

Aceeri:
@torkleyy For the `WorldSerializer`, I just realized why I originally had it. Some `serde` libraries don't expose `Serializer` in their interfaces which is annoying, but the only way to serialize with them is to pass in a type that implements `Serialize`.

src/world.rs
```diff
@@ -780,3 +794,66 @@ mod tests {
...
+    entities: &'a [Entity],
+    phantom: PhantomData<G>,
+}
+
+#[cfg(feature="serialize")]
+impl<'a, G> WorldDeserializer<'a, G> {
+    /// Creates a new world deserializer out of a world and a list of entities.
+    ///
+    /// The list of entities will be used to merge into the component storages.
+    pub fn new(world: &'a mut World, entities: &'a [Entity]) -> WorldDeserializer<'a, G> {
```
torkleyy: I wonder if it really makes sense to pass a slice of entities. Is there any case where you want to deserialize components and insert them for existing entities? I mean can you really ensure this works reliable? Which brings me to the question if it should be allowed for an entity to not have a component of the component group.

torkleyy:
@Aceeri I see. It's up to you how you implement it then.

src/world.rs
Aceeri: > Is there any case where you want to deserialize components and insert them for existing entities? 

I was mainly thinking stuff like prefabs here, so a prefab could be deserialized into an existing entity.

> Which brings me to the question if it should be allowed for an entity to not have a component of the component group.

Well, if we disallowed that then every entity would have to have every component in that component group if we wanted to serialize the world properly.

torkleyy: > I was mainly thinking stuff like prefabs here, so a prefab could be deserialized into an existing entity.

Exactly, so a new entity has to be created.

Aceeri: What if you want to put in multiple prefabs for a single entity though?

Aceeri:
Closing this for now since I'm not sure if it is worth the complexity yet.

PR #203, opened 2017-06-20
CheckStorage soundness fixes

Aceeri:
Splits `CheckStorage` into `RestrictedStorage` and `CheckStorage` for different usecases.

`CheckStorage` will just not borrow the storage, however it does not protect against the original storage having its bitset modified, so no component or entry is given for unchecked access. `RestrictedStorage` returns an entry and a reference back to the restricted storage every iteration. This allows restricted access to the storage to get and possibly modify components without invalidating the bitset.

### DIFF
--- a/examples/full.rs
+++ b/examples/full.rs
@@ -85,6 +85,11 @@ impl<'a> System<'a> for SysCheckPositive {
     type SystemData = IntAndBoolData<'a>;
 
     fn run(&mut self, mut data: IntAndBoolData) {
+        
+        for (entry, restricted) in data.comp_int.restrict().join() {
+            let component = restricted.get_unchecked(&entry);
+            println!("{:?}: {:?}", entry, component);
+        }
         // Join merges the two component storages,
         // so you get all (CompInt, CompBool) pairs.
         for (ci, cb) in (&data.comp_int, &mut data.comp_bool).join() {
--- a/examples/full.rs
+++ b/examples/full.rs
@@ -50,7 +50,6 @@ struct Sum(usize);
 struct IntAndBoolData<'a> {
     comp_int: ReadStorage<'a, CompInt>,
     comp_bool: WriteStorage<'a, CompBool>,
-    entities: Entities<'a>,
 }
 
 #[derive(SystemData)]
@@ -86,19 +85,6 @@ impl<'a> System<'a> for SysCheckPositive {
     type SystemData = IntAndBoolData<'a>;
 
     fn run(&mut self, mut data: IntAndBoolData) {
-        
-        for (entity, (entry, restricted)) in (&*data.entities, &mut data.comp_bool.restrict_mut()).join() {
-            {
-                let component = restricted.get_unchecked(&entry);
-                println!("immutable: {:?}: {:?}", entry, component);
-            }
-
-            {
-                let mut component = restricted.get_mut_unchecked(&entry);
-                component.0 = !component.0;
-                println!("mutable: {:?}: {:?}", entry, component);
-            }
-        }
         // Join merges the two component storages,
         // so you get all (CompInt, CompBool) pairs.
         for (ci, cb) in (&data.comp_int, &mut data.comp_bool).join() {
--- a/examples/full.rs
+++ b/examples/full.rs
@@ -50,6 +50,7 @@ struct Sum(usize);
 struct IntAndBoolData<'a> {
     comp_int: ReadStorage<'a, CompInt>,
     comp_bool: WriteStorage<'a, CompBool>,
+    entities: Entities<'a>,
 }
 
 #[derive(SystemData)]
@@ -86,9 +87,17 @@ impl<'a> System<'a> for SysCheckPositive {
 
     fn run(&mut self, mut data: IntAndBoolData) {
         
-        for (entry, restricted) in data.comp_int.restrict().join() {
-            let component = restricted.get_unchecked(&entry);
-            println!("{:?}: {:?}", entry, component);
+        for (entity, (entry, restricted)) in (&*data.entities, &mut data.comp_bool.restrict_mut()).join() {
+            {
+                let component = restricted.get_unchecked(&entry);
+                println!("immutable: {:?}: {:?}", entry, component);
+            }
+
+            {
+                let mut component = restricted.get_mut_unchecked(&entry);
+                component.0 = !component.0;
+                println!("mutable: {:?}: {:?}", entry, component);
+            }
         }
         // Join merges the two component storages,
         // so you get all (CompInt, CompBool) pairs.
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -206,8 +206,8 @@ pub use shred::{Dispatcher, DispatcherBuilder, Fetch, FetchId, FetchIdMut,
 pub use shred::{AsyncDispatcher};
 
 pub use storage::{BTreeStorage, CheckStorage, DenseVecStorage, DistinctStorage, Entry, FlaggedStorage,
-                  HashMapStorage, InsertResult, MaskedStorage, NullStorage, ReadStorage, RestrictedStorage,
-                  Storage, UnprotectedStorage, VecStorage, WriteStorage, restrict_type};
+                  HashMapStorage, InsertResult, MaskedStorage, NormalRestriction, NullStorage, ParallelRestriction, ReadStorage, RestrictedStorage,
+                  Storage, UnprotectedStorage, VecStorage, WriteStorage};
 pub use world::{Component, CreateIter, CreateIterAtomic, EntitiesRes, Entity, EntityBuilder,
                 Generation, LazyUpdate, World};
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -207,7 +207,7 @@ pub use shred::{AsyncDispatcher};
 
 pub use storage::{BTreeStorage, CheckStorage, DenseVecStorage, DistinctStorage, Entry, FlaggedStorage,
                   HashMapStorage, InsertResult, MaskedStorage, NullStorage, ReadStorage, RestrictedStorage,
-                  Storage, UnprotectedStorage, VecStorage, WriteStorage};
+                  Storage, UnprotectedStorage, VecStorage, WriteStorage, restrict_type};
 pub use world::{Component, CreateIter, CreateIterAtomic, EntitiesRes, Entity, EntityBuilder,
                 Generation, LazyUpdate, World};
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -205,9 +205,9 @@ pub use shred::{Dispatcher, DispatcherBuilder, Fetch, FetchId, FetchIdMut,
 #[cfg(not(target_os = "emscripten"))]
 pub use shred::{AsyncDispatcher};
 
-pub use storage::{BTreeStorage, CheckStorage, DenseVecStorage, DistinctStorage, FlaggedStorage,
-                  HashMapStorage, InsertResult, MaskedStorage, NullStorage, ReadStorage, Storage,
-                  UnprotectedStorage, VecStorage, WriteStorage};
+pub use storage::{BTreeStorage, CheckStorage, DenseVecStorage, DistinctStorage, Entry, FlaggedStorage,
+                  HashMapStorage, InsertResult, MaskedStorage, NullStorage, ReadStorage, RestrictedStorage,
+                  Storage, UnprotectedStorage, VecStorage, WriteStorage};
 pub use world::{Component, CreateIter, CreateIterAtomic, EntitiesRes, Entity, EntityBuilder,
                 Generation, LazyUpdate, World};
 
--- a/src/storage/check.rs
+++ b/src/storage/check.rs
@@ -1,118 +1,158 @@
 
+use std::cell::UnsafeCell;
 use std::marker::PhantomData;
 use std::ops::{Deref, DerefMut, Not};
 
+use shred::Fetch;
 use hibitset::BitSet;
 
 use storage::{AntiStorage, MaskedStorage};
-use world::EntityIndex;
-use {Component, DistinctStorage, Index, Join, Storage, UnprotectedStorage};
+use world::{EntityIndex, EntitiesRes};
+use {Component, DistinctStorage, Entity, Index, Join, Storage, UnprotectedStorage};
 
-/// A storage type that iterates entities that have
-/// a particular component type, but does not return the
-/// component.
-pub struct CheckStorage<'a, T, D> {
-    bitset: BitSet,
-    // Pointer back to the storage the CheckStorage was created from.
-    original: *const Storage<'a, T, D>,
+/// Similar to a `MaskedStorage` and a `Storage` combined, but restricts usage
+/// to only getting and modifying the components. That means nothing that would
+/// modify the inner bitset so the iteration cannot be invalidated. For example,
+/// no insertion or removal is allowed.
+pub struct RestrictedStorage<'rf, 'st: 'rf, T: Component> {
+    bitset: &'rf BitSet,
+    data: &'rf T::Storage,
+    entities: &'rf Fetch<'st, EntitiesRes>,
 }
 
-impl<'a, 'e, T, D> Join for &'a CheckStorage<'e, T, D> {
-    type Type = Entry<'a, 'e, T, D>;
-    type Value = *const Storage<'e, T, D>;
-    type Mask = &'a BitSet;
-
-    fn open(self) -> (Self::Mask, Self::Value) {
-        (&self.bitset, self.original)
+impl<'rf, 'st, T> RestrictedStorage<'rf, 'st, T>
+    where T: Component,
+{
+    pub fn get(&self, entity: Entity) -> Option<&T> {
+        if self.bitset.contains(entity.id()) && self.entities.is_alive(entity) {
+            Some(unsafe { self.data.get(entity.id()) })
+        } else {
+            None
+        }
     }
 
-    unsafe fn get(storage: &mut *const Storage<'e, T, D>, id: Index) -> Entry<'a, 'e, T, D> {
-        Entry {
+    pub fn get_unchecked(&self, entry: &Entry<'rf, T>) -> &T {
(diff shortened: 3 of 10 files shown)


src/storage/check.rs
```diff
@@ -1,111 +1,186 @@
...
-    phantom: PhantomData<&'a ()>,
+    pointer: *const Storage<'a, T, D>,
+}
+
+impl<'a, T, D> Entry<'a, T, D>
+    where T: Component,
+          D: 'a,
+{
+    #[inline]
+    fn assert_same_storage(&self, storage: &Storage<'a, T, D>) {
```
WaDelma: Would it be possible to use something like what [this](https://github.com/bluss/indexing) library does to get rid of these asserts? Just remembered the idea so I am not sure how feasible it would be.

Aceeri: Not sure... I don't really understand how it works.

WaDelma: Afaik it forces creation of new lifetime for the index and then ensures that that lifetimed index can only be used with right container. You actually need closure for it too. I didn't remember that was the case.

src/storage/check.rs
```diff
@@ -1,111 +1,41 @@
...
-use world::EntityIndex;
-use {Component, DistinctStorage, Index, Join, Storage, UnprotectedStorage};
+use {Component, Index, Join, Storage};
 
-/// A storage type that iterates entities that have
-/// a particular component type, but does not return the
-/// component.
-pub struct CheckStorage<'a, T, D> {
+/// Allows iterating over a storage without borrowing the storage itself.
+pub struct CheckStorage {
```
torkleyy: Looks good overall, but as far as I see it it requires a version bump, right?

src/storage/restrict.rs
```diff
@@ -0,0 +1,249 @@
...
+}
+
+impl<'rf, 'st, B, T, R> RestrictedStorage<'rf, 'st, B, T, R, restrict_type::Normal>
+    where T: Component,
+          R: BorrowMut<T::Storage>,
+          B: Borrow<BitSet>,
+{
+    /// Attempts to get the component related to the entity mutably.
+    ///
+    /// Functions similar to the normal `Storage::get_mut` implementation.
```
torkleyy: Maybe we should add a note that it only works for `restrict_type::Normal`?

src/storage/restrict.rs
```diff
@@ -0,0 +1,249 @@
...
+            bitset: &self.data.mask,
+            data: &self.data.inner,
+            entities: &self.entities,
+            phantom: PhantomData,
+        }
+    }
+
+    /// Builds a parallel `RestrictedStorage`, does not allow mutably getting other components
+    /// aside from the current iteration.
+    pub fn par_restrict<'rf>(&'rf self)
```
torkleyy: Do we really need to methods for immutable restriction storages? Because they both don't allow mutable access, I don't see how their functionality is different.

src/storage/restrict.rs
```diff
@@ -0,0 +1,249 @@
...
+            bitset: mask,
+            data: data,
+            entities: &self.entities,
+            phantom: PhantomData,
+        }
+    }
+}
+
+/// An entry to a storage.
+pub struct Entry<'rf, T>
```
torkleyy: At this point, we should evaluate which of `Send` / `Sync` this type conforms to.

Aceeri: I think it is safe for it to both be `Send/Sync`? It lives as long as the reference to the storage it belongs to, and only works in a `Join`, however I don't see how much use it has in those scenarios.

src/storage/restrict.rs
Aceeri: Probably not necessary.

examples/full.rs
```diff
@@ -50,6 +50,7 @@ struct Sum(usize);
 struct IntAndBoolData<'a> {
     comp_int: ReadStorage<'a, CompInt>,
     comp_bool: WriteStorage<'a, CompBool>,
+    entities: Entities<'a>,
```
torkleyy: I don't think we need this.

src/storage/restrict.rs
```diff
@@ -0,0 +1,243 @@
...
+use hibitset::BitSet;
+
+use storage::MaskedStorage;
+use world::{EntityIndex, EntitiesRes};
+use {Component, Entity, Index, Join, Storage, UnprotectedStorage};
+
+/// Generic flags for the restricted storage to determine
+/// the type of restriction.
+pub mod restrict_type {
+    pub struct Normal;
```
torkleyy: docs missing

examples/full.rs
Aceeri: Yeah, that was mainly for quick testing reasons.

src/lib.rs
```diff
@@ -190,9 +190,9 @@ pub use shred::{ Dispatcher, DispatcherBuilder, Fetch, FetchId, FetchIdMut,
 #[cfg(not(target_os = "emscripten"))]
 pub use shred::{AsyncDispatcher};
 
-pub use storage::{BTreeStorage, CheckStorage, DenseVecStorage, DistinctStorage, FlaggedStorage,
-                  HashMapStorage, InsertResult, NullStorage, ReadStorage, Storage,
-                  UnprotectedStorage, VecStorage, WriteStorage};
+pub use storage::{BTreeStorage, CheckStorage, DenseVecStorage, DistinctStorage, Entry, FlaggedStorage,
+                  HashMapStorage, InsertResult, NullStorage, ReadStorage, restrict_type, RestrictedStorage,
```
torkleyy: Let's move `restrict_type` to the end of the list.

src/storage/tests.rs
```diff
@@ -408,24 +408,26 @@ mod test {
...
             {
-                s1.get_mut_unchecked(&mut entry);
+                restricted.get_mut_unchecked(&entry);
             }
         }
     }
 
     #[test]
-    #[should_panic]
+    //#[should_panic]
```
torkleyy: err..is there anything wrong with the check or did you fix it already?

Aceeri: Forgot about that, just needed to add insertions to the `s2` storage.

src/storage/restrict.rs
```diff
@@ -0,0 +1,251 @@
...
+/// modify the inner bitset so the iteration cannot be invalidated. For example,
+/// no insertion or removal is allowed.
+pub struct RestrictedStorage<'rf, 'st: 'rf, B, T, R, RT>
+    where T: Component,
+          R: Borrow<T::Storage> + 'rf,
+          B: Borrow<BitSet> + 'rf,
+{
+    bitset: B,
+    data: R,
+    entities: &'rf Fetch<'st, EntitiesRes>,
```
torkleyy: Could just use `Entities` here.

src/storage/mod.rs
```diff
@@ -18,6 +19,7 @@ use shred::Fetch;
 use {Component, EntitiesRes, Entity, Index, Join, ParJoin};
 
 mod check;
+mod restrict;
```
torkleyy: Could you move the module declaration down by one, please?

Aceeri: Ah sorry, forgot about the alphabetical order for modules.

Aceeri:
@torkleyy I think this is ready for merging, unless you want to wait before making the version bump.

torkleyy:
I'm not sure when we should make the version bump. @kvark What do you think?

kvark:
I don't think there is any reason to wait till the version bump. We should always be able to make extra patches to 0.9 when needed by working in a branch. Thus, I think the PR is good in the current shape, although I haven't looked in detail.

torkleyy:
I'm ready to merge this one, but I'd like to have another review.

src/storage/restrict.rs
```diff
@@ -0,0 +1,250 @@
...
+}
+
+impl<'st, T, D> Storage<'st, T, D>
+    where T: Component,
+          D: Deref<Target = MaskedStorage<T>>,
+{
+    /// Builds an immutable `RestrictedStorage` out of a `Storage`. Allows restricted
+    /// access to the inner components without allowing invalidating the
+    /// bitset for iteration in `Join`.
+    pub fn restrict<'rf>(&'rf self)
```
kvark: Is there a point in this read-only restricted storage? I'd assume that a regular `MaskedStorage` is equivalent since it doesn't allow adding/removing the components (and thus modifying the bitmask) either.

src/storage/check.rs
```diff
@@ -1,118 +1,49 @@
...
-use world::EntityIndex;
-use {Component, DistinctStorage, Index, Join, Storage, UnprotectedStorage};
+use {Component, DistinctStorage, Index, Join, Storage};
 
-/// A storage type that iterates entities that have
-/// a particular component type, but does not return the
-/// component.
-pub struct CheckStorage<'a, T, D> {
+/// Allows iterating over a storage without borrowing the storage itself.
+pub struct CheckStorage {
```
kvark: What's a good use case for this new `CheckStorage`?

Aceeri: If you need to insert/remove components from the storage, this allows you to do that with a single loop rather than loop over all of them, find the ones you need to remove, store them in a `Vec`, then loop over those and remove them.

src/storage/restrict.rs
Aceeri: There isn't too much difference, this one just defers the "getting" of the component to when the user wants to get it, I'm not sure how useful that is with just being able to get an immutable reference though, so it probably doesn't have much of a point being immutable.

src/storage/check.rs
OvermindDL1: > If you need to insert/remove components from the storage, this allows you to do that with a single loop rather than loop over all of them, find the ones you need to remove, store them in a Vec, then loop over those and remove them.

What is the performance difference with this single loop over the loop/deleteloop style?

kvark:
Alright, the use case for `CheckStorage` is a bit weak, but it's there.
Read-only `RestrictedStorage` doesn't seem to have one, so I'd prefer this concept limited to mutability only.

Aceeri:
Alright, removed the immutable restricted storage and just made the `restrict_mut` into `restrict` instead.

src/storage/restrict.rs
```diff
@@ -0,0 +1,231 @@
...
+    where T: Component,
+          R: BorrowMut<T::Storage>,
+          B: Borrow<BitSet>,
+{
+    /// Attempts to get the component related to the entity mutably.
+    ///
+    /// Functions similar to the normal `Storage::get_mut` implementation.
+    ///
+    /// Note: This only works if this is a parallel and mutable `RestrictedStorage`.
+    /// You can get one by using the `par_restrict_mut` method instead of the normal
```
kvark: documentation is outdated

src/storage/restrict.rs
```diff
@@ -0,0 +1,231 @@
...
+impl<'rf, 'st, B, T, R> RestrictedStorage<'rf, 'st, B, T, R, restrict_type::Normal>
+    where T: Component,
+          R: BorrowMut<T::Storage>,
+          B: Borrow<BitSet>,
+{
+    /// Attempts to get the component related to the entity mutably.
+    ///
+    /// Functions similar to the normal `Storage::get_mut` implementation.
+    ///
+    /// Note: This only works if this is a parallel and mutable `RestrictedStorage`.
```
kvark: Note says this is for parallel, but the actual implementation is for `Normal`. Where is the truth?

src/storage/restrict.rs
```diff
@@ -0,0 +1,231 @@
...
+
+use storage::MaskedStorage;
+use world::EntityIndex;
+use {Component, DistinctStorage, Entities, Entity, Index, Join, Storage, UnprotectedStorage};
+
+/// Generic flags for the restricted storage to determine
+/// the type of restriction.
+pub mod restrict_type {
+    /// Specifies that the `RestrictedStorage` cannot run in parallel.
+    pub struct Normal;
```
kvark: enums are better phantom types, and no need for a separate module, really.
```rust
enum NormalRestriction {}
enum ParallelRestriction {}
```

src/storage/restrict.rs
```diff
@@ -0,0 +1,231 @@
...
+    /// Specifies that the `RestrictedStorage` can run in parallel.
+    pub struct Parallel;
```
kvark: could you explain a bit on how these types help you exactly?

src/storage/restrict.rs
Aceeri: True, didn't know about never types before the asset management stuff.

src/storage/check.rs
Aceeri: @OvermindDL1 I'm not entirely sure? It would probably be faster if the compiler didn't optimize anything out, but in this case it probably is, so this is more for just ease of use and less trying to slide around the borrow checker.

src/storage/restrict.rs
Aceeri: I have to prevent `get_mut` on it when you use rayon, since that would mean you could get 2 mutable references to the same entity.

src/storage/restrict.rs
Aceeri: Ah, I probably meant to put "non-parallel" there.

src/storage/restrict.rs
Aceeri: Basically, I'm enforcing that you can do this:
```rust
(&comp_int, &mut comp_float.par_restrict())
    .par_join()
    .for_each(|(i, (entry, restricted))| {
        let f = restricted.get_mut_unchecked(&entry);
    });
```
But you can't do this:
```rust
(&comp_int, &mut comp_float.par_restrict())
    .par_join()
    .for_each(|(i, (entry, restricted))| {
        let f = restricted.get_mut(&entity); // Some entity stored somewhere.
    });
```

src/storage/restrict.rs
```diff
@@ -0,0 +1,231 @@
...
+          B: Borrow<BitSet> + 'rf,
+{
+    bitset: B,
+    data: R,
+    entities: &'rf Entities<'st>,
+    phantom: PhantomData<(T, RT)>,
+}
+
+
+unsafe impl<'rf, 'st: 'rf, B, T, R, ParallelRestriction> ParJoin for &'rf RestrictedStorage<'rf, 'st, B, T, R, ParallelRestriction>
```
kvark: I don't think you want `ParallelRestriction` in the `impl` block here

src/storage/restrict.rs
```diff
@@ -0,0 +1,231 @@
...
+    where T: Component,
+          R: BorrowMut<T::Storage> + 'rf,
+          B: Borrow<BitSet> + 'rf,
+{ }
+
+unsafe impl<'rf, 'st: 'rf, B, T, R, ParallelRestriction> ParJoin for &'rf mut RestrictedStorage<'rf, 'st, B, T, R, ParallelRestriction>
```
kvark: same here... what's the difference between these blocks anyway?

src/storage/restrict.rs
Aceeri: `ParallelRestriction` was the correct one, since it removes the `get_mut` method for the `RestrictedStorage`.

src/storage/restrict.rs
Aceeri: It does the same thing as the `ParJoin` implementation for `Storage`.

src/storage/restrict.rs
kvark: I read it as: you put `ParallelRestriction` into the `impl<...>` section, which defines a new local generic type with the same name as your phantom type. This is not what you want here.

Aceeri: Ah true, I must've done a `s/RT/ParallelRestriction` instead of just replacing the `RestrictedStorage` one.

src/storage/restrict.rs
```diff
@@ -0,0 +1,224 @@
...
+
+impl<'st, T, D> Storage<'st, T, D>
+    where T: Component,
+          D: DerefMut<Target = MaskedStorage<T>>,
+{
+    /// Builds a mutable `RestrictedStorage` out of a `Storage`. Allows restricted
+    /// access to the inner components without allowing invalidating the
+    /// bitset for iteration in `Join`.
+    pub fn restrict<'rf>(&'rf mut self)
+        -> RestrictedStorage<'rf, 'st, &'rf BitSet, T, &'rf mut T::Storage, NormalRestriction>
```
kvark: I think lifetime elision could be used here to omit `rf` entirely?

src/storage/restrict.rs
```diff
@@ -0,0 +1,224 @@
...
+}
+
+impl<'rf, 'st, B, T, R, RT> RestrictedStorage<'rf, 'st, B, T, R, RT>
+    where T: Component,
+          R: BorrowMut<T::Storage>,
+          B: Borrow<BitSet>,
+{
+    /// Gets the component related to the current entry without checking whether
+    /// the storage has it or not.
+    pub fn get_mut_unchecked(&mut self, entry: &Entry<'rf, T>) -> &mut T {
```
kvark: shouldn't this be unsafe?

Aceeri: This is actually safe since it requires an `Entry`, which can't live longer than the iteration.

src/storage/restrict.rs
```diff
@@ -0,0 +1,224 @@
...
+/// Specifies that the `RestrictedStorage` cannot run in parallel.
+pub enum NormalRestriction { }
+/// Specifies that the `RestrictedStorage` can run in parallel.
+pub enum ParallelRestriction { }
+
+/// Similar to a `MaskedStorage` and a `Storage` combined, but restricts usage
+/// to only getting and modifying the components. That means nothing that would
+/// modify the inner bitset so the iteration cannot be invalidated. For example,
+/// no insertion or removal is allowed.
+pub struct RestrictedStorage<'rf, 'st: 'rf, B, T, R, RT>
```
torkleyy: Could we have an example usage?

torkleyy:
@Aceeri Please rebase so we can merge this PR.

Aceeri:
@torkleyy Rebased so it should be fine now.

torkleyy:
Nice, thank you!

bors r+

PR #205, opened 2017-06-24
disable AsyncDispatcher use for emscripten

oussama:
I guess it is work in progress since you already implemented the dispatch_seq. it compiles to wasm & works for me after this fix.


torkleyy:
Ah, thanks! I didn't check whether or not Specs works with emscripten, but only shred.

bors r+

PR #206, opened 2017-06-25
Change component definitions to use Self and destruct system data tuple in parameter list.

WaDelma:
This PR changes component definitions from:
````rust
impl Component for MyTotallyRadComponent {
    type Storage = VecStorage<MyTotallyRadComponent>;
}
````
to
````rust
impl Component for MyTotallyRadComponent {
    type Storage = VecStorage<Self>;
}
````
which is shorter for long types and easier to rename. Benefits are especially great when the type has to be repeated multiple times (eg. `FlaggedStorage<MyTotallyRadComponent, VecStorage<MyTotallyRadComponent>>` vs `FlaggedStorage<Self, VecStorage<Self>>`)

It also moves system data destructuring for tuples:
````rust
fn run(&mut self, data: Self::SystemData) {
    let (first_component, mut second_component) = data;
````
to
````rust
fn run(&mut self, (first_component, mut second_component): Self::SystemData) {
````
which is shorter and conveys the fact that these component storages are takes as parameter (same way as tuple desugaring takes parameters).

NOTE: Removal of spaces at end of lines was done automatically by editor

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@
 [gi]: https://badges.gitter.im/slide-rs/specs.svg
 [gl]: https://gitter.im/slide-rs/specs
 
-Specs is an Entity-Component System written in Rust. 
+Specs is an Entity-Component System written in Rust.
 Unlike most other ECS libraries out there, it provides
 
 * easy parallelism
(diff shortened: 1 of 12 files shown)


kvark:
Thanks, looks great!
bors r+

PR #208, opened 2017-07-01
AntiStorage for CheckStorage

Aceeri:
Just allows using `Not` with a `CheckStorage`.


kvark:
Looks reasonable!
bors r+

PR #209, opened 2017-07-01
Integrate futures

torkleyy:
cc @ebkalderon @Xaeroxe

Will be used by https://github.com/amethyst/amethyst/pull/244

### DIFF
--- a/src/common.rs
+++ b/src/common.rs
@@ -1,0 +1,173 @@
+//! Common functionality you might need when using Specs.
+//!
+//! At the moment, this module provides two types:
+//!
+//! * `Errors`: A resource you can use to store errors that occurred outside of
+//!             the ECS but were catched inside, therefore should be handled by the user
+//!
+//! * `Merge`: A system generic over `T` which automatically merges `Ready` futures into
+//!            the component storage for `T`.
+//!
+//! To make use of these features, you need to ask for the `common` feature
+//! like this:
+//!
+//! ```toml
+//! [dependencies.specs]
+//! # version = "..."
+//! features = ["common"]
+//! ```
+
+use std::convert::AsRef;
+use std::error::Error;
+use std::fmt::{Debug, Display, Formatter, Result as FmtResult};
+use std::io::Write;
+use std::marker::PhantomData;
+
+use futures::{Async, Future};
+use {Component, DenseVecStorage, Entities, FetchMut, Join, RunningTime, System, WriteStorage};
+
+/// A boxed error implementing `Debug`, `Display` and `Error`.
+pub struct BoxedErr(pub Box<Error + Send + Sync + 'static>);
+
+impl BoxedErr {
+    /// Creates a new boxed error.
+    pub fn new<T>(err: T) -> Self
+        where T: Error + Send + Sync + 'static
+    {
+        BoxedErr(Box::new(err))
+    }
+}
+
+impl AsRef<Error> for BoxedErr {
+    fn as_ref(&self) -> &(Error + 'static) {
+        self.0.as_ref()
+    }
+}
+
+impl Debug for BoxedErr {
+    fn fmt(&self, f: &mut Formatter) -> FmtResult {
+        write!(f, "{:?}", self.as_ref())
+    }
+}
+
+impl Display for BoxedErr {
+    fn fmt(&self, f: &mut Formatter) -> FmtResult {
+        write!(f, "{}", self.as_ref())
+    }
+}
+
+impl Error for BoxedErr {
+    fn description(&self) -> &str {
+        self.as_ref().description()
+    }
+}
+
+/// The component type for futures of `T`.
+pub type FutureComponent<T> = Box<Future<Item = T, Error = BoxedErr> + Send + Sync>;
+
+impl<T> Component for FutureComponent<T>
+    where T: Send + Sync + 'static
+{
+    type Storage = DenseVecStorage<FutureComponent<T>>;
+}
+
+/// A resource you can use to store errors that occurred outside of
+/// the ECS but were catched inside, therefore should be handled by the user.
+#[derive(Debug)]
+pub struct Errors {
+    /// The collection of errors.
+    pub errors: Vec<BoxedErr>,
+}
+
+impl Errors {
+    /// Creates a new instance of `Errors`.
+    pub fn new() -> Self {
+        Errors { errors: Vec::new() }
+    }
+
+    /// Add an error to the error `Vec`.
+    pub fn add(&mut self, error: BoxedErr) {
+        self.errors.push(error);
+    }
+
+    /// Prints all errors and exits. Useful for debugging.
+    pub fn print_and_exit(&mut self) {
+        use std::io::stderr;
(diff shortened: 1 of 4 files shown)


src/common.rs
```diff
@@ -0,0 +1,173 @@
...
+                 self.errors.len()).unwrap();
+
+        for error in self.errors.drain(..) {
+            let error = error.as_ref();
+
+            writeln!(&mut stderr(),
+                     "-----\n\tDescription: {}\n\tDebug: {:?}\n\tDisplay: {}",
+                     error.description(),
+                     error,
+                     error).unwrap();
```
torkleyy: Feel free to suggest a better format.

src/common.rs
```diff
@@ -0,0 +1,173 @@
...
+        Errors { errors: Vec::new() }
+    }
+
+    /// Add an error to the error `Vec`.
+    pub fn add(&mut self, error: BoxedErr) {
+        self.errors.push(error);
+    }
+
+    /// Prints all errors and exits. Useful for debugging.
+    pub fn print_and_exit(&mut self) {
```
WaDelma: I would prefer if debug printing came from `Debug` implementation instead of this kind of debug function.

src/common.rs
```diff
@@ -0,0 +1,173 @@
...
+}
+
+impl Error for BoxedErr {
+    fn description(&self) -> &str {
+        self.as_ref().description()
+    }
+}
+
+/// The component type for futures of `T`.
+pub type FutureComponent<T> = Box<Future<Item = T, Error = BoxedErr> + Send + Sync>;
```
WaDelma: I understand that if the component comes from multiple different source then boxing the future is nessary. Does it make sense when `impl Trait` becomes stable to support unboxed futures too? Not sure if there is usecase.

src/common.rs
```diff
@@ -0,0 +1,173 @@
...
+{
+    type SystemData = (Entities<'a>,
+                       FetchMut<'a, Errors>,
+                       WriteStorage<'a, FutureComponent<T>>,
+                       WriteStorage<'a, T>);
+
+    fn run(&mut self, data: Self::SystemData) {
+        let (entities, mut errors, mut future, mut pers) = data;
+
+        let mut delete = Vec::new();
```
WaDelma: Would it be possible to move this allocation to the struct and reuse it?

src/common.rs
torkleyy: I don't think that limiting it to one specific implementation of `Future` is a very good idea since you wouldn't be able to do any nice `map`ping or something like that.

src/common.rs
```diff
@@ -0,0 +1,173 @@
...
+    }
+}
+
+/// The component type for futures of `T`.
+pub type FutureComponent<T> = Box<Future<Item = T, Error = BoxedErr> + Send + Sync>;
+
+impl<T> Component for FutureComponent<T>
+    where T: Send + Sync + 'static
+{
+    type Storage = DenseVecStorage<FutureComponent<T>>;
```
torkleyy: I'd say `DenseVecStorage` is pretty much a good choice here.

src/common.rs
WaDelma: Well if you just use trait bounds then the user gets to choose do they want to box it. The point is that if you know that certain component comes only from certain future there is no need for boxing.

src/common.rs
WaDelma: Yeah, it's probably the best if we can't give choice for the user, but it's still restricting.

src/common.rs
```diff
@@ -0,0 +1,173 @@
...
+    fn run(&mut self, data: Self::SystemData) {
+        let (entities, mut errors, mut future, mut pers) = data;
+
+        let mut delete = Vec::new();
+
+        for (e, future) in (&*entities, &mut future).join() {
+            match future.poll() {
+                Ok(Async::Ready(x)) => {
+                    pers.insert(e, x);
+                    delete.push(e);
```
omni-viral: Why you defer `future.remove(e);` instead of calling here?

ebkalderon:
@torkleyy Have you thought of possibly using Alex Crichton's `futures-await`  crate instead of explicitly polling each `FutureComponent` in a system?

src/common.rs
torkleyy: Because `future` (the storage) is borrowed already. There's no way of removing from the storage you're iterating over.

src/common.rs
```diff
@@ -0,0 +1,173 @@
...
+use std::error::Error;
+use std::fmt::{Debug, Display, Formatter, Result as FmtResult};
+use std::io::Write;
+use std::marker::PhantomData;
+
+use futures::{Async, Future};
+use {Component, DenseVecStorage, Entities, FetchMut, Join, RunningTime, System, WriteStorage};
+
+/// A boxed error implementing `Debug`, `Display` and `Error`.
+pub struct BoxedErr(pub Box<Error + Send + Sync + 'static>);
```
torkleyy: Well, `Errors` is supposed to contain all the errors because everything else would be super complicated.

torkleyy:
Added another commit so it's easier to review, will rebase once approved.

torkleyy:
Okay, I'd consider it done now, please leave a comment / review if there are still issues to resolve.

src/common.rs
torkleyy: It might be possible with newtypes and more type parameters but I don't think it's worth it. After all, restricting it to one particular implementation is quite limiting. I usually try to avoid `Box`es, but I think it'd be good to have one here.

torkleyy: Oh well, let's still allow it.

torkleyy: It would theoretically be possible to implement `Component` for futures themselves, but that would be a breaking change.

torkleyy:
@WaDelma Lifted both restrictions, more boilerplate is now required...

src/common.rs
```diff
@@ -0,0 +1,167 @@
...
+use std::error::Error;
+use std::fmt::{Debug, Display, Formatter, Result as FmtResult};
+use std::io::Write;
+use std::marker::PhantomData;
+
+use futures::{Async, Future};
+use {Component, Entity, Entities, FetchMut, Join, RunningTime, System, WriteStorage};
+
+/// A boxed error implementing `Debug`, `Display` and `Error`.
+pub struct BoxedErr(pub Box<Error + Send + Sync + 'static>);
```
kvark: can we just have `type BoxedErr = ...` ? In this case simple `Box::new()` will be used instead of `BoxedErr::new()`

src/common.rs
kvark: Indeed not possible right now, but if it really becomes a usability problem, we may just have `Join` implementation for `&mut Storage` to return thin wrappers that are dereffed to `&mut T` with an ability to remove the component in place.

src/common.rs
torkleyy: I did it this way because `Box<Error>` does not implement `Error` - dunno why that is.

torkleyy:
Okay, seems there are no concerns left. Thanks for the feedback!

bors r+

PR #212, opened 2017-07-02
Add other examples, required-features for serialize example

Aceeri:
Cargo can't seem to find the other examples without them being specified now for whatever reason, also cleans up the serialize example by using `required-features` instead of manually doing it in the example.


torkleyy:
I assume it's not finished yet?

torkleyy:
Thanks, merging.

bors r+

PR #214, opened 2017-07-03
Add LazyInsertions

### LINKED ISSUE #213: Lazy Insertion
Constructing an entity while the world is running isn't very ergonomic right now and the naive way of doing it can lead to almost no parallel systems (Accessing each storage required for construction mutably, so it blocks all other systems using that storage). A better approach is something like an "initialization" system where it is possible to insert a flag component that signifies that a certain component should be inserted. For example:
```rust
struct Position { x: f32, y: f32, z: f32 };
struct InitPosition;
struct InitPositionSystem;
impl<'a> System<'a> for InitPositionSystem {
    type SystemData = (
        Entities<'a>, 
        WriteStorage<'a, Position>, 
        ReadStorage<'a, InitPosition>
    );
    fn run(&mut self, (entities, positions, inits): Self::SystemData) {
        for (entity, _) in (&*entities, &inits).join() {
            positions.insert(entity, Position { x: 0.0, y: 0.0, z: 0.0 });
        }
    }
}
```

I feel like this could be abstracted over a bit by allowing some sort of lazy insertion method for the world. So instead of having to set up all this for an initialization system, you could just do something like:
```rust
impl<'a> System<'a> for SomeSys {
    type SystemData = (
        LazyInsert<'a>,
        Entities<'a>, 
    );
    fn run(&mut self, (lazy, entities): Self::SystemData) {
        for entity in (&*entities).join() {
            if ... { // some condition to insertion or something
                lazy.insert::<Position>(entity, Position { x: 0.0, y: 0.0, z: 0.0 });
            }
        }
    }
}
```
This would put the component into some queue in the world, so then once the world is maintained it inserts the components.

torkleyy:
Fixes #213

### DIFF
--- a/src/world.rs
+++ b/src/world.rs
@@ -1,5 +1,6 @@
 use std::sync::atomic::{AtomicUsize, Ordering};
 
+use crossbeam::sync::TreiberStack;
 use hibitset::{AtomicBitSet, BitSet, BitSetOr};
 use mopa::Any;
 use shred::{Fetch, FetchMut, Resource, Resources};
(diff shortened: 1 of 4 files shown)


src/world.rs
```diff
@@ -366,6 +366,60 @@ impl Generation {
...
+}
+
+/// Lazy insertions can be used after creating a new
+/// entity in a system. This way, none of the actual
+/// component storages have to be borrowed mutably.
+///
+/// This resource is added to the world by default.
+#[derive(Default)]
+pub struct LazyInsertions {
+    vec: Vec<Box<LazyInsertInternal>>
```
torkleyy: Should use an atomic linked list.

src/world.rs
```diff
@@ -366,6 +366,60 @@ impl Generation {
...
+
+impl<L> LazyInsertInternal for Option<L>
+    where L: LazyInsert
+{
+    fn insert(&mut self, world: &World) {
+        self.take().unwrap().insert(world);
+    }
+}
+
+/// Lazy insertions can be used after creating a new
```
torkleyy: Needs docs somewhere that the insertions are in reverse order.

src/world.rs
```diff
@@ -366,6 +366,60 @@ impl Generation {
...
+///
+/// This resource is added to the world by default.
+#[derive(Default)]
+pub struct LazyInsertions {
+    vec: Vec<Box<LazyInsertInternal>>
+}
+
+impl LazyInsertions {
+    /// Adds an insertion.
+    pub fn add<L>(&mut self, l: L)
```
torkleyy: Missing usage example

Aceeri:
https://github.com/slide-rs/specs/issues/213

src/world.rs
```diff
@@ -366,6 +366,60 @@ impl Generation {
...
+
+trait LazyInsertInternal: Send + Sync {
+    fn insert(&mut self, world: &World);
+}
+
+impl<L> LazyInsertInternal for Option<L>
+    where L: LazyInsert
+{
+    fn insert(&mut self, world: &World) {
+        self.take().unwrap().insert(world);
```
kvark: should we check for `None` here?

torkleyy: It's local to the module, so cannot be triggered by the user. I doubt that `expect` or some special casing would be of much use here.

src/world.rs
```diff
@@ -366,6 +367,66 @@ impl Generation {
...
+{
+    fn insert(self, world: &World) {
+        for item in self {
+            item.insert(world);
+        }
+    }
+}
+
+trait LazyInsertInternal: Send + Sync {
+    fn insert(&mut self, world: &World);
```
WaDelma: I am slightly confused why doesn't this method take ownership of `self` and instead uses `Option::take`.

torkleyy: That's because `Box<Foo>` cannot be the receiver of a self-moving method since `self` is `!Sized`.

kvark:
@torkleyy r=me when ready

src/world.rs
WaDelma: Hmmm... Can't you take it as `Box<self>` then?

WaDelma: Hmm... Can't you take it as `self: Box<Self>` then?

torkleyy: I tried that first but it didn't work out at some point. Well, I found another solution using `Box<Self>`.

kvark:
Might be a good idea to squash the commits a bit, otherwise all looking good.

PR #217, opened 2017-07-03
Also register Velocity component

svenstaro:
If we don't have this:

> thread 'main' panicked at 'No component with the given id. Did you forget to register the component with `World::register::<ComponentName>()`?', /checkout/src/libcore/option.rs:823:4


torkleyy:
Oops, didn't get that mail. Thanks.

bors r+

PR #221, opened 2017-07-09
LazyUpdate: lazily insert and remove components, execute closures on world

eldyer:
Remove a component lazily without having to borrow its storage. To be used like this:

`let lazy = world.read_resource::<LazyUpdate>();`
`lazy.add_removal::<SomeComponentType>(some_entity);`

### DIFF
--- a/src/world.rs
+++ b/src/world.rs
@@ -460,6 +460,42 @@ impl Drop for LazyInsertions {
     }
 }
 
+/// Lazy removals can be used after creating a new
+/// entity in a system. This way, none of the actual
+/// component storages have to be borrowed mutably.
+///
+/// This resource is added to the world by default.
+pub struct LazyRemovals {
+    stack: TreiberStack<Box<Fn(&World) + Send + Sync>>,
+}
+
+impl LazyRemovals {
+    /// Adds a removal. Please note that this method takes `&self`
+    /// so there's no need to fetch it mutably.
+    pub fn add<C>(&self, e: Entity)
+    where
+        C: Component + Send + Sync,
+    {
+        self.stack.push(Box::new(
+            move |world| { world.write::<C>().remove(e); },
+        ));
+    }
+}
+
+impl Default for LazyRemovals {
+    fn default() -> Self {
+        // TODO: derive (`Default` is not yet implemented for `TreiberStack`)
+        LazyRemovals { stack: TreiberStack::new() }
+    }
+}
+
+impl Drop for LazyRemovals {
+    fn drop(&mut self) {
+        // TODO: remove as soon as leak is fixed in crossbeam
+        while self.stack.pop().is_some() {}
+    }
+}
+
 /// The `World` struct contains the component storages and
 /// other resources.
 ///
@@ -725,6 +761,13 @@ impl World {
         while let Some(l) = lazy.pop() {
             l.insert(&*self);
         }
+        
+        let mut lazy_removals = self.write_resource::<LazyRemovals>();
+        let lazy = &mut lazy_removals.stack;
+
+        while let Some(remove) = lazy.pop() {
+            remove(self);
+        }
     }
 
     fn delete_components(&mut self, delete: &[Entity]) {
@@ -743,6 +786,7 @@ impl Default for World {
         let mut res = Resources::new();
         res.add(EntitiesRes::default());
         res.add(LazyInsertions::default());
+        res.add(LazyRemovals::default());
 
         World {
             res: res,
@@ -778,5 +822,20 @@ mod tests {
 
         world.maintain();
         assert!(world.read::<Pos>().get(e).is_some());
+    }
+    
+    #[test]
+    fn lazy_removal() {
+        let mut world = World::new();
+        world.register::<Pos>();
+
+        let e = world.create_entity().with(Pos).build();
+        {
+            let lazy = world.read_resource::<LazyRemovals>();
+            lazy.add::<Pos>(e);
+        }
+
+        world.maintain();
+        assert!(world.read::<Pos>().get(e).is_none());
     }
 }
--- a/src/world.rs
+++ b/src/world.rs
@@ -391,18 +391,7 @@ impl<L> LazyInsert for Vec<L>
     }
 }
 
-struct LazyRemovalMarker<F>(PhantomData<F>);
-
-impl<F> LazyRemovalMarker<F> {
-    pub fn new() -> Self {
-        LazyRemovalMarker(PhantomData)
-    }
-}
-
-struct LazyRemovalInfo<C> {
-    entity: Entity,
-    marker: LazyRemovalMarker<C>,
-}
+struct LazyRemovalInfo<C>(Entity, PhantomData<C>);
 
 trait LazyUpdateInternal: Send + Sync {
     fn update(self: Box<Self>, world: &World);
@@ -420,22 +409,29 @@ impl<C> LazyUpdateInternal for LazyRemovalInfo<C>
     where C: Component + Send + Sync
 {
     fn update(self: Box<Self>, world: &World) {
-        world.write::<C>().remove(self.entity);
+        world.write::<C>().remove(self.0);
     }
 }
 
-/// Lazy updates can be used after creating a new
-/// entity in a system. This way, none of the actual
-/// component storages have to be borrowed mutably.
-///
+impl LazyUpdateInternal for Box<Fn(&World) + Send + Sync> {
+    fn update(self: Box<Self>, world: &World) {
+        self(world);
+    }
+}
+
+/// Lazy updates can be used for world updates
+/// that need to borrow a lot of resources
+/// and as such should better be done at the end
+/// when maintaining the world.
 /// This resource is added to the world by default.
+/// Please note that the provided methods take `&self`
+/// so there's no need to fetch `LazyUpdate` mutably.
 pub struct LazyUpdate {
     stack: TreiberStack<Box<LazyUpdateInternal>>,
 }
 
 impl LazyUpdate {
-    /// Adds an insertion. Please note that this method takes `&self`
-    /// so there's no need to fetch it mutably.
+    /// Lazily inserts component(s) for one or more entities.
     ///
     /// ## Examples
     ///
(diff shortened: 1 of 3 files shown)


kvark:
Looks ok to me. Thanks for the PR!

src/world.rs
```diff
@@ -460,6 +460,42 @@ impl Drop for LazyInsertions {
     }
 }
 
+/// Lazy removals can be used after creating a new
+/// entity in a system. This way, none of the actual
+/// component storages have to be borrowed mutably.
+///
+/// This resource is added to the world by default.
+pub struct LazyRemovals {
```
torkleyy: There is no reason to have a second struct, actually. Given that `LazyInsertions` haven't been released yet, let's just update it to `LazyUpdate` and have a method there.

src/world.rs
```diff
@@ -407,11 +407,12 @@ impl<L> LazyInsertInternal for L
 /// component storages have to be borrowed mutably.
 ///
 /// This resource is added to the world by default.
-pub struct LazyInsertions {
-    stack: TreiberStack<Box<LazyInsertInternal>>
+pub struct LazyUpdate {
+    insert_stack: TreiberStack<Box<LazyInsertInternal>>,
```
torkleyy: Let's also rename `LazyInsertInternal` to `LazyUpdateInternal` so we do not need two stacks.

torkleyy: To elaborate this a bit, `LazyInsertInternal` can just access the `World`, so we can just use it for both (after renaming). Just accept entities and a type parameter then, have a private struct with `PhantomData` and implement the removal, please.

eldyer: Sounds good, I will try!

src/world.rs
```diff
@@ -390,26 +391,41 @@ impl<L> LazyInsert for Vec<L>
     }
 }
 
-trait LazyInsertInternal: Send + Sync {
-    fn insert(self: Box<Self>, world: &World);
+struct LazyRemovalMarker<F>(PhantomData<F>);
+
+impl<F> LazyRemovalMarker<F> {
```
torkleyy: Could we just have a struct containing both the entity and the marker?

src/world.rs
```diff
@@ -390,28 +391,49 @@ impl<L> LazyInsert for Vec<L>
...
+    pub fn new() -> Self {
+        LazyRemovalMarker(PhantomData)
+    }
+}
+
+struct LazyRemovalInfo<C> {
+    entity: Entity,
+    marker: LazyRemovalMarker<C>,
```
torkleyy: I think we can actually inline that type.

src/world.rs
```diff
@@ -390,28 +391,49 @@ impl<L> LazyInsert for Vec<L>
...
-/// Lazy insertions can be used after creating a new
+impl<C> LazyUpdateInternal for LazyRemovalInfo<C>
+    where C: Component + Send + Sync
+{
+    fn update(self: Box<Self>, world: &World) {
+        world.write::<C>().remove(self.entity);
+    }
+}
+
+/// Lazy updates can be used after creating a new
```
torkleyy: I'd say in general, they can be used for stuff that should better happen at the end because it needs to borrow many resources. Dealing with initialization of components could be shown as an example of that.

src/world.rs
```diff
@@ -429,31 +451,45 @@ impl LazyInsertions {
...
     ///         let a = ent.create();
     ///         let b = ent.create();
     ///
-    ///         lazy.add(vec![(a, Pos(3.0, 1.0)), (b, Pos(0.0, 4.0))]);
+    ///         lazy.add_insertions(vec![(a, Pos(3.0, 1.0)), (b, Pos(0.0, 4.0))]);
     ///     }
     /// }
     /// ```
-    pub fn add<L>(&self, l: L)
+    pub fn add_insertions<L>(&self, l: L)
```
torkleyy: Maybe it would be nicer to do a lazy insert - so just `pub fn insert`? Also, the plural is not always applicable.

src/world.rs
```diff
@@ -429,31 +451,45 @@ impl LazyInsertions {
...
         where L: LazyInsert + 'static
     {
         self.stack.push(Box::new(l));
     }
+
+    /// Adds a removal. Please note that this method takes `&self`
+    /// so there's no need to fetch it mutably.
+    pub fn add_removal<C>(&self, e: Entity)
```
torkleyy: Then this one would be just `remove`.

src/world.rs
```diff
@@ -429,31 +451,45 @@ impl LazyInsertions {
...
     /// }
     /// ```
-    pub fn add<L>(&self, l: L)
+    pub fn add_insertions<L>(&self, l: L)
         where L: LazyInsert + 'static
     {
         self.stack.push(Box::new(l));
     }
+
+    /// Adds a removal. Please note that this method takes `&self`
```
torkleyy: It would also be nice to have a usage example somewhere.

eldyer:
Thanks for your feedback and no need to apologize, I agree with all points and I'm working on it!

src/world.rs
eldyer: Do you mean using #[inline] on getters / setters like with the Entity struct?

torkleyy: No, just use `PhantomData` here.

eldyer: Ah right, of course!

eldyer:
Also implemented execute. Tell me what you think, I'm happy to get your feedback.

src/world.rs
```diff
@@ -420,22 +409,29 @@ impl<C> LazyUpdateInternal for LazyRemovalInfo<C>
...
-        world.write::<C>().remove(self.entity);
+        world.write::<C>().remove(self.0);
     }
 }
 
-/// Lazy updates can be used after creating a new
-/// entity in a system. This way, none of the actual
-/// component storages have to be borrowed mutably.
-///
+impl LazyUpdateInternal for Box<Fn(&World) + Send + Sync> {
```
torkleyy: Do we really need `Box` here?

src/world.rs
```diff
@@ -457,26 +453,84 @@ impl LazyUpdate {
...
         C: Component + Send + Sync,
     {
-        self.stack.push(Box::new(LazyRemovalInfo {
-            entity: e,
-            marker: LazyRemovalMarker::<C>::new(),
-        }));
+        self.stack.push(Box::new(LazyRemovalInfo::<C>(e, PhantomData)));
+    }
+
+    /// Lazily executes a closure with world access.
```
torkleyy: Could also say that "lazily" means "when calling maintain()".

src/world.rs
```diff
@@ -429,31 +447,103 @@ impl LazyInsertions {
...
+    ///             lazy.remove::<Pos>(entity);
+    ///         }
+    ///     }
+    /// }
+    /// ```
+    pub fn remove<C>(&self, e: Entity)
+    where
+        C: Component + Send + Sync,
+    {
+        self.stack.push(Box::new(LazyRemovalInfo::<C>(e, PhantomData)));
```
torkleyy: Actually, this could just use `execute` now. That way, there's no need for the removal struct.

src/world.rs
```diff
@@ -429,31 +447,103 @@ impl LazyInsertions {
...
     ///         let a = ent.create();
     ///         let b = ent.create();
     ///
-    ///         lazy.add(vec![(a, Pos(3.0, 1.0)), (b, Pos(0.0, 4.0))]);
+    ///         lazy.insert(vec![(a, Pos(3.0, 1.0)), (b, Pos(0.0, 4.0))]);
     ///     }
     /// }
     /// ```
-    pub fn add<L>(&self, l: L)
+    pub fn insert<L>(&self, l: L)
```
torkleyy: We could probably divide this method into `insert` and `insert_all` to increase performance for multiple insertions; `insert_all` could just accept an iterator.

src/world.rs
eldyer: Nice!

src/world.rs
eldyer: Not sure, I tried without that Box first, but ran into problems. The compiler suggests to add a 'static bound to F in execute, but the passed in closures apparently don't have 'static lifetime. Any idea how to solve this?

src/world.rs
eldyer: Agreed, will do that.
It would be nice if we could use execute in insert as well, to get rid of LazyInsert. But I think then FnOnce would be required, because the component would have to be moved into the closure, right?

torkleyy: Yes. I'm not sure, however, if executing `call_once` on `Box<FnOnce` is permitted / if there's a workaround.

src/world.rs
torkleyy: Did you try `impl<F> LazyUpdateInternal for F where F: Fn(&World) + Send + Sync + 'static`?

eldyer: Ah, works!

eldyer:
Could you take a look? Still needs cleanup, docs and examples, just wanted to get some feedback at this point.

eldyer:
Hm, would it make sense to rename`LazyUpdate` into `LazyExecute` or `LazyExecution`?

src/world.rs
```diff
@@ -521,16 +520,14 @@ impl LazyUpdate {
...
     /// ```
     pub fn execute<F>(&self, f: F)
     where
-        F: Fn(&World) + 'static + Send + Sync,
+        F: FnOnce(&World) + 'static + Send + Sync,
     {
-        self.stack.push(Box::new(
-            Box::new(f) as Box<Fn(&World) + Send + Sync>)
-        );
+        self.stack.push(Box::new(f));
```
torkleyy: Much better :wink:

src/world.rs
```diff
@@ -390,30 +390,30 @@ impl<L> LazyInsert for Vec<L>
...
+    fn update(self: Box<Self>, world: &World);
 }
 
-impl<L> LazyInsertInternal for L
-    where L: LazyInsert
-{
-    fn insert(self: Box<Self>, world: &World) {
-        L::insert(*self, world);
+impl<F> LazyUpdateInternal for F
+    where F: FnOnce(&World) + Send + Sync + 'static {
```
torkleyy: Formatting: `{` is supposed to be on a new line.

src/world.rs
```diff
@@ -390,30 +390,30 @@ impl<L> LazyInsert for Vec<L>
...
-///
+/// Lazy updates can be used for world updates
+/// that need to borrow a lot of resources
+/// and as such should better be done at the end
+/// when maintaining the world.
 /// This resource is added to the world by default.
-pub struct LazyInsertions {
-    stack: TreiberStack<Box<LazyInsertInternal>>
+/// Please note that the provided methods take `&self`
+/// so there's no need to fetch `LazyUpdate` mutably.
```
torkleyy: Let's leave the note that it is added to the world by default.

src/world.rs
```diff
@@ -429,31 +429,118 @@ impl LazyInsertions {
...
+            I: IntoIterator<Item=(Entity, C)> + Send + Sync + 'static,
+    {
+        self.execute(move |world| {
+            let mut storage = world.write::<C>();
+            for (e, c) in iter {
+                storage.insert(e, c);
+            }
+        });
+    }
+    ///
```
torkleyy: Missing new line, docs and example usage.

torkleyy:
I guess we can merge this now.

bors r+

torkleyy:
Hold on, rebase is wrong

bors r-

eldyer:
Yes I was just about to fix the Changelog...

torkleyy:
The line in the changelog should be in the 0.9.3 section

eldyer:
Yeah just committed... 😄

torkleyy:
Okay, now it's perfect.

bors r+

eldyer:
Thank you by the way for guiding me through this. It's actually my first PR and I learned a lot of things. I totally like that you want to get all the details right... it's a great thing to learn from such talented developers! 👍

Issue #204, opened 2017-06-22
Add judy array as storage type


WaDelma:
Add and benchmark [judy array](https://github.com/adevore/rudy/) as possibly faster storage.

@torkleyy I noticed you were interested in adding it as storage so I figured I would create issue for it so it doesn't get forgotten.

willglynn:
I took a whack at this and hit a couple snags, first of which was that the current `rudy` crate's `RudyMap` has no `remove()`. After switching to a git checkout (and building with nightly), I got everything wired up behind a feature flag, and found five of the usual storage tests fail against `RudyMap`.

I pushed this to a [`rudy` branch](https://github.com/willglynn/specs/tree/rudy) but have no immediate plans to open a PR.

adevore:
The bug in Rudy has been fixed in master. I ran the tests locally on the rudy branch of specs and all tests pass.

willglynn:
@WaDelma @torkleyy Would you like me to open a PR now (as an optional dependency pointing to git), or would you prefer me to wait for a followup `rudy` release?

WaDelma:
I would prefer if I didn't accidentaly introduce the bug you encountered :P

We cannot release new version of `specs` before the fixed version of `rudy` is released, so I would prefer waiting for it.

torkleyy:
A PR would be nice, yeah :+1:

willglynn:
`RudyStorage` landed in #222, so this issue can be closed.

torkleyy:
Probably not, we still need proper benchmark results and a secrion in the book.

willglynn:
Ah! You are correct. This is what I get for trying to cover all my tabs with time-limited Internet access :-) Cheers!

PR #222, opened 2017-07-10
Add RudyStorage

willglynn:
Per #204, `RudyStorage` presently has some caveats:

* Depends on a git checkout of `rudy`
* Depends on Rust nightly

`RudyStorage` therefore lives behind `--features rudy`. Travis now also knows about this feature flag.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,6 +19,9 @@ script:
 - if [ "$TRAVIS_RUST_VERSION" == "nightly" ]; then
     cargo bench --verbose --no-run;
   fi
+- if [ "$TRAVIS_RUST_VERSION" == "nightly" ]; then
+    cargo test --verbose --features rudy;
+  fi
 
 after_success: |
   [ $TRAVIS_BRANCH = master ] &&
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,11 +16,9 @@ script:
 - cargo test --verbose --no-default-features
 - cargo build --verbose --features serialize
 - cargo test --verbose --features serialize
+- cargo test --verbose --features rudy
 - if [ "$TRAVIS_RUST_VERSION" == "nightly" ]; then
     cargo bench --verbose --no-run;
-  fi
-- if [ "$TRAVIS_RUST_VERSION" == "nightly" ]; then
-    cargo test --verbose --features rudy;
   fi
 
 after_success: |
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,7 @@ script:
 - cargo test --verbose --no-default-features
 - cargo build --verbose --features serialize
 - cargo test --verbose --features serialize
+- cargo build --verbose --features rudy
 - cargo test --verbose --features rudy
 - if [ "$TRAVIS_RUST_VERSION" == "nightly" ]; then
     cargo bench --verbose --no-run;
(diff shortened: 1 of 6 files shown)


.travis.yml
```diff
@@ -19,6 +19,9 @@ script:
 - if [ "$TRAVIS_RUST_VERSION" == "nightly" ]; then
     cargo bench --verbose --no-run;
   fi
+- if [ "$TRAVIS_RUST_VERSION" == "nightly" ]; then
```
torkleyy: Why is it nightly-only?

willglynn: Rudy is nightly-only: [src/lib.rs +4-6](https://github.com/adevore/rudy/blob/d3d97233e51cdc9714ecfeffa589fb7028ee7cec/src/lib.rs#L4-L6)

torkleyy: Yeah, I see it now. Is there an intention to make it / parts of it stable?

willglynn: I imagine so, but can't speak for the project. @adevore? @WaDelma?

WaDelma: `conservative_impl_trait` seems to be used only in commented out iterator methods.
`associated_consts` seems to be used for `Key` trait.
`slice_patterns` seems to be used in `Leaf2::shrink_remove` (I actually added this code... Didn't even realise it was only allowed because of feature).
These all can be worked around, so it could be made to work in stable.

adevore: I have opened adevore/rudy#16 to organize removing nightly features.

adevore: After adevore/rudy#16 is completed, I plan on doing a 0.1 release of Rudy. My goal is to do a release this weekend.

Cargo.toml
```diff
@@ -41,6 +41,9 @@ futures = { version = "0.1.14", optional = true }
 serde = { version = "1.0", optional = true }
 serde_derive = { version = "1.0", optional = true }
 
+# enable rudy via -
+rudy = { optional = true, git = "https://github.com/adevore/rudy" } # pending >0.0.1 release
```
Aceeri: Should use a `commit` with a ref/hash here.

willglynn: My intent was to track `rudy` git on the expectation that its public API is unlikely to change between now and the next crate release. That said, the next release should be Soon™, after which I'll happily point to that instead.

adevore:
Rudy 0.1.0 has been released to crates.io. I am committing to keeping the master branch and releases working on stable.

willglynn:
I updated the PR. Rudy is still optional and gated behind `--features rudy`, but when selected, it now pulls from crates.io and can build on stable Rust.

Cargo.toml
```diff
@@ -41,6 +41,9 @@ futures = { version = "0.1.14", optional = true }
 serde = { version = "1.0", optional = true }
 serde_derive = { version = "1.0", optional = true }
 
+# enable rudy via -
```
torkleyy: I think you accidentally deleted the rest of the line.

.travis.yml
```diff
@@ -16,6 +16,7 @@ script:
 - cargo test --verbose --no-default-features
 - cargo build --verbose --features serialize
 - cargo test --verbose --features serialize
+- cargo test --verbose --features rudy
```
torkleyy: Let's also do a build before testing.

willglynn:
I updated the PR to incorporate feedback.

PR #225, opened 2017-07-15
Storage benchmarks

Aceeri:
Just sets up some basic benchmarks for different storages.


torkleyy:
No concerns, merging. Thanks again!

bors r+

PR #230, opened 2017-07-23
Join implementations for BitSetLikes

Aceeri:


torkleyy:
Isn't that what `CheckStorage` is (will be) for?

Aceeri:
Sort of? This is kind of a superset of that where you can just use any `BitSetLike`, so you could unwrap the storage and then send in the bitset, which would make `CheckStorage` a convenience method really.

torkleyy:
Okay, I think we can do that. But I would really want `check` to return a `BitSet` then.

Aceeri:
I'll just include that change in this PR once I get a moment.

Aceeri:
@torkleyy This should be good to go now.

src/bitset.rs
```diff
@@ -0,0 +1,27 @@
+
+use hibitset::{BitSet, BitSetAnd, BitSetLike, BitSetNot, BitSetOr};
+
+use ::{Index, Join};
+
+macro_rules! define_bit_join {
+    ( $bitset:ident [ $( $arg:ident ),* ] ) => {
+        impl<'a, $( $arg ),*> Join for &'a $bitset<$( $arg ),*>
```
torkleyy: Do we also need `ParJoin` here? @WaDelma

WaDelma: Yeah it looks like it.

Aceeri:
Can we get this merged? I was hoping to have it merged before the next release.

PR #231, opened 2017-07-25
Move FlaggedStorage to its own file

Aceeri:
Was just getting more difficult to read the file with half of the file being the `FlaggedStorage`, plus it would probably be a better convention for `src/storage/storage.rs` to hold "real" storages like `HashMapStorage` and `DenseVecStorage`, while keeping the "wrapper" storages (`FlaggedStorage`) and "operation" (`AntiStorage`, `CheckStorage`, `RestrictedStorage`) in their own locations.

### DIFF
--- a/src/storage/flagged.rs
+++ b/src/storage/flagged.rs
@@ -1,0 +1,158 @@
+
+use std::marker::PhantomData;
+
+use hibitset::BitSet;
+
+use world::EntityIndex;
+use {Index, Join, UnprotectedStorage};
+
+/// Wrapper storage that stores modifications to components in a bitset.
+///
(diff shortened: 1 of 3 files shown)


src/storage/flagged.rs
```diff
@@ -0,0 +1,147 @@
+
+use std::marker::PhantomData;
+
+use hibitset::BitSet;
+
+use world::EntityIndex;
+use ::{Index, Join, UnprotectedStorage};
```
torkleyy: No need for `::`.

torkleyy:
Oops, there are conflicts.

torkleyy:
Now bors also noticed :)

PR #233, opened 2017-07-31
Added MaskedStorage to the re-exports

VoxWave:
This allows being generic over read and write storages.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -206,7 +206,7 @@ pub use shred::{Dispatcher, DispatcherBuilder, Fetch, FetchId, FetchIdMut,
 pub use shred::{AsyncDispatcher};
 
 pub use storage::{BTreeStorage, CheckStorage, DenseVecStorage, DistinctStorage, FlaggedStorage,
-                  HashMapStorage, InsertResult, NullStorage, ReadStorage, Storage,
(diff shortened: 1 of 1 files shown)


kvark:
> This allows being generic over read and write storages.

Could you elaborate? What user code needs to know about `MaskedStorage`?

torkleyy:
@kvark 

https://github.com/slide-rs/specs/blob/master/src/storage/mod.rs#L165

That code works for both `ReadStorage` and `WriteStorage`, since both `Fetch` and `FetchMut` implement `Deref`.

torkleyy:
> `MaskedStorage` does show up in the public type signatures like `ReadStorage`

I don't think a type def is a signature. But a type in a signature doesn't have to be public; only if it is a struct field. This "feature" is even used to define traits which cannot be implemented outside a library, to allow backwards-compatible addition of methods to a trait.

kvark:
@torkleyy 

> I don't think a type def is a signature. But a type in a signature doesn't have to be public; only if it is a struct field. This "feature" is even used to define traits which cannot be implemented outside a library, to allow backwards-compatible addition of methods to a trait.

I recall fighting with compiler w.r.t what types I don't want to be public when developing earlier three-rs versions. I guess I was doing something wrong, or maybe the compiler became smarter now.

Issue #236, opened 2017-08-06
Update to rayon 0.8


torkleyy:
Let's move to rayon 0.8 before we release a new version.

eldyer:
Hibitset needs rayon 0.8 then, too.
Crossbeam could be bumped to 0.3.

torkleyy:
Seriously?

PR #237, opened 2017-08-08
Bump crossbeam, hibitset and rayon versions.

Binero:
Closes https://github.com/slide-rs/specs/issues/236

Does not bump the specs version number yet. Can be added if ready. Will compile once hibitset has been updated on crates.io.


torkleyy:
I think we can fix this TODO now: https://github.com/slide-rs/specs/blob/master/src/world.rs#L539

torkleyy:
Thank you!

bors r+

PR #238, opened 2017-08-12
Less scary transform for Join implementations, keeps type safety

Aceeri:
Little less scary transformations here, since `mem::transmute` should only be used as a last resort.


src/storage/mod.rs
```diff
@@ -280,8 +280,8 @@ impl<'a, 'e, T, D> Join for &'a mut Storage<'e, T, D>
         // This is horribly unsafe. Unfortunately, Rust doesn't provide a way
```
WaDelma: There is unused `use std::mem;` here.

PR #239, opened 2017-08-14
Add `SortedStorage`

torkleyy:
cc @Rhuagh @Aceeri


torkleyy:
This doesn't quite work out.. I need to change the implementation.

torkleyy:
Postponing, this is no breaking change, therefore no requirement for 0.10.

PR #241, opened 2017-08-23
ParJoin implementation for CheckStorage

WaDelma:
Adresses #190


WaDelma:
I don't think so atleast.

PR #242, opened 2017-08-23
Added missing spaces to RestrictedStorage example

WaDelma:


Aceeri:
I left them out because rustdoc renders those in and that pushes all the code forward 1 space. So I don't really see a need for this.

WaDelma:
I should've tested if that was the problem or this.

torkleyy:
Thank you @WaDelma!

bors r+

PR #243, opened 2017-08-25
Fix Merge System

omni-viral:
Fix `Merge` `System`
`Future`s can't be `poll`ed without `Task`.
`Spawn` creates such `Task` and provide API to run inner `Future`
Without this `Merge` will `panic!` when it'll try to call `Future::poll` except if it is trivial custom `Future`

### DIFF
--- a/src/common.rs
+++ b/src/common.rs
@@ -17,7 +17,6 @@
 //! features = ["common"]
 //! ```
 
-use std::cell::RefCell;
 use std::convert::AsRef;
 use std::error::Error;
 use std::fmt::{Debug, Display, Formatter, Result as FmtResult};
@@ -25,7 +24,7 @@ use std::io::Write;
 use std::marker::PhantomData;
 
 use futures::{Async, Future};
-use futures::executor::{Notify, spawn, Spawn};
+use futures::executor::{Notify, Spawn, spawn};
 
 use {Component, Entity, Entities, FetchMut, Join, RunningTime, System, WriteStorage};
 
@@ -120,7 +119,7 @@ impl Errors {
 /// In case of an error, it will be added to the `Errors` resource.
 pub struct Merge<F> {
     future_type: PhantomData<F>,
-    spawns: Vec<RefCell<(Entity, Spawn<F>)>>,
+    spawns: Vec<(Entity, Spawn<F>)>,
 }
 
 impl<F> Merge<F> {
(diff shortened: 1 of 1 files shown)


src/common.rs
```diff
@@ -117,19 +120,28 @@ impl Errors {
...
 }
 
+struct NotifyIgnore;
+impl Notify for NotifyIgnore {
+    fn notify(&self, _: usize) {
+        /* Intentionally ignore */
+    }
+}
+
+static NOTIFY_IGNORE: &&'static NotifyIgnore = &&NotifyIgnore;
```
torkleyy: No need for lifetimes in static items.

src/common.rs
```diff
@@ -139,26 +151,28 @@ impl<'a, T, F> System<'a> for Merge<F>
...
-        for (e, future) in (&*entities, &mut future).join() {
-            match future.poll() {
-                Ok(Async::Ready(x)) => {
-                    pers.insert(e, x);
-                    delete.push(e);
+        for (e, ()) in (&*entities, &futures.check()).join() {
+            self.spawns.push(RefCell::new((e, spawn(futures.remove(e).unwrap()))));
+        }
+
+        self.spawns.retain(|spawn| {
```
torkleyy: In order to get rid of `RefCell`, we can use

```rust
let len = spawns.len();
let mut del = 0;
{
    for i in 0..len {
        let remove = { /* Do stuff */ };

        if remove {
            del += 1;
        } else if del > 0 {
            spawns.swap(i - del, i);
        }
    }
}
if del > 0 {
    spawns.truncate(len - del);
}
```

You can also just make this a helper function. The fact that `retain` only allows immutable access was a design mistake.

src/common.rs
```diff
@@ -17,13 +17,16 @@
...
 
+use std::cell::RefCell;
 use std::convert::AsRef;
 use std::error::Error;
 use std::fmt::{Debug, Display, Formatter, Result as FmtResult};
 use std::io::Write;
 use std::marker::PhantomData;
 
 use futures::{Async, Future};
+use futures::executor::{Notify, spawn, Spawn};
```
torkleyy: Please reorder so that `spawn` is at the end.

src/common.rs
```diff
@@ -117,19 +120,28 @@ impl Errors {
...
-            tmp: Vec::new(),
+            spawns: Vec::new(),
         }
     }
 }
 
+struct NotifyIgnore;
+impl Notify for NotifyIgnore {
+    fn notify(&self, _: usize) {
+        /* Intentionally ignore */
```
torkleyy: Could use line comment.

torkleyy:
When I tried the system out it worked, so yeah, maybe my `Future` was trivial.

src/common.rs
Xaeroxe: Ha!  That's basically exactly what I wrote for SmallVec's retain function.  I love code re-use.

omni-viral: That is almost like my first variant :smile:

torkleyy:
It would be really nice to have a test / an example using futures, so we can validate it actually works now.

Xaeroxe:
[Shared](http://alexcrichton.com/futures-rs/futures/future/struct.Shared.html) is a really easy way to generate a panic if this isn't working, as it depends on `task::current()`

torkleyy:
Okay, I'm going to merge this, but I would like to add an example before releasing a new version. We could also port this fix back to 0.9 if you want to.

bors r+

PR #250, opened 2017-09-02
Format whole codebase

torkleyy:
The new formatting makes diffs for imports easier to read by using vertical formatting if horizontal doesn't work.


kvark:
Not yet convinced of the vertical import layout, but all the other changes look good!

torkleyy:
CI finally passed :) @kvark import layout has been changed back to normals.

bors r+

PR #251, opened 2017-09-02
LazyUpdate and ParJoin explanations to book

WaDelma:
Note: I haven't actually built the book so can't say if I have done everything right.


book/src/06_system_data.md
```diff
@@ -17,6 +17,39 @@ methods, respectively. After dynamic entity creation / deletion,
 a call `World::maintain` is necessary in order to make the changes
 persistent and delete associated components.
 
-You cannot, however, easily build an entity
-with associated components. For that, you have to write to these component
-storages and insert a component for your entity.
+## Adding and removing components
+
+To add or remove components of existing or dynamically created entity
```
torkleyy: Maybe "of entities"?

book/src/06_system_data.md
```diff
@@ -17,6 +17,39 @@ methods, respectively. After dynamic entity creation / deletion,
...
+    type SystemData = (
+        Entities<'a>,
+        WriteStorage<'a, Stone>,
+        Fetch<'a, LazyUpdate>
+    );
+
+    fn run(&mut self, (entities, stones, updater): Self::SystemData) {
+        let stone = entities.create();
+
+        // 1) Either we insert the component by writing to it's storage
```
torkleyy: Should be "its"

book/src/06_system_data.md
```diff
@@ -17,6 +17,39 @@ methods, respectively. After dynamic entity creation / deletion,
...
+        stones.insert(stone, Stone);
+
+        // 2) or we can lazily insert it with `LazyUpdate`
+        updater.insert(stone, Stone);
+    }
+}
+```
+
+> **Note:** After using adding or removing components using `LazyUpdate`
```
torkleyy: "After ~~using adding or removing components~~ using `LazyUpdate`"

book/src/07_parallel_join.md
```diff
@@ -0,0 +1,45 @@
+# [DRAFT] Parallel Join
+
+As mentioned in chapter dedicated on how to [dispatch][c3] systems
+Specs automatically parallises execution of systems that have
```
torkleyy: IIRC I used British spelling so far, so we should keep it consistent (either everything with AE spelling or BE, I don't care).

book/src/07_parallel_join.md
```diff
@@ -0,0 +1,45 @@
...
+non-conflicting component requirements (mutability XOR aliasing).
+
+What isn't automatically parallised are the joins made within single system:
```
torkleyy: within **a** single system

book/src/07_parallel_join.md
```diff
@@ -0,0 +1,45 @@
...
+
+```rust,ignore
+    fn run(&mut self, (vel, mut pos): Self::SystemData) {
+        use specs::Join;
+        // This loop is ran sequentially on a single thread.
```
torkleyy: I cannot tell if this sentence is correct.

book/src/07_parallel_join.md
```diff
@@ -0,0 +1,45 @@
...
+        for (vel, pos) in (&vel, &mut pos).join() {
+            pos.x += vel.x * 0.05;
+            pos.y += vel.y * 0.05;
+        }
+    }
+```
+
+This means that if there are hundreds of thousands of entities and only few
+systems that can be executed in parallel then the full power
+of CPU cores will get wasted.
```
torkleyy: I think *getting wasted* has another meaning. Maybe we could say "cannot be utilized"?

book/src/07_parallel_join.md
```diff
@@ -0,0 +1,45 @@
...
+
+To fix this inefficiency and to parallelize joining the `join`
+method call can be changed to `par_join` one:
```
torkleyy: I think it would sound better without "one" or with a "the" before `par_join`.

book/src/07_parallel_join.md
```diff
@@ -0,0 +1,45 @@
...
+
+```rust,ignore
+    fn run(&mut self, (vel, mut pos): Self::SystemData) {
+        use specs::ParJoin;
+        use rayon::iter::ParallelIterator;
+
+        // Parallel joining behaves similarly to normal joining
+        // but now each iteration can potentially executed in parallel
```
torkleyy: I think this would better be written as "can be executed".

book/src/07_parallel_join.md
```diff
@@ -0,0 +1,45 @@
...
+        // by a thread pool same way as the systems are executed.
```
torkleyy: Maybe "the same way systems are executed"?

book/src/SUMMARY.md
```diff
@@ -6,5 +6,6 @@
 - [Resources](04_resources.md)
 - [Storages](05_storages.md)
 - [System Data](06_system_data.md)
-- [Rendering](07_rendering.md)
+- [Parallel Join](07_parallel_join.md)
+- [Rendering](08_rendering.md)
```
torkleyy: I think you forgot to delete the old file.

book/src/07_parallel_join.md
minecrawler: Should probably rather be "This loop **runs** sequentially on a single thread."

torkleyy: Oh yes, sounds better.

book/src/07_parallel_join.md
minecrawler: There should be a comma after "joining". Alternatively to @torkleyy 's idea, you can just leave out "one" at the end.
"To fix this inefficiency and to parallelize joining, the `join` method call can be changed to `par_join`:"

book/src/07_parallel_join.md
minecrawler: Instead of "now", I'd use "with the difference that".

book/src/08_rendering.md
```diff
@@ -0,0 +1,54 @@
+# Rendering
+
+Rendering is often a little bit tricky when you're dealing with a multi-threaded ECS.
+That's why we got something called "thread-local systems".
+
+There are two things special about thread-local systems:
```
minecrawler: "... two things to keep in mind about..."?

book/src/SUMMARY.md
WaDelma: Interesting... I renamed the file but apparently I didn't add the removal to the commit...

book/src/08_rendering.md
```diff
@@ -0,0 +1,54 @@
+# Rendering
+
+Rendering is often a little bit tricky when you're dealing with a multi-threaded ECS.
+That's why we got something called "thread-local systems".
```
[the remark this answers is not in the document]
WaDelma: I didn't change anything in the rendering chapter just renamed it, but I'll fix these too.

WaDelma:
Thanks for the suggestions!
I was kinda suprised how many problems the text had.
My non-native english is showing :P

minecrawler:
@WaDelma don't worry about it. I'm German and I'm very sure @torkleyy is, too, so we are no native speakers either. It's just that it's easier to spot someone else's errors than one own's. That's why people do code reviews. ;)

Hmm. thinking about it, there are many places I am not sure about. It's mostly comma related, however my English teacher at work, who is native Irish, says herself that she doesn't know the exact English comma rules. I guess, we should hope that we can find a native speaker who can correct the text and make a high-quality version. However, for now we did what we can, and it's good imho.

torkleyy:
@minecrawler @phowell is native speaker. They once fixed grammar mistakes in the book.

torkleyy:
Guess we can merge this now ;)

bors r+

PR #255, opened 2017-09-09
Use atomic crossbeam structure to add interior mutability to common::Errors

torkleyy:
This allows fetching `Errors` without limiting concurrency.

### DIFF
--- a/src/common.rs
+++ b/src/common.rs
@@ -23,10 +23,11 @@ use std::fmt::{Debug, Display, Formatter, Result as FmtResult};
 use std::io::Write;
 use std::marker::PhantomData;
 
+use crossbeam::sync::MsQueue;
 use futures::{Async, Future};
 use futures::executor::{spawn, Notify, Spawn};
 
-use {Component, Entities, Entity, FetchMut, Join, RunningTime, System, WriteStorage};
+use {Component, Entities, Entity, Fetch, Join, RunningTime, System, WriteStorage};
 
 /// A boxed error implementing `Debug`, `Display` and `Error`.
 pub struct BoxedErr(pub Box<Error + Send + Sync + 'static>);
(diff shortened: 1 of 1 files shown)


torkleyy:
cc @Xaeroxe @omni-viral @ebkalderon This should also improve performance for Amethyst in the future, and is the same strategy we should use for events in case we want to implement them as described in https://github.com/amethyst/amethyst/issues/229

omni-viral:
@torkleyy I believe that crossbeam will add more overhead than collecting errors that `System`s return.

torkleyy:
@omni-viral I don't think so, but the bigger problem I have with `System`s returning `Result` is that it is a breaking change that forces all users to change all of their system implementations and that you have to add `Ok(())` everywhere, even though most systems don't need results. To improve ergonomics, I planned to add a method `Errors::execute`:

```rust
fn run(..) {
    let x = errors.execute(|| {
        let val = do_something_that_can_fail()?;
        
        val + 5
    });

    ints.insert(entity, x);
} 
```

Just look at these systems used in an actual game built with Specs: https://github.com/agmcleod/ld39/tree/master/src/systems

None of them needs `Result`s and their (performance-wise and syntactical) overhead.

torkleyy:
The difference here is also that the crossbeam structure only does something if errors get pushed, but the collecting of `Result`s happens every frame. This means the best-case performance (no error), which is the common one, is better with this solution.

omni-viral:
> This means the best-case performance (no error), which is the common one, is better with this solution.

The good reason I often forget about 😄

src/common.rs
```diff
@@ -73,20 +74,42 @@ pub type BoxedFuture<T> = Box<Future<Item = T, Error = BoxedErr> + Send + Sync +
...
+        !self.errors.is_empty()
+    }
+
+    /// Removes the first error from the queue.
+    pub fn pop_err(&mut self) -> Option<BoxedErr> {
+        self.errors.try_pop()
+    }
+
+    /// Collect all errors into a `Vec`.
+    pub fn collect(&mut self) -> Vec<BoxedErr> {
```
omni-viral: Maybe it is better return draining iterator instead?

torkleyy:
@omni-viral So can I take this as an approval?

src/common.rs
```diff
@@ -73,20 +74,42 @@ pub type BoxedFuture<T> = Box<Future<Item = T, Error = BoxedErr> + Send + Sync +
...
+        let mut errors = Vec::new();
+        while let Some(e) = self.errors.try_pop() {
```
omni-viral: Isn't there a method to pop all queue at once?

torkleyy: No.

src/common.rs
torkleyy: Crossbeam doesn't offer that, but I could still implement it in this module.

src/common.rs
omni-viral: I'll take a look if it may be added 😄

torkleyy:
@omni-viral Done.

torkleyy:
Merging this now, as this is blocking me atm.

bors r+

omni-viral:
Great!

PR #258, opened 2017-09-10
Add benchmark for RudyStorage

### LINKED ISSUE #204: Add judy array as storage type

torkleyy:
Fixes #204

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,7 +33,7 @@ script:
 - cargo build --verbose --features common
 - cargo test --verbose --features common
 - if [ "$TRAVIS_RUST_VERSION" == "nightly" ]; then
-    cargo bench --verbose --no-run;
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -33,7 +33,7 @@ script:
 - cargo build --verbose --features common
 - cargo test --verbose --features common
```
Aceeri: Should we maybe add another test for building with rudy?

WaDelma: There is already commands to build and test with rudy.

torkleyy:
Thanks for reviewing!

bors r+

PR #260, opened 2017-09-10
Use Result in several places and solve double deletion problem

### LINKED ISSUE #94: Deleting entity twice caused assertion error
Example Code:

```
let mut world = World::new();

let entity = world.create_now().build();

world.delete_now(entity);
world.delete_later(entity);

world.maintain();
```

Triggered assertion: https://github.com/slide-rs/specs/blob/master/src/lib.rs#L50

torkleyy:
Not meant to be merged as is

Fixes #94

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -1,0 +1,130 @@
+//! Specs error module.
+//!
+//! There are specific types for errors (e.g. `WrongGeneration`)
+//! and additionally one `Error` type that can represent them all.
+//! Each error in this module has an `Into<Error>` implementation.
+
+use std::error::Error as StdError;
+use std::fmt::{Debug, Display, Formatter, Result as FmtResult};
+
+use world::{Entity, Generation};
+
+/// A boxed error implementing `Debug`, `Display` and `Error`.
+pub struct BoxedErr(pub Box<StdError + Send + Sync + 'static>);
+
+impl BoxedErr {
+    /// Creates a new boxed error.
+    pub fn new<T>(err: T) -> Self
+    where
+        T: StdError + Send + Sync + 'static,
+    {
+        BoxedErr(Box::new(err))
+    }
+}
+
+impl AsRef<StdError> for BoxedErr {
+    fn as_ref(&self) -> &(StdError + 'static) {
+        self.0.as_ref()
+    }
+}
+
+impl Debug for BoxedErr {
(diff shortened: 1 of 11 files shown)


torkleyy:
Okay, I think this is ready for review now.

src/error.rs
```diff
@@ -0,0 +1,130 @@
...
+
+        Some(e)
+    }
+}
+
+/// Wrong generation error.
+#[derive(Debug)]
+pub struct WrongGeneration {
+    /// The action that failed because of the wrong generation.
+    pub action: String,
```
csherratt: Could this be an enum or a &'static string?

src/error.rs
```diff
@@ -0,0 +1,130 @@
...
+/// without breakage.
+#[derive(Debug)]
+pub enum Error {
+    /// A custom, boxed error.
+    Custom(BoxedErr),
+    /// Wrong generation error.
+    WrongGeneration(WrongGeneration),
+
+    #[doc(hidden)]
+    __NonExhaustive,
```
Aceeri: What kind of breakage would occur without this?

torkleyy: If a client matches against all the variants and we add one, it would cause an error about non-exhaustive matching. By having this additional variant which should never be used, the user is forced to match against `_`. This is a very common pattern (even the standard library uses it, see https://doc.rust-lang.org/src/core/sync/atomic.rs.html#220) and there's an RFC about creating an attribute `#[non_exhaustive]` for just this case.

torkleyy:
@kvark A deletion without checks wouldn't really be unsafe, the semantics are just not clearly defined (whether it panics or it's silently ignored). Benchmarking both versions, I cannot see any difference in performance (it may be the case that other checks are eliminated, idk). So if the difference is smaller than `1 ns/iter`, would it still make sense to add a second variant?

kvark:
> A deletion without checks wouldn't really be unsafe, the semantics are just not clearly defined (whether it panics or it's silently ignored)

Even better then - let's have a safe version that ignores the errors. For ergonomics and this undetectable bit of performance ;)

torkleyy:
So should the `_unchecked` version not do the check or just silently ignore a deletion of a dead entity?

torkleyy:
I'm assuming the latter right now.

kvark:
Ok, that's convincing enough. Thank you!
bors r+

PR #262, opened 2017-09-11
Add Errors::execute method

torkleyy:
This allows using ? for error handling.

### DIFF
--- a/src/common.rs
+++ b/src/common.rs
@@ -18,6 +18,7 @@
 //! ```
 
 use std::convert::AsRef;
+use std::error::Error;
 use std::io::Write;
 use std::marker::PhantomData;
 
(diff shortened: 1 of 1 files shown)


src/common.rs
```diff
@@ -105,6 +105,35 @@ impl Errors {
...
+    /// ```
+    ///
+    /// So the closure you pass to this method is essentially an environment where you can
+    /// use `?` for error handling.
+    pub fn execute<E, F>(&self, f: F)
+    where
+        E: Error + Send + Sync + 'static,
+        F: FnOnce() -> Result<(), E>,
+    {
+        if let Err(e) = f() {
```
Aceeri: I'm a bit confused how this builds, does rust allow coercing from T -> () in return values or something?

torkleyy: No, why do you think that would be the case?

torkleyy:
So this is once again something that only bors could catch.

torkleyy:
So let's try again ;)

bors r+

PR #263, opened 2017-09-12
Remove useless lifetime parameter from tuple `impl` heads.

pthariensflame:
This was bugging me while reading the docs; this has absolutely zero possible effect, except to maybe make both `rustc` and outside readers slightly more efficient at scanning the code.

### DIFF
--- a/src/join.rs
+++ b/src/join.rs
@@ -225,7 +225,7 @@ where
 macro_rules! define_open {
     // use variables to indicate the arity of the tuple
     ($($from:ident),*) => {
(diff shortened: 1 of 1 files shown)


pthariensflame:
Why did CI fail?  It doesn't seem to have actually failed at all, from the log, but it's marked as red…

torkleyy:
Oh ah sorry. It's a known issue and I fixed it, but it hasn't been merged yet.

torkleyy:
This should work now.

bors r+

PR #265, opened 2017-09-17
Code coverage using cargo-travis on coveralls.

WaDelma:
This PR adds automatic code coverage testing using `cargo-travis` and `coveralls`.

Note: I cannot add the repo to coveralls as I am not member of the github organization.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,20 @@ rust:
 - nightly
 - stable
 
-cache: cargo
+cache:
+  cargo: true
+  apt: true
+
+addons:
+  apt:
+    packages:
+      - libcurl4-openssl-dev
+      - libelf-dev
+      - libdw-dev
+      - binutils-dev
+      - cmake
+    sources:
+      - kalakris-cmake
 
 branches:
   only:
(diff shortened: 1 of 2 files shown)


torkleyy:
https://travis-ci.org/slide-rs/specs/jobs/276531441#L2188 :

> kcov: warning: kcov: WARNING: kcov has been built without libbfd-dev (or
> binutils-dev), so the --verify option will not do anything.

Should we do anything about that?

WaDelma:
I have created [issue]( https://github.com/roblabla/cargo-travis/issues/15) to `cargo-travis` for it.
Afaik kcovs `--verify` flag needs it to work and if it's not used it [can cause segfaults](https://github.com/SimonKagstrom/kcov/issues/172).

WaDelma:
Good idea. It also tests that my change actually sends the coverage to coveralls.

WaDelma:
You should remove those coveralls messages (I don't know why they are on by default).

Also the coverage report doesn't open the files because it has empty prefix and it should be changed to `src`.

torkleyy:
90%? That's much more than I expected.

torkleyy:
Btw, why remove the coveralls messages?

WaDelma:
It creates after every CI round. So if you push 10 commits in rapid succession it will spam 10 messages to the PR.

torkleyy:
Oh I see.

torkleyy:
Thank you @WaDelma!

bors r+

PR #269, opened 2017-09-23
Fix insertion into occupied `NullStorage` entry

### LINKED ISSUE #266: inserting into a NullStorage slot that already contains a value panics
This is due to [generic code accessing the slot to swap its contents](https://github.com/slide-rs/specs/blob/master/src/storage/mod.rs#L212). Since `NullStorage` contains a copy of the value anyway, perhaps it could just return a (useless) reference instead of panicing on `get_mut`?

torkleyy:
Fixes #266

### DIFF
--- a/src/storage/mod.rs
+++ b/src/storage/mod.rs
@@ -86,7 +86,7 @@ where
 pub unsafe trait DistinctStorage {}
 
 /// The status of an `insert()`ion into a storage.
-#[derive(Debug)]
+#[derive(Debug, PartialEq)]
 pub enum InsertResult<T> {
     /// The value was inserted and there was no value before
     Inserted,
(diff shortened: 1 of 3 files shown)


torkleyy:
r? @Ralith

Sorry for the delay!

Ralith:
Is there a principled basis on which to conclude this approach is actually safe? In the issue I proposed a workaround that doesn't actually require potentially taking multiple mutable references to the same thing in parallel, though I can't imagine how that could go wrong in practice.

torkleyy:
I don't see how this could potentially go wrong. Here's why:

* Nobody reads any bytes from the data
* Nobody writes any bytes to the data
* Even if there was a data race, it wouldn't matter since we don't care about the value

Ralith:
I agree that a data race is unlikely to be an issue. I'm more concerned about whether this is strictly defined behavior, which rust does not actually have pinned down yet, so perhaps it's a futile worry.

torkleyy:
@Ralith Thanks for the review.

bors r+

PR #273, opened 2017-09-24
Add `drain` method to `Storage`

### LINKED ISSUE #264: Draining iterator for storages
On occasion, I find myself writing code like this:

```rust
for (_, eid) in (&foo, &entities).join() {
    if let Some(bar) = bar.remove(eid) {
        // Do something with bar
    }
}
```

Or this:

```rust
// To reduce allocation costs, working_set is owned by the system struct
for (_, eid) in (&bar, &entities).join() {
    self.working_set.push(eid);
}

for eid in self.working_set.drain(..) {
    if let Some(bar) = bar.remove(eid) {
        // ...
    }
}
```

Both of these rub me the wrong way, especially the second pattern with its redundant loop and owned list of entities. It seems to me that this logic should be available as some kind of modifier on the usual iteration patterns, since the storage already maintains a list of which entities have its associated components.

What I'd like to be able to write is something like the following:

```rust
// Remove "bar" from all entities that have it
for bar in bar.drain() {
    // ...
}

// Remove "bar" only from entities that also have "foo"
for (_, bar) in (&foo, bar.drain()).join() {
    // Do something with bar
}
```

I brought this up earlier on IRC earlier. @kvark suggested the name "drain" and @Aceeri suggested implementing it as a new Storage type. I'm not convinced the latter makes sense but I'm open to more suggestions.

torkleyy:
Fixes #264

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 * Make `common::Errors` atomic and more convenient ([#255], [#262])
 * Add `World::delete_all` to clear the world ([#257])
 * Fix insertion into occupied `NullStorage` entry ([#269])
+* Add `Storage::drain` method ([#273])
 
 [#203]: https://github.com/slide-rs/specs/pull/203
 [#243]: https://github.com/slide-rs/specs/pull/243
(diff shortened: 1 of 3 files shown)


torkleyy:
I'm not sure why coverage goes down.. does it really want me to test `drain` for every type of storage?

WaDelma:
You should be able to make coveralls to show the file and which lines are covered by test.
Login on coveralls using github, click file and it should let you fill the where the sources are (which is src for us=

torkleyy:
The problem is that the coverage is inaccurate. E.g. I had to test derived traits, but that doesn't make any sense. I ran the coverage locally and it was not obvious what test I should add, see Gitter.

malleusinferni:
Looks good to me. I tried it out with a few reduced test cases and it does *exactly* what I was asking for. Thanks @torkleyy!

PR #274, opened 2017-09-27
Adds StorageEntry to Storage for convenience

### LINKED ISSUE #268: Entry API for `&mut Storage`
It would be great to have API similar to to `HashMap`s [Entry API](https://doc.rust-lang.org/std/collections/hash_map/struct.HashMap.html#method.entry) for mutable `Storage`.

This common pattern
```rust
if let Some(component) = storage.get(entity) {
    component
} else {
    storage.insert(entity, create_component());
    storage.get(entity).unwrap()
}
```
will be reduced to less error prone
```rust
storage.entry(entity).or_insert_with(create_component)
```

Aceeri:
Fixes #268 

Adds `StorageEntry` to allow for in place insertions, also some other methods for dealing with the occupied and vacant entries, similar to `std`s HashMap Entry.

Still need to add some tests for further verification (I'll do this tomorrow).

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -146,19 +146,3 @@ impl StdError for NoError {
         match *self {}
     }
 }
-
-/// Storage entry error when the request entity was dead.
-#[derive(Debug)]
-pub struct EntryIsDead(pub Entity);
-
-impl Display for EntryIsDead {
-    fn fmt(&self, f: &mut Formatter) -> FmtResult {
-        write!(f, "Attempted to get an entry for entity {:?} after it died.", self.0)
-    }
-}
-
-impl StdError for EntryIsDead {
-    fn description(&self) -> &str {
-        "Attempted to get a storage entry for an entity that is no longer valid/alive"
-    }
-}
--- a/src/error.rs
+++ b/src/error.rs
@@ -129,7 +129,6 @@ impl StdError for WrongGeneration {
     }
 }
 
-
 /// An error type which cannot be instantiated.
 /// Used as a placeholder for associated error types if
 /// something cannot fail.
@@ -145,5 +144,21 @@ impl Display for NoError {
 impl StdError for NoError {
     fn description(&self) -> &str {
         match *self {}
+    }
+}
+
+/// Storage entry error when the request entity was dead.
+#[derive(Debug)]
+pub struct EntryIsDead(pub Entity);
+
+impl Display for EntryIsDead {
+    fn fmt(&self, f: &mut Formatter) -> FmtResult {
+        write!(f, "Attempted to get an entry for entity {:?} after it died.", self.0)
+    }
+}
+
+impl StdError for EntryIsDead {
+    fn description(&self) -> &str {
+        "Attempted to get a storage entry for an entity that is no longer valid/alive"
     }
 }
--- a/src/storage/mod.rs
+++ b/src/storage/mod.rs
@@ -193,6 +193,120 @@ where
     }
 }
 
+/// An entry to a storage which has a component associated to the entity.
+pub struct OccupiedEntry<'a, 'b: 'a, T: 'b, D: 'b> {
+    entity: Entity,
+    storage: &'a mut Storage<'b, T, D>,
+}
+
+impl<'a, 'b, T, D> OccupiedEntry<'a, 'b, T, D>
+    where T: Component,
+          D: Deref<Target = MaskedStorage<T>>,
+{
+    /// Get a reference to the component associated with the entity.
+    pub fn get(&self) -> Option<&T> {
+        self.storage.get(self.entity)
+    }
+}
+
+impl<'a, 'b, T, D> OccupiedEntry<'a, 'b, T, D>
+    where T: Component,
+          D: DerefMut<Target = MaskedStorage<T>>,
+{
+    /// Get a mutable reference to the component associated with the entity.
+    pub fn get_mut(&mut self) -> Option<&mut T> {
+        self.storage.get_mut(self.entity)
+    }
+
+    /// Inserts a value into the storage and returns the old one.
+    ///
+    /// Return value can be `None` if the entity passed in was dead.
+    pub fn insert(&mut self, component: T) -> Option<T> {
+        match self.storage.insert(self.entity, component) {
+            InsertResult::Updated(old) => Some(old),
+            _ => None,
+        }
+    }
+
+    /// Removes the component from the storage and returns it.
+    ///
+    /// Return value can be `None` if the entity passed in was dead.
+    pub fn remove(&mut self) -> Option<T> {
+        self.storage.remove(self.entity)
+    }
+}
+
+/// An entry to a storage which does not have a component associated to the entity.
+pub struct VacantEntry<'a, 'b: 'a, T: 'b, D: 'b> {
+    entity: Entity,
+    storage: &'a mut Storage<'b, T, D>,
+}
+
+impl<'a, 'b, T, D> VacantEntry<'a, 'b, T, D>
+    where T: Component,
+          D: DerefMut<Target = MaskedStorage<T>>,
+{
+    /// Inserts a value into the storage and returns the old one.
+    ///
+    /// Return value can be `None` if the entity passed in was dead.
+    pub fn insert(&mut self, component: T) -> Option<T> {
+        match self.storage.insert(self.entity, component) {
+            InsertResult::Updated(old) => Some(old),
+            _ => None,
+        }
+    }
+}
+
+/// Entry to a storage for convenient filling of components or removal based on whether
+/// the entity has a component.
+pub enum StorageEntry<'a, 'b: 'a, T: 'b, D: 'b> {
+    /// Entry variant that is returned if the entity does has a component.
+    Occupied(OccupiedEntry<'a, 'b, T, D>),
+    /// Entry variant that is returned if the entity does not have a component.
+    Vacant(VacantEntry<'a, 'b, T, D>),
+}
+
+impl<'a, 'b, T, D> StorageEntry<'a, 'b, T, D>
+    where T: Component,
+          D: DerefMut<Target = MaskedStorage<T>>
+{
+
+    /// Inserts a component if the entity does not contain a component.
+    ///
+    /// Returns `None` if the entity passed in to get the `Entry` was dead.
+    pub fn or_insert(self, component: T) -> Option<&'a mut T> {
+        match self {
+            StorageEntry::Occupied(occupied) => occupied.storage.get_mut(occupied.entity),
+            StorageEntry::Vacant(vacant) => {
+                let insert_result = vacant.storage.insert(vacant.entity, component);
+                if let InsertResult::EntityIsDead(_) = insert_result {
+                    None
+                }
+                else {
+                    vacant.storage.get_mut(vacant.entity)
+                }
+            }
+        }
+    }
+
+    /// Inserts a component using a default function.
+    ///
+    /// Returns `None` if the entity passed in to get the `Entry` was dead.
+    pub fn or_insert_with<F>(self, component: F) -> Option<&'a mut T>
+        where F: FnOnce() -> T,
+    {
+        self.or_insert(component())
+    }
+
+    /// Returns the entity of the current entry.
+    pub fn entity(&self) -> &Entity {
+        match self {
+            &StorageEntry::Occupied(ref occupied) => &occupied.entity,
+            &StorageEntry::Vacant(ref vacant) => &vacant.entity,
+        }
+    }
+}
+
 impl<'e, T, D> Storage<'e, T, D>
 where
     T: Component,
@@ -204,6 +318,20 @@ where
             Some(unsafe { self.data.inner.get_mut(e.id()) })
         } else {
             None
+        }
+    }
+
+    /// Returns an entry to the component associated to the entity.
+    ///
+    /// Note: This does not immediately error if the entity is dead.
+    pub fn entry<'a>(&'a mut self, e: Entity) -> StorageEntry<'a, 'e, T, D>
+        where 'e: 'a,
+    {
+        if self.data.mask.contains(e.id()) && self.entities.is_alive(e) {
+            StorageEntry::Occupied(OccupiedEntry { entity: e, storage: self })
+        }
+        else {
+            StorageEntry::Vacant(VacantEntry { entity: e, storage: self })
         }
     }
 
--- a/src/storage/mod.rs
+++ b/src/storage/mod.rs
@@ -20,6 +20,7 @@ use shred::Fetch;
 
 use self::drain::Drain;
 use {Component, EntitiesRes, Entity, Index, Join, ParJoin};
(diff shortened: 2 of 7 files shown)


src/storage/mod.rs
```diff
@@ -203,6 +317,20 @@ where
         }
     }
 
+    /// Returns an entry to the component associated to the entity.
+    ///
+    /// Note: This does not immediately error if the entity is dead.
```
torkleyy: I think we should use `Result` here.

Aceeri: Are we sure? That would make this require a bit more boilerplate for this stuff then:
```rust
if let Ok(entry) = storage.entry(entity) {
    entry.or_insert(component);
}
```
At least until we get `try` to work outside of just functions.

omni-viral: And now the boilerplate is at another location
```
if let Some(component) = storage.entry(entity).or_insert(...) {
    ...
}
```

src/storage/mod.rs
```diff
@@ -189,6 +189,120 @@ where
...
+        }
+    }
+
+    /// Inserts a component using a default function.
+    ///
+    /// Returns `None` if the entity passed in to get the `Entry` was dead.
+    pub fn or_insert_with<F>(self, component: F) -> Option<&'a mut T>
+        where F: FnOnce() -> T,
+    {
+        self.or_insert(component())
```
omni-viral: The whole idea of the method is not to call closure unless component is required.
Instead `or_insert` should be implemented via this one.

src/storage/mod.rs
Aceeri: I'm somewhat worried about enforcing that `Entry` is only created with a live `Entity`, is there a case where an entity could be deleted after you create an entry?

omni-viral: In another thread. But it is not visible until maintenance.

omni-viral: Shrödinger entity

omni-viral: It can be killed atomically. But atomically killed enities are considered alive until maintenance.

Aceeri: True, I messed around with it and couldn't find an easy way to kill the entity while having an entry, so I guess it should be fine.

src/storage/mod.rs
```diff
@@ -203,6 +317,25 @@ where
         }
     }
 
+    /// Returns an entry to the component associated to the entity.
+    ///
```
torkleyy: Please add a usage example here.

src/storage/mod.rs
```diff
@@ -189,6 +189,120 @@ where
...
+                }
+                else {
+                    vacant.storage.get_mut(vacant.entity).unwrap()
+                }
+            }
+        }
+    }
+
+    /// Returns the entity of the current entry.
+    pub fn entity(&self) -> &Entity {
```
torkleyy: Is there a reason for us to give a reference to the entity?

src/storage/mod.rs
```diff
@@ -189,6 +189,120 @@ where
...
+    where T: Component,
+          D: DerefMut<Target = MaskedStorage<T>>
+{
+
+    /// Inserts a component if the entity does not contain a component.
+    pub fn or_insert(self, component: T) -> &'a mut T {
+        self.or_insert_with(|| component)
+    }
+
+    /// Inserts a component using a default function.
```
torkleyy: "using a default function"? Why "default"?

src/storage/mod.rs
```diff
@@ -189,6 +189,120 @@ where
...
+/// Entry to a storage for convenient filling of components or removal based on whether
+/// the entity has a component.
+pub enum StorageEntry<'a, 'b: 'a, T: 'b, D: 'b> {
+    /// Entry variant that is returned if the entity does has a component.
+    Occupied(OccupiedEntry<'a, 'b, T, D>),
+    /// Entry variant that is returned if the entity does not have a component.
+    Vacant(VacantEntry<'a, 'b, T, D>),
+}
+
+pub enum EntryError {
```
torkleyy: This belongs into the `error` module. Also, we should not have enums with just a single variant.

src/storage/mod.rs
```diff
@@ -189,6 +189,120 @@ where
...
+          D: DerefMut<Target = MaskedStorage<T>>,
+{
+    /// Get a mutable reference to the component associated with the entity.
+    pub fn get_mut(&mut self) -> Option<&mut T> {
+        self.storage.get_mut(self.entity)
+    }
+
+    /// Inserts a value into the storage and returns the old one.
+    ///
+    /// Return value can be `None` if the entity passed in was dead.
```
torkleyy: Is this still true? Given that we now return an error, shouldn't this return just `T`?

src/storage/mod.rs
```diff
@@ -189,6 +189,120 @@ where
...
+        match self.storage.insert(self.entity, component) {
+            InsertResult::Updated(old) => Some(old),
+            _ => None,
+        }
+    }
+
+    /// Removes the component from the storage and returns it.
+    ///
+    /// Return value can be `None` if the entity passed in was dead.
+    pub fn remove(&mut self) -> Option<T> {
```
torkleyy: Same here.

src/storage/mod.rs
```diff
@@ -189,6 +189,120 @@ where
...
+    storage: &'a mut Storage<'b, T, D>,
+}
+
+impl<'a, 'b, T, D> VacantEntry<'a, 'b, T, D>
+    where T: Component,
+          D: DerefMut<Target = MaskedStorage<T>>,
+{
+    /// Inserts a value into the storage and returns the old one.
+    ///
+    /// Return value can be `None` if the entity passed in was dead.
```
torkleyy: Same here

src/storage/mod.rs
Aceeri: This isn't true, thought I removedid that line.

src/storage/mod.rs
Aceeri: Default should probaby be "lazily called function"

src/error.rs
```diff
@@ -147,3 +146,19 @@ impl StdError for NoError {
         match *self {}
     }
 }
+
+/// Storage entry error when the request entity was dead.
+#[derive(Debug)]
+pub struct EntryIsDead(pub Entity);
```
torkleyy: Why aren't you just using `WrongGeneration`? And in case this actually needs its own error, it should still be representable with `specs::error::Error` (see `WrongGeneration`; the idea is that every error can be converted to that enum).

src/storage/mod.rs
```diff
@@ -191,6 +191,108 @@ where
...
+    entity: Entity,
+    storage: &'a mut Storage<'b, T, D>,
+}
+
+impl<'a, 'b, T, D> OccupiedEntry<'a, 'b, T, D>
+    where T: Component,
+          D: Deref<Target = MaskedStorage<T>>,
+{
+    /// Get a reference to the component associated with the entity.
+    pub fn get(&self) -> Option<&T> {
```
torkleyy: If the entry is occupied, how could this call return `None`?

src/storage/mod.rs
```diff
@@ -191,6 +191,108 @@ where
...
+
+impl<'a, 'b, T, D> OccupiedEntry<'a, 'b, T, D>
+    where T: Component,
+          D: DerefMut<Target = MaskedStorage<T>>,
+{
+    /// Get a mutable reference to the component associated with the entity.
+    pub fn get_mut(self) -> &'a mut T {
+        match self.storage.get_mut(self.entity) {
+            Some(component) => component,
+            None => { panic!("`OccupiedEntry` on an entity without a component."); },
```
torkleyy: No need for `{` and `}`

src/storage/mod.rs
```diff
@@ -191,6 +191,108 @@ where
...
+            InsertResult::Inserted => { panic!("`OccupiedEntry` on an entity without a component."); },
+            InsertResult::EntityIsDead(_) => { panic!("`OccupiedEntry` on a dead entity: {:?}.", self.entity); },
+        }
+    }
+
+    /// Removes the component from the storage and returns it.
+    pub fn remove(self) -> T {
+        match self.storage.remove(self.entity) {
+            Some(old) => old,
+            None => { panic!("`OccupiedEntry` on an entity without a component."); },
```
torkleyy: Meh, the whole sense of Entry API is that you usually don't do any checks twice which is the case for all these methods :(

src/storage/mod.rs
```diff
@@ -191,6 +191,108 @@ where
...
+}
+
+impl<'a, 'b, T, D> VacantEntry<'a, 'b, T, D>
+    where T: Component,
+          D: DerefMut<Target = MaskedStorage<T>>,
+{
+    /// Inserts a value into the storage.
+    pub fn insert(self, component: T) -> &'a mut T {
+        match self.storage.insert(self.entity, component) {
+            InsertResult::Updated(_) => { panic!("`VacantEntry` on an entity with a component."); },
```
torkleyy: No need for `{` and `}`

src/storage/mod.rs
```diff
@@ -191,6 +191,108 @@ where
...
+            InsertResult::Inserted => {
+                match self.storage.get_mut(self.entity) {
+                    Some(component) => component,
+                    None => { panic!("Storage insertion into entity {:?} failed on `VacantEntry`", self.entity); },
+                }
+            },
+            InsertResult::EntityIsDead(_) => { panic!("`VacantEntry` on a dead entity: {:?}.", self.entity); },
```
torkleyy: Same here

src/storage/mod.rs
```diff
@@ -205,6 +307,46 @@ where
         }
     }
 
+    /// Returns an entry to the component associated to the entity.
+    ///
+    /// Behaves somewhat similarly to `std::collections::HashMap`'s entry api.
+    ///
+    ///## Example
```
torkleyy: missing space

src/storage/mod.rs
```diff
@@ -205,6 +307,46 @@ where
...
+    /// 
+    /// ```rust
+    ///# extern crate specs;
```
torkleyy: missing space in this and the following lines

src/storage/mod.rs
Aceeri: Ah, missed that one when removing all the others.

src/storage/mod.rs
torkleyy: I know why the coverage goes down: coveralls thinks that this is an uncovered line, but that's not the case. Speaking of which, this should use `unreachable!`.

src/storage/mod.rs
torkleyy: That's strange; our other examples have spaces, too.

torkleyy:
Are you still interested in this feature? I'm not sure now if I should be waiting for it, that's why I'm asking.

Aceeri:
I'll finish it up, just been kind of got carried away with othe stuff

src/storage/mod.rs
Aceeri: Alright, seems to work now?

src/storage/mod.rs
```diff
@@ -191,6 +191,96 @@ where
...
+          D: DerefMut<Target = MaskedStorage<T>>,
+{
+    /// Get a mutable reference to the component associated with the entity.
+    pub fn get_mut(self) -> &'a mut T {
+        unsafe { self.storage.data.inner.get_mut(self.entity.id()) }
+    }
+
+    /// Inserts a value into the storage and returns the old one.
+    pub fn insert(&mut self, mut component: T) -> T {
+        std::mem::swap(&mut component, unsafe { self.storage.data.inner.get_mut(self.entity.id()) });
```
kvark: could use `self.get_mut()`

src/storage/mod.rs
```diff
@@ -191,6 +191,96 @@ where
...
+        component
+    }
+
+    /// Removes the component from the storage and returns it.
+    pub fn remove(self) -> T {
```
kvark: random impractical thought: would be fun to return `(T, VacantEntry<...>)`

src/storage/mod.rs
```diff
@@ -191,6 +191,96 @@ where
...
+/// the entity has a component.
+pub enum StorageEntry<'a, 'b: 'a, T: 'a, D: 'a> {
+    /// Entry variant that is returned if the entity does has a component.
+    Occupied(OccupiedEntry<'a, 'b, T, D>),
+    /// Entry variant that is returned if the entity does not have a component.
+    Vacant(VacantEntry<'a, 'b, T, D>),
+}
+
+impl<'a, 'b, T, D> StorageEntry<'a, 'b, T, D>
+    where T: Component,
```
kvark: is this an actual formatting of rustfmt-nightly?

src/storage/mod.rs
```diff
@@ -191,6 +191,96 @@ where
...
+{
+
+    /// Inserts a component if the entity does not contain a component.
+    pub fn or_insert(self, component: T) -> &'a mut T {
+        self.or_insert_with(|| component)
+    }
+
+    /// Inserts a component using a lazily called function that is only called
+    /// when inserting the component.
+    pub fn or_insert_with<F>(self, component: F) -> &'a mut T
```
kvark: I think the `F`-typed argument should be named something other than `component`? stdlib uses `default`

src/storage/mod.rs
```diff
@@ -205,6 +295,55 @@ where
...
+    /// # let mut world = specs::World::new();
+    /// # world.register::<Comp>();
+    /// # let entity = world.create_entity().build();
+    /// # let mut storage = world.write::<Comp>();
+    ///  if let Ok(entry) = storage.entry(entity) {
+    ///      entry.or_insert(Comp { field: 55 });
+    ///  }
+    /// # }
+    /// ```
+    pub fn entry<'a>(&'a mut self, e: Entity) -> Result<StorageEntry<'a, 'e, T, D>, ::error::WrongGeneration>
```
kvark: any reason we need to use a global type path here?

src/storage/mod.rs
```diff
@@ -205,6 +295,55 @@ where
...
+            }
+            else {
+                Ok(StorageEntry::Vacant(VacantEntry { entity: e, storage: self }))
+            }
+        }
+        else {
+            let gen = self.entities
+                .alloc
+                .generations.get(e.id() as usize)
+                .map(|gen| *gen)
```
kvark: nit: could be `cloned`?

src/storage/mod.rs
torkleyy: No, I don't think so

src/storage/mod.rs
Aceeri: Annoyingly I can't figure out the lifetimes that would work for `get_mut` that allows just `&mut self`, so I can't get that to work properly.

torkleyy:
Only conflicts left to resolve :+1: Once that's done, feel free to merge.

bors delegate=@Aceeri

Aceeri:
bors cancel

Aceeri:
Should probably add this to the changelog.

bors r-

PR #275, opened 2017-09-27
Add an ability to save and load entities.

omni-viral:
Add (de)serialization support for entities to specs.
Main parts are:
* `Marker` trait.
`Marker` is used for several purposes:
  * Mark which entities should be serialized.
  * It contains ID to replace `Entity` stored in `Component`s during serialization process.
  * It may contain additional data to be used for tracking owner of the component.
* `MarkerAllocator` trait for allocating `Marker`s, storing `Marker -> Entity` mapping and other stuff.
* `SerializableComponent` trait that is auto implemented by components if they implements `Serialize`, `DeserializeOwned` and `Copy` (later is rather strict, maybe it should be replaced with `Clone` or removed somehow).
Mainly `Component` implements `SerializableComponent` manually when it contains `Entity`.
`Component` should replace `Entity` with provided `Marker` in `SerializableComponent::save` function. And replace `Marker` with `Entity` in `SerializableComponent::load` function.
* `Components` trait is implemented for tuples of `SerializableComponent`s
It allows to save all existing and mentioned in tuple components of one entity from `ReadStorage`s into serializable structure. And load it into `WriteStorage`s again.

Helper types `WorldSerialize` and `WorldDeserialize` are good example how to use this facility.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -128,3 +128,22 @@ impl StdError for WrongGeneration {
          (e.g. because the entity has been deleted)"
     }
 }
+
+
+/// An error type which cannot be instantiated.
+/// Used as a placeholder for associated error types if
+/// something cannot fail.
+#[derive(Debug)]
+pub enum NoError {}
+
+impl Display for NoError {
+    fn fmt(&self, _: &mut Formatter) -> FmtResult {
+        match *self {}
+    }
+}
+
+impl StdError for NoError {
+    fn description(&self) -> &str {
+        match *self {}
+    }
+}
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -193,6 +193,9 @@ extern crate futures;
 #[cfg(feature = "serde")]
 #[macro_use]
 extern crate serde;
+#[cfg(feature = "serde")]
+#[macro_use]
+extern crate shred_derive;
 
 #[cfg(feature = "rudy")]
 extern crate rudy;
@@ -213,6 +216,9 @@ pub use world::{Component, CreateIter, CreateIterAtomic, EntitiesRes, Entity, En
 
 #[cfg(feature = "common")]
 pub mod common;
+
+#[cfg(feature = "serde")]
+pub mod saveload;
 
 #[cfg(feature = "rudy")]
 pub use storage::RudyStorage;
--- a/src/saveload/de.rs
+++ b/src/saveload/de.rs
@@ -1,4 +1,3 @@
-
 use std::fmt::{self, Display, Formatter};
 use std::marker::PhantomData;
 
@@ -47,7 +46,7 @@ where
         let ids = |marker: M::Identifier| Some(allocator.get_marked(marker, entities, markers));
         match T::load(entity, data.components, storages, ids) {
             Ok(()) => Ok(()),
-            Err(err) => Err(de::Error::custom(err))
+            Err(err) => Err(de::Error::custom(err)),
         }
     }
 }
@@ -83,7 +82,9 @@ where
             markers: self.markers,
             allocator: self.allocator,
             pd: self.pd,
-        })?.is_some() {}
+        })?
+            .is_some()
+        {}
 
         Ok(())
     }
--- a/src/saveload/de.rs
+++ b/src/saveload/de.rs
@@ -1,0 +1,147 @@
+
+use std::fmt::{self, Display, Formatter};
+use std::marker::PhantomData;
+
+use serde::de::{self, Deserialize, DeserializeSeed, Deserializer, SeqAccess, Visitor};
+
+use {Entities, FetchMut, WriteStorage};
+
+use saveload::{Components, EntityData, Storages};
+use saveload::marker::{Marker, MarkerAllocator};
+
+
+/// Wrapper for `Entity` and tuple of `WriteStorage`s that implements `serde::Deserialize`
+struct DeserializeEntity<'a, 'b: 'a, M: Marker, E, T: Components<M::Identifier, E>> {
+    entities: &'a Entities<'b>,
+    storages: &'a mut <T as Storages<'b>>::WriteStorages,
+    markers: &'a mut WriteStorage<'b, M>,
+    allocator: &'a mut FetchMut<'b, M::Allocator>,
+    pd: PhantomData<(M, E, T)>,
+}
+
+impl<'de, 'a, 'b: 'a, M, E, T> DeserializeSeed<'de> for DeserializeEntity<'a, 'b, M, E, T>
+where
+    M: Marker,
+    E: Display,
+    T: Components<M::Identifier, E>,
+{
+    type Value = ();
+    fn deserialize<D>(self, deserializer: D) -> Result<(), D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        let DeserializeEntity {
+            entities,
+            storages,
+            markers,
+            allocator,
+            ..
+        } = self;
+        let data = EntityData::<M, E, T>::deserialize(deserializer)?;
+        let entity = allocator.get_marked(data.marker.id(), entities, markers);
+        markers
+            .get_mut(entity)
+            .ok_or("Allocator is broken")
+            .map_err(de::Error::custom)?
+            .update(data.marker);
+        let ids = |marker: M::Identifier| Some(allocator.get_marked(marker, entities, markers));
+        match T::load(entity, data.components, storages, ids) {
+            Ok(()) => Ok(()),
+            Err(err) => Err(de::Error::custom(err))
+        }
+    }
+}
+
+/// Wrapper for `Entities` and tuple of `WriteStorage`s that implements `serde::de::Visitor`
+struct VisitEntities<'a, 'b: 'a, M: Marker, E, T: Components<M::Identifier, E>> {
+    entities: &'a Entities<'b>,
+    storages: &'a mut <T as Storages<'b>>::WriteStorages,
+    markers: &'a mut WriteStorage<'b, M>,
+    allocator: &'a mut FetchMut<'b, M::Allocator>,
+    pd: PhantomData<(M, E, T)>,
+}
+
+impl<'de, 'a, 'b: 'a, M, E, T> Visitor<'de> for VisitEntities<'a, 'b, M, E, T>
+where
+    M: Marker,
+    E: Display,
+    T: Components<M::Identifier, E>,
+{
+    type Value = ();
+
+    fn expecting(&self, formatter: &mut Formatter) -> fmt::Result {
+        write!(formatter, "Sequence of serialized entities")
+    }
+
+    fn visit_seq<A>(mut self, mut seq: A) -> Result<(), A::Error>
+    where
+        A: SeqAccess<'de>,
+    {
+        while seq.next_element_seed(DeserializeEntity {
+            entities: self.entities,
+            storages: self.storages,
+            markers: self.markers,
+            allocator: self.allocator,
+            pd: self.pd,
+        })?.is_some() {}
+
+        Ok(())
+    }
+}
+
+
+/// Deserialize entities
+pub fn deserialize<'a, 'de, D, M, E, T>(
+    entities: &Entities<'a>,
+    storages: &mut <T as Storages<'a>>::WriteStorages,
+    markers: &mut WriteStorage<'a, M>,
+    allocator: &mut FetchMut<'a, M::Allocator>,
+    deserializer: D,
+) -> Result<(), D::Error>
+where
+    M: Marker,
+    E: Display,
+    T: Components<M::Identifier, E>,
+    D: Deserializer<'de>,
+{
+    deserializer.deserialize_seq(VisitEntities::<M, E, T> {
+        entities,
+        storages,
+        markers,
+        allocator,
+        pd: PhantomData,
+    })
+}
+
+
+/// `DeerializeSeed` implementation for `World`
+#[derive(SystemData)]
+pub struct WorldDeserialize<'a, M: Marker, E, T: Components<M::Identifier, E>> {
+    entities: Entities<'a>,
+    storages: <T as Storages<'a>>::WriteStorages,
+    markers: WriteStorage<'a, M>,
+    allocator: FetchMut<'a, M::Allocator>,
+    pd: PhantomData<E>,
+}
+
+impl<'de, 'a, M, E, T> DeserializeSeed<'de> for WorldDeserialize<'a, M, E, T>
+where
+    M: Marker,
+    E: Display,
+    T: Components<M::Identifier, E>,
+{
+    type Value = ();
+
+    fn deserialize<D>(mut self, deserializer: D) -> Result<(), D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        deserialize::<D, M, E, T>(
+            &mut self.entities,
+            &mut self.storages,
+            &mut self.markers,
+            &mut self.allocator,
+            deserializer,
+        )
+    }
+}
--- a/src/saveload/de.rs
+++ b/src/saveload/de.rs
@@ -16,7 +16,7 @@ struct DeserializeEntity<'a, 'b: 'a, M: Marker, E, T: Components<M::Identifier, 
     storages: &'a mut <T as Storages<'b>>::WriteStorages,
     markers: &'a mut WriteStorage<'b, M>,
     allocator: &'a mut FetchMut<'b, M::Allocator>,
-    pd: PhantomData<(M, E, T)>,
+    pd: PhantomData<(E, T)>,
 }
 
 impl<'de, 'a, 'b: 'a, M, E, T> DeserializeSeed<'de> for DeserializeEntity<'a, 'b, M, E, T>
@@ -58,7 +58,7 @@ struct VisitEntities<'a, 'b: 'a, M: Marker, E, T: Components<M::Identifier, E>> 
     storages: &'a mut <T as Storages<'b>>::WriteStorages,
     markers: &'a mut WriteStorage<'b, M>,
     allocator: &'a mut FetchMut<'b, M::Allocator>,
-    pd: PhantomData<(M, E, T)>,
+    pd: PhantomData<(E, T)>,
 }
 
 impl<'de, 'a, 'b: 'a, M, E, T> Visitor<'de> for VisitEntities<'a, 'b, M, E, T>
--- a/src/saveload/details.rs
+++ b/src/saveload/details.rs
@@ -5,8 +5,8 @@ use serde::ser::Serialize;
 
 use {Component, Entity, ReadStorage, SystemData, WriteStorage};
 
-use saveload::marker::Marker;
 use error::NoError;
+use saveload::marker::Marker;
 
 #[derive(Serialize, Deserialize)]
 #[serde(bound = "")]
--- a/src/saveload/details.rs
+++ b/src/saveload/details.rs
@@ -142,8 +142,10 @@ macro_rules! impl_components {
         impl_components!(@ $($a|$b),*);
     };
 
+    // List depleted. End of recursion
     (@) => {};
 
+    // Cut head of the list and call macro again
     (@ $ah:ident|$bh:ident $(,$a:ident|$b:ident)*) => {
         // Call again for tail
         impl_components!($($a|$b),*);
--- a/src/saveload/details.rs
+++ b/src/saveload/details.rs
@@ -1,0 +1,169 @@
+use std::error::Error;
+
+use serde::de::DeserializeOwned;
+use serde::ser::Serialize;
+
+use {Component, Entity, ReadStorage, SystemData, WriteStorage};
+
+use saveload::marker::Marker;
+use error::NoError;
+
+#[derive(Serialize, Deserialize)]
+#[serde(bound = "")]
+pub struct EntityData<M: Marker, E, T: Components<M::Identifier, E>> {
+    pub marker: M,
+    pub components: T::Data,
+}
+
+/// This trait should be implemented in order to allow component
+/// to be serializeble with `SerializeSystem`.
+/// It is automatically implemented for all
+/// `Component + DeserializeOwned + Serialize + Copy`
+pub trait SaveLoadComponent<M>: Component {
+    /// Serializable data representation for component
+    type Data: Serialize + DeserializeOwned;
+
+    /// Error may occur duing serialization or deserialization of component
+    type Error: Error;
+
+    /// Convert this component into serializable form (`Data`) using
+    /// entity to marker mapping function
+    fn save<F>(&self, ids: F) -> Result<Self::Data, Self::Error>
+    where
+        F: FnMut(Entity) -> Option<M>;
+
+    /// Convert this component into deserializable form (`Data`) using
+    /// marker to entity mapping function
+    fn load<F>(data: Self::Data, ids: F) -> Result<Self, Self::Error>
+    where
+        F: FnMut(M) -> Option<Entity>;
+}
+
+impl<C, M> SaveLoadComponent<M> for C
+where
+    C: Component + DeserializeOwned + Serialize + Copy,
+{
+    type Data = Self;
+    type Error = NoError;
+
+    fn save<F>(&self, _ids: F) -> Result<Self::Data, NoError> {
+        Ok(*self)
+    }
+
+    fn load<F>(data: Self, _ids: F) -> Result<Self, NoError> {
+        Ok(data)
+    }
+}
+
+/// Helper trait defines storages tuples for components tuple
+pub trait Storages<'a> {
+    /// Storages for read
+    type ReadStorages: SystemData<'a> + 'a;
+    /// Storages for write
+    type WriteStorages: SystemData<'a> + 'a;
+}
+
+/// This trait is implemented by any tuple where all elements are
+/// `Component + Serialize + DeserializeOwned`
+pub trait Components<M, E>: for<'a> Storages<'a> {
+    /// Serializable and deserializable intermediate representation
+    type Data: Serialize + DeserializeOwned;
+
+    /// Saves `Component`s of entity into `Data` serializable representation
+    fn save<'a, F>(
+        entity: Entity,
+        storages: &<Self as Storages<'a>>::ReadStorages,
+        ids: F,
+    ) -> Result<Self::Data, E>
+    where
+        F: FnMut(Entity) -> Option<M>;
+
+    /// Loads `Component`s to entity from `Data` deserializable representation
+    fn load<'a, F>(
+        entity: Entity,
+        components: Self::Data,
+        storages: &mut <Self as Storages<'a>>::WriteStorages,
+        ids: F,
+    ) -> Result<(), E>
+    where
+        F: FnMut(M) -> Option<Entity>;
+}
+
+macro_rules! impl_components {
+    ($($a:ident|$b:ident),*) => {
+        impl<'a, $($a),*> Storages<'a> for ($($a,)*)
+            where $(
+                $a: Component,
+            )*
+        {
+            type ReadStorages = ($(ReadStorage<'a, $a>,)*);
+            type WriteStorages = ($(WriteStorage<'a, $a>,)*);
+        }
+
+        impl<M, E $(,$a)*> Components<M, E> for ($($a,)*)
+            where $(
+                $a: SaveLoadComponent<M>,
+                E: From<$a::Error>,
+            )*
+        {
+            type Data = ($(Option<$a::Data>,)*);
+
+            #[allow(unused_variables, unused_mut, non_snake_case)]
+            fn save<'a, F>(entity: Entity, storages: &($(ReadStorage<'a, $a>,)*), mut ids: F)
+                -> Result<($(Option<$a::Data>,)*), E>
+                where F: FnMut(Entity) -> Option<M>
+            {
+                let ($(ref $b,)*) = *storages;
+                Ok(($(
+                    $b.get(entity).map(|c| c.save(&mut ids).map(Some)).unwrap_or(Ok(None))?,
+                )*))
+            }
+
+            #[allow(unused_variables, unused_mut, non_snake_case)]
+            fn load<'a, F>(entity: Entity, components: ($(Option<$a::Data>,)*),
+                           storages: &mut ($(WriteStorage<'a, $a>,)*), mut ids: F)
+                -> Result<(), E>
+                where F: FnMut(M) -> Option<Entity>
+            {
+                let ($($a,)*) = components;
+                let ($(ref mut $b,)*) = *storages;
+                $(
+                    if let Some(a) = $a {
+                        $b.insert(entity, $a::load(a, &mut ids)?);
+                    } else {
+                        $b.remove(entity);
+                    }
+                )*
+                Ok(())
+            }
+        }
+
+        // Recursivly implement for smaller tuple
+        impl_components!(@ $($a|$b),*);
+    };
+
+    (@) => {};
+
+    (@ $ah:ident|$bh:ident $(,$a:ident|$b:ident)*) => {
+        // Call again for tail
+        impl_components!($($a|$b),*);
+    };
+}
+
+impl_components!(
+    LA | LB,
+    MA | MB,
+    NA | NB,
+    OA | OB,
+    PA | PB,
+    QA | QB,
+    RA | RB,
+    SA | SB,
+    TA | TB,
+    UA | UB,
+    VA | VB,
+    WA | WB,
+    XA | XB,
+    YA | YB,
+    ZA | ZB
+);
--- a/src/saveload/marker.rs
+++ b/src/saveload/marker.rs
@@ -4,11 +4,11 @@ use std::collections::HashMap;
 use std::fmt::Debug;
 use std::hash::Hash;
 
-use shred::Resource;
 use {Component, DenseVecStorage, Entities, Entity, Join, ReadStorage, WriteStorage};
+use shred::Resource;
 
-use serde::ser::Serialize;
 use serde::de::DeserializeOwned;
+use serde::ser::Serialize;
 
 
 /// This trait should be implemetened by a component which is going to be used as marker.
@@ -18,7 +18,7 @@ use serde::de::DeserializeOwned;
 ///
 /// ## Example
 ///
-/// ```rust,no_run
+/// ```rust
 /// extern crate specs;
 /// #[macro_use] extern crate serde;
 /// use std::collections::HashMap;
@@ -78,7 +78,22 @@ use serde::de::DeserializeOwned;
 ///     }
 /// }
 ///
-/// fn main() {}
+/// fn main() {
+///     use specs::World;
+///
+///     let mut world = World::new();
+///     world.register::<NetMarker>();
+///
+///     let mut node = NetNode {
+///         range: 0..100,
+///         mapping: HashMap::new(),
+///     };
+///
+///     let entity = world.create_entity().build();
+///     let (marker, added) = node.mark(entity, &mut world.write::<NetMarker>());
+///     assert!(added);
+///     assert_eq!(node.get_marked(marker.id(), &world.entities(), &mut world.write::<NetMarker>()), entity);
+/// }
 /// ```
 pub trait Marker: Component + DeserializeOwned + Serialize + Copy {
     /// Id of the marker
@@ -120,13 +135,13 @@ pub trait MarkerAllocator<M: Marker>: Resource {
 
     /// Create new unique marker `M` and attach it to entity.
     /// Or get old marker if this entity is already marked.
-    fn mark<'a>(&mut self, entity: Entity, storage: &mut WriteStorage<'a, M>) -> M {
+    fn mark<'a>(&mut self, entity: Entity, storage: &mut WriteStorage<'a, M>) -> (M, bool) {
         match storage.get(entity).cloned() {
-            Some(marker) => marker,
+            Some(marker) => (marker, false),
             None => {
                 let marker = self.allocate(entity, None);
                 storage.insert(entity, marker);
-                marker
+                (marker, true)
             }
         }
     }
--- a/src/saveload/marker.rs
+++ b/src/saveload/marker.rs
@@ -20,11 +20,11 @@ use serde::de::DeserializeOwned;
 ///
 /// ```rust,no_run
 /// extern crate specs;
-/// #[macro_use] extern crate serde_derive;
+/// #[macro_use] extern crate serde;
 /// use std::collections::HashMap;
 /// use std::ops::Range;
 /// use specs::{Component, Entity, DenseVecStorage};
-/// use specs::saveload::marker::{Marker, MarkerAllocator};
+/// use specs::saveload::{Marker, MarkerAllocator};
 ///
 /// // Marker for entities that should be synced over network
 /// #[derive(Clone, Copy, Serialize, Deserialize)]
--- a/src/saveload/marker.rs
+++ b/src/saveload/marker.rs
@@ -11,10 +11,10 @@ use serde::ser::Serialize;
 use serde::de::DeserializeOwned;
 
 
-/// This trait should be implemetened by component which is gonna be used as marker.
+/// This trait should be implemetened by a component which is going to be used as marker.
 /// This marker should be set to entity that should be serialized.
-/// If serialization strategy needs to set marker to some entity it should use
-/// new marker allocated for `Marker::Allocator`.
+/// If serialization strategy needs to set marker to some entity
+/// then it should use newly allocated marker from `Marker::Allocator`.
 ///
 /// ## Example
 ///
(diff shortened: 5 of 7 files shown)


src/saveload/de.rs
```diff
@@ -0,0 +1,179 @@
...
+
+    fn expecting(&self, formatter: &mut Formatter) -> fmt::Result {
+        write!(formatter, "Sequence of serialized entities")
+    }
+
+    fn visit_seq<A>(mut self, mut seq: A) -> Result<(), A::Error>
+    where
+        A: SeqAccess<'de>,
+    {
+        while let Some(()) = seq.next_element_seed(DeserializeEntity {
```
torkleyy: style: replace with `.is_some`.

src/saveload/de.rs
```diff
@@ -0,0 +1,179 @@
...
+        storages: storages,
+        markers: markers,
+        allocator: allocator,
+        pd: PhantomData,
+    })
+}
+
+
+/// `DeerializeSeed` implementation for `World`
+//#[derive(SystemData)]
```
torkleyy: Why not use derive here? And further, if derive doesn't work here, why did you keep the line?

src/saveload/de.rs
```diff
@@ -0,0 +1,179 @@
...
+pub struct WorldDeserialize<'a, M: Marker, E: Error, T: Components<M::Identifier, E>> {
```
torkleyy: You shouldn't create unnecessary bounds on the struct itself, e.g. `E: Error` should be removed here.

src/saveload/details.rs
```diff
@@ -0,0 +1,199 @@
...
+    fn load<F>(data: Self::Data, ids: F) -> Result<Self, Self::Error>
+    where
+        F: FnMut(M) -> Option<Entity>;
+}
+
+/// An error type which cannot be instantiated.
+/// Used as a placeholder for associated error types if
+/// something cannot fail.
+#[derive(Debug)]
+pub enum NoError {}
```
torkleyy: Could this go into the `errors` module?

src/saveload/details.rs
```diff
@@ -0,0 +1,199 @@
...
+pub trait Storages<'a> {
+    /// Storages for read
+    type ReadStorages: SystemData<'a> + 'a;
+    /// Storages for write
+    type WriteStorages: SystemData<'a> + 'a;
+}
+
+/// This trait is implemented by any tuple where all elements are
+/// `Component + Serialize + DeserializeOwned`
+pub trait Components<M, E: Error>: for<'a> Storages<'a> {
```
torkleyy: No need for the `Error` bound here.

src/saveload/details.rs
```diff
@@ -0,0 +1,199 @@
...
+                Ok(())
+            }
+        }
+
+        impl_components!(@ $($a|$b),*);
+    };
+
+    (@) => {};
+
+    (@ $ah:ident|$bh:ident $(,$a:ident|$b:ident)*) => {
```
torkleyy: Could you explain that?

src/saveload/marker.rs
```diff
@@ -0,0 +1,209 @@
...
+
+
+/// This trait should be implemetened by component which is gonna be used as marker.
+/// This marker should be set to entity that should be serialized.
+/// If serialization strategy needs to set marker to some entity it should use
+/// new marker allocated for `Marker::Allocator`.
+///
+/// # Example
+///
+/// ```rust,no_run
```
torkleyy: Why `no_run`?

src/saveload/marker.rs
```diff
@@ -0,0 +1,209 @@
...
+///     fn get(&self, id: u64) -> Option<Entity> {
+///         self.mapping.get(&id).cloned()
+///     }
+/// }
+///
+/// fn main() {}
+/// ```
+pub trait Marker: SerializableComponent + Copy {
+    /// Id of the marker
+    type Identifier: Copy + ::std::fmt::Debug + Eq + Hash;
```
torkleyy: Let's import the `Debug` trait.

src/saveload/marker.rs
```diff
@@ -0,0 +1,209 @@
...
+    /// Allocator for this `Marker`
+    type Allocator: MarkerAllocator<Self>;
+
+    /// Get this marker internal id
+    fn id(&self) -> Self::Identifier;
+
+    /// Update marker with new value.
+    /// It must preserve internal `Identifier`.
+    ///
+    /// # Panics
```
torkleyy: Should be `## Panics`

src/saveload/marker.rs
```diff
@@ -0,0 +1,209 @@
...
+    fn update(&mut self, update: Self) {
+        ::std::mem::drop(update);
+    }
+}
+
+/// This allocator is used with `Marker` trait
+/// It provides method for allocation of `Marker`s
+/// And also should provide `Marker -> Entity` mapping
+/// `maintain` method can be implemented for cleanup and actualization
+/// # Example
```
torkleyy: Create blank lines before and after this line and change it to `## Examples`. Oh, I just see there is no actual example. Did you forget that?

src/saveload/marker.rs
```diff
@@ -0,0 +1,209 @@
...
+    /// Get `Entity` by `Marker::Identifier`
+    fn get(&self, id: M::Identifier) -> Option<Entity>;
+
+    /// Create new unique marker `M` and attach it to entity.
+    /// Or get old marker if already marked.
+    fn mark<'a>(&mut self, entity: Entity, storage: &mut WriteStorage<'a, M>) -> M {
+        if let Some(marker) = storage.get(entity).cloned() {
+            marker
+        } else {
+            let marker = self.allocate(entity, None);
```
torkleyy: Should we just use `None` here?

src/saveload/marker.rs
```diff
@@ -0,0 +1,209 @@
...
+    }
+}
+
+/// This allocator is used with `Marker` trait
+/// It provides method for allocation of `Marker`s
+/// And also should provide `Marker -> Entity` mapping
+/// `maintain` method can be implemented for cleanup and actualization
+/// # Example
+/// see docs for `Marker`
+pub trait MarkerAllocator<M: Marker>: Resource {
```
torkleyy: Why not make the `Marker` an associated?

src/saveload/ser.rs
```diff
@@ -0,0 +1,160 @@
...
+    serseq.end()
+}
+
+
+
+/// Serialize components in tuple `T` of entities marked by `M`
+/// with serializer `S`
+/// Doesn't recursively mark referenced entities.
+/// Closure passed in `SerializableComponent::save` returns `None` for unmarked `Entity`
+/// In this case `SerializableComponent::save` may perform workaround
```
torkleyy: This should be a bit more detailed.

src/saveload/ser.rs
```diff
@@ -0,0 +1,160 @@
...
+    }
+    serseq.end()
+}
+
+/// This type implements `Serialize` so that it may be used in generic environment
+/// where `Serialize` implementer is expected
+/// It may be constructed manually (TODO: Add `new` function)
+/// Or fetched by `System` as `SystemData`
+/// Serializes components in tuple `T` with marker `M`
+pub struct WorldSerialize<'a, M: Marker, E: Error, T: Components<M::Identifier, E>> {
```
torkleyy: Again, too many bounds.

src/saveload/ser.rs
```diff
@@ -0,0 +1,160 @@
...
+            marker: *marker,
+            components: T::save(entity, storages, &ids).map_err(ser::Error::custom)?,
+        })?;
+    }
+    serseq.end()
+}
+
+/// This type implements `Serialize` so that it may be used in generic environment
+/// where `Serialize` implementer is expected
+/// It may be constructed manually (TODO: Add `new` function)
```
torkleyy: Please fix the TODO or make it a comment.

src/saveload/ser.rs
```diff
@@ -0,0 +1,160 @@
...
+    /// Use it if you want to delete entities that was just serialized
+    pub fn remove_serialized(&mut self) {
+        for (entity, _) in (&*self.entities, &self.markers.check()).join() {
+            self.entities.delete(entity);
+        }
+    }
+}
+
+
+impl<'a, M, E, T> SystemData<'a> for WorldSerialize<'a, M, E, T>
```
torkleyy: Same as before, could this be derived?

src/saveload/de.rs
omni-viral: I think it should start to work. It requires `PhantomData:  SystemData`. So I kept the line as a remainder

src/saveload/details.rs
omni-viral: Yeap

src/saveload/details.rs
omni-viral: But it is required everywhere is is used. Well. Maybe I can just keep it there and remove here.

src/saveload/details.rs
omni-viral: This creates recursion. See it is called few lines above. It cuts head of list and call macro again.

src/saveload/marker.rs
omni-viral: Nothing to run here. No need to waste cpu

src/saveload/marker.rs
omni-viral: The example fir `Marker` covers both traits

src/saveload/marker.rs
omni-viral: What if you want to use one allocator for many markers?

src/saveload/marker.rs
omni-viral: The `None` means allocator should come up with new id.

omni-viral: Probably it must be documented

src/saveload/ser.rs
omni-viral: Chukcha isn't a writer 😦

src/saveload/ser.rs
omni-viral: Yes. After `PhantomData` gets covered

src/saveload/marker.rs
omni-viral: Nothing to run here. It only needs to compile

src/saveload/marker.rs
omni-viral: One allocator can implement this trait for different M

src/saveload/marker.rs
omni-viral: This `None` asks to allocate new `Marker` with new unique `Identifier`.
If one has `Identifier` already but need new `Marker` he provides it in this function.

omni-viral: I'll add comment to `allocate` method about `id` argument.

src/saveload/ser.rs
omni-viral: Not sure how 🤷‍♂️

src/saveload/marker.rs
torkleyy: Please still run it.

src/saveload/details.rs
omni-viral: Sure

src/saveload/details.rs
omni-viral: This make macro call itself with one less ident recursively

src/saveload/details.rs
```diff
@@ -0,0 +1,179 @@
+
+
```
torkleyy: Extra lines?

src/saveload/marker.rs
```diff
@@ -0,0 +1,211 @@
...
+    fn id(&self) -> Self::Identifier;
+
+    /// Update marker with new value.
+    /// It must preserve internal `Identifier`.
+    ///
+    /// ## Panics
+    ///
+    /// Allowed to panic if `self.id() != update.id()`.
+    /// But usually implementer may ignore `update.id()` value
+    /// as deserialization algorithm ensures `id()`'s match.
```
torkleyy: should be `id()`s (without apostrophe)

src/saveload/marker.rs
```diff
@@ -0,0 +1,211 @@
...
+    fn update(&mut self, update: Self) {
+        ::std::mem::drop(update);
+    }
+}
+
+/// This allocator is used with `Marker` trait
```
torkleyy: Please use full stops here.

src/saveload/marker.rs
```diff
@@ -0,0 +1,211 @@
...
+/// It provides method for allocation of `Marker`s
+/// And also should provide `Marker -> Entity` mapping
+/// `maintain` method can be implemented for cleanup and actualization
+/// See docs for `Marker` for example.
+pub trait MarkerAllocator<M: Marker>: Resource {
+    /// Allocate new `Marker`.
+    /// Stores mapping `Marker` -> `Entity`.
+    /// If _id_ arguemtn is `Some(id)` then new marker will have this `id`.
```
torkleyy: "argumetn"

omni-viral: ¯\_(ツ)_/¯

src/error.rs
```diff
@@ -128,3 +128,22 @@ impl StdError for WrongGeneration {
          (e.g. because the entity has been deleted)"
     }
 }
+
+
+/// An error type which cannot be instantiated.
+/// Used as a placeholder for associated error types if
+/// something cannot fail.
+#[derive(Debug)]
+pub enum NoError {}
```
kvark: Should be `PhantomError`

src/error.rs
```diff
@@ -128,3 +128,22 @@ impl StdError for WrongGeneration {
...
+    fn fmt(&self, _: &mut Formatter) -> FmtResult {
+        match *self {}
+    }
+}
+
+impl StdError for NoError {
+    fn description(&self) -> &str {
+        match *self {}
+    }
+}
```
kvark: newline missing

src/lib.rs
```diff
@@ -216,6 +219,9 @@ pub use world::{Component, CreateIter, CreateIterAtomic, EntitiesRes, Entity, En
 #[cfg(feature = "common")]
 pub mod common;
 
+#[cfg(feature = "serialize")]
+pub mod saveload;
```
kvark: why is it called `saveload` and not `serialize`/`serial`?

src/saveload/de.rs
```diff
@@ -0,0 +1,146 @@
+
+use std::fmt::{self, Display, Formatter};
```
kvark: I'd prefer either just `use std::fmt` or no `self` in the list for consistency.

src/saveload/de.rs
```diff
@@ -0,0 +1,146 @@
...
+use std::marker::PhantomData;
+
+use serde::de::{self, Deserialize, DeserializeSeed, Deserializer, SeqAccess, Visitor};
```
kvark: nit, but same here

src/saveload/de.rs
```diff
@@ -0,0 +1,146 @@
...
+    where
+        D: Deserializer<'de>,
+    {
+        let DeserializeEntity {
+            entities,
+            storages,
+            markers,
+            allocator,
+            ..
+        } = self;
```
kvark: nit: why do you need to deconstruct `self` here?

src/saveload/de.rs
```diff
@@ -0,0 +1,146 @@
...
+        let data = EntityData::<M, E, T>::deserialize(deserializer)?;
+        let entity = allocator.marked(data.marker.id(), entities, markers);
+        markers
+            .get_mut(entity)
+            .ok_or("Allocator is broken")
```
kvark: this would hide the actual error?

src/saveload/de.rs
```diff
@@ -0,0 +1,146 @@
...
+            .map_err(de::Error::custom)?
+            .update(data.marker);
+        let ids = |marker: M::Identifier| Some(allocator.marked(marker, entities, markers));
+        T::load(entity, data.components, storages, ids).map_err(de::Error::custom)?;
```
kvark: oh my, sugar!
nit: could we please use less of it here, i.e. just a simple
```rust
match T::load(...) {...}
```

src/saveload/de.rs
```diff
@@ -0,0 +1,146 @@
...
+        Ok(())
+    }
+}
+
+
+/// Wrapper for `Entities` and tuple of `WriteStorage`s that implements `serde::Deserialize`
```
kvark: is the comment copied from `DeserializeEntity`? what's the difference?

src/saveload/details.rs
```diff
@@ -0,0 +1,176 @@
...
+impl<C> SerializableComponent for C
+where
+    C: Component + DeserializeOwned + Serialize,
+{
+}
+
+/// This trait should be implemented in order to allow component
+/// to be serializeble with `SerializeSystem`.
+/// It is automatically implemented for all `SerializableComponent`s
+pub trait SaveLoadComponent<M>: Component {
```
kvark: probably needs a rename: save/load is a game term, too specific for specs ECS (although, I'm not 100% confident about this)

src/saveload/details.rs
```diff
@@ -0,0 +1,176 @@
...
+
+#[derive(Serialize, Deserialize)]
+#[serde(bound = "")]
+pub struct EntityData<M: Marker, E, T: Components<M::Identifier, E>> {
+    pub marker: M,
+    pub components: T::Data,
+}
+
+/// Trivial component is anything that is Component + Copy + DeserializeOwned + Serialize
+pub trait SerializableComponent: Component + DeserializeOwned + Serialize {}
```
kvark: where do we actually need this? It looks like `SaveLoadComponent` could be implemented straight for `C: Component + DeserializeOwned + Serialize`

src/saveload/details.rs
```diff
@@ -0,0 +1,176 @@
...
+pub trait SaveLoadComponent<M>: Component {
+    /// Serializable data representation for component
+    type Data: Serialize + DeserializeOwned;
+
+    /// Error may occur duing serialization or deserialization of component
+    type Error: Error;
+
+    /// Convert this component into serializable form (`Data`) using
+    /// entity to marker mapping function
+    fn save<F>(&self, ids: F) -> Result<Self::Data, Self::Error>
```
kvark: hmm, on the other hand, `save`/`load` methods do sound good :)

src/saveload/details.rs
```diff
@@ -0,0 +1,176 @@
...
+    type WriteStorages: SystemData<'a> + 'a;
+}
+
+/// This trait is implemented by any tuple where all elements are
+/// `Component + Serialize + DeserializeOwned`
+pub trait Components<M, E>: for<'a> Storages<'a> {
+    /// Serializable and deserializable intermediate representation
+    type Data: Serialize + DeserializeOwned;
+
+    /// Saves `Component`s of entity into `Intermediate` serializable representation
```
kvark: where is this `Intermediate` defined?

src/saveload/ser.rs
```diff
@@ -0,0 +1,169 @@
...
+                    None => {
+                        let marker = allocator.mark(entity, markers);
+                        add.push((entity, marker));
+                        Some(marker.id())
+                    }
+                }
+            };
+            for (entity, marker) in to_serialize.into_iter() {
+                serseq.serialize_element(&EntityData::<M, E, T> {
+                    marker: marker,
```
kvark: nit: could just be `marker,`

src/saveload/ser.rs
```diff
@@ -0,0 +1,169 @@
...
+                match markers.get(entity).cloned() {
+                    Some(marker) => Some(marker.id()),
+                    None => {
+                        let marker = allocator.mark(entity, markers);
+                        add.push((entity, marker));
+                        Some(marker.id())
+                    }
+                }
+            };
+            for (entity, marker) in to_serialize.into_iter() {
```
kvark: I think you can omit `into_iter()`, this is done by `for` sugar

src/saveload/ser.rs
```diff
@@ -0,0 +1,169 @@
...
+        if to_serialize.is_empty() {
+            break;
+        }
+        let mut add = vec![];
+        {
+            let mut ids = |entity| -> Option<M::Identifier> {
+                match markers.get(entity).cloned() {
+                    Some(marker) => Some(marker.id()),
+                    None => {
+                        let marker = allocator.mark(entity, markers);
```
kvark: this is a bit convoluted for the fact `allocator.mark` modifies the `markers` parameter
could this be rewritten so that we use a simple `Entry` API here, e.g. `match markers.entry(entity) {...}`?

src/saveload/marker.rs
```diff
@@ -0,0 +1,211 @@
...
+    /// Otherwise allocator creates marker with new unique id.
+    fn allocate(&mut self, entity: Entity, id: Option<M::Identifier>) -> M;
+
+    /// Get `Entity` by `Marker::Identifier`
+    fn get(&self, id: M::Identifier) -> Option<Entity>;
+
+    /// Create new unique marker `M` and attach it to entity.
+    /// Or get old marker if already marked.
+    fn mark<'a>(&mut self, entity: Entity, storage: &mut WriteStorage<'a, M>) -> M {
+        if let Some(marker) = storage.get(entity).cloned() {
```
kvark: nit: just use a `match` here

src/saveload/marker.rs
```diff
@@ -0,0 +1,211 @@
...
+        } else {
+            let marker = self.allocate(entity, None);
+            storage.insert(entity, marker);
+            marker
+        }
+    }
+
+    /// Find `Entity` by `Marker` with same id and update `Marker` attached instance.
+    /// Or create new and mark it.
+    fn marked<'a>(
```
kvark: please rename to be (or start with) a verb

src/saveload/marker.rs
```diff
@@ -0,0 +1,211 @@
...
+    /// Update marker with new value.
+    /// It must preserve internal `Identifier`.
+    ///
+    /// ## Panics
+    ///
+    /// Allowed to panic if `self.id() != update.id()`.
+    /// But usually implementer may ignore `update.id()` value
+    /// as deserialization algorithm ensures `id()`s match.
+    fn update(&mut self, update: Self) {
+        ::std::mem::drop(update);
```
kvark: nit: let's not use global imports

src/saveload/ser.rs
```diff
@@ -0,0 +1,169 @@
...
+    T: Components<M::Identifier, E>,
+    S: Serializer,
+{
+    let mut serseq = serializer.serialize_seq(None)?;
+    let mut to_serialize = (&**entities, &*markers)
+        .join()
+        .map(|(e, m)| (e, *m))
+        .collect::<Vec<_>>();
+    loop {
+        if to_serialize.is_empty() {
```
kvark: any reason to not write it as `while !to_serialize.is_empty() {...}`

src/saveload/ser.rs
```diff
@@ -0,0 +1,169 @@
...
+            where X: Marker,
+                $(
+                    $a: super::SaveLoadComponent<X::Identifier>,
+                    Z: From<$a::Error>,
+                )*
+        {
+            /// Create serializable structure from storages
+            #[allow(non_snake_case)]
+            pub fn new(entities: Entities<'a>,
+                       markers: ReadStorage<'a, X>
```
kvark: nit: please use the new formatting

src/saveload/ser.rs
```diff
@@ -0,0 +1,169 @@
...
+                       $(,$a: ReadStorage<'a, $a>)*) -> Self
+            {
+                WorldSerialize {
+                    entities: entities,
+                    storages: ($($a,)*),
+                    markers: markers,
```
kvark: nit: could be just `markers,` and `entities,`

src/saveload/ser.rs
```diff
@@ -0,0 +1,169 @@
...
+    E: Display,
+    T: Components<M::Identifier, E>,
+    S: Serializer,
+{
+    let mut serseq = serializer.serialize_seq(None)?;
+    let ids = |entity| -> Option<M::Identifier> { markers.get(entity).map(Marker::id) };
+    for (entity, marker) in (&**entities, &*markers).join() {
+        serseq.serialize_element(&EntityData::<M, E, T> {
+            marker: *marker,
+            components: T::save(entity, storages, &ids).map_err(ser::Error::custom)?,
```
kvark: this looks a bit convoluted, `ids` gets an entity's marker while we are already iterating entities with markers (so we already have that marker)

src/error.rs
omni-viral: Should be `!` but it isn't stable yet.
Why `PhantomError` anyway?

src/lib.rs
omni-viral: Because it is also do some tricky data conversion by functions `save` and `load`.
If you insist I'll rename.

src/saveload/de.rs
omni-viral: Why don't you like `self` here?

src/saveload/de.rs
omni-viral: Otherwise whole `self` gets borrowed by closure.

src/saveload/de.rs
omni-viral: This shouldn't happen unless implementation of `MarkerAllocator` is broken .

src/saveload/de.rs
omni-viral: Yeap. The comment isn't necessary here as this is implementation detail.

src/saveload/details.rs
omni-viral: I can rename it after someone make a better suggestion 😄

src/saveload/details.rs
omni-viral: Of course it could as this one is implemented for `C: Component + DeserializeOwned + Serialize`. It's just a handy bundle of traits to make bounds look nicer.

src/saveload/details.rs
omni-viral: The `Data` was named `Intermediate`

src/saveload/ser.rs
omni-viral: No reason now.

src/saveload/ser.rs
omni-viral: If component contains some `Entity` if must use this closure to get marker of that entity.
Then `SaveLoadComponent::Data` `Entity` gets replaced with marker`

src/saveload/details.rs
jojolepro: Serialize/Deserialize ?

src/saveload/de.rs
omni-viral:
```diff
- `Entity`
+ `Entities`
```

src/lib.rs
```diff
@@ -216,6 +219,9 @@ pub use world::{Component, CreateIter, CreateIterAtomic, EntitiesRes, Entity, En
 #[cfg(feature = "common")]
 pub mod common;
 
+#[cfg(feature = "serialize")]
```
torkleyy: This wasn't rebased correctly, it's called `serde` now.

src/saveload/marker.rs
```diff
@@ -0,0 +1,211 @@
+//! Provides with `Marker` and `MarkerAllocator` traits
```
torkleyy: Provides ~~with~~ `Marker`

src/saveload/mod.rs
```diff
@@ -0,0 +1,13 @@
+//! Save and load entites from various formats with serde
```
torkleyy: I would really appreciate if you could add a top-level overview here.

src/saveload/mod.rs
```diff
@@ -0,0 +1,13 @@
...
+
+mod de;
+mod ser;
+mod details;
+
+use self::details::{Components, EntityData, SerializableComponent, Storages};
+
+pub mod marker;
```
torkleyy: Could we not make the hierarchy any deeper and just reexport from `marker`?

src/saveload/ser.rs
```diff
@@ -0,0 +1,169 @@
+use std::fmt::Display;
+use std::marker::PhantomData;
+
+use serde::ser::{self, Serialize, SerializeSeq, Serializer};
+
+use {Entities, FetchMut, Join, ReadStorage, WriteStorage};
+
+use super::{Components, EntityData, Storages};
```
torkleyy: These are all local imports, please move them together and use `saveload::` instead of `super::`.

src/saveload/details.rs
```diff
@@ -0,0 +1,176 @@
...
+use serde::ser::Serialize;
+
+use {Component, Entity, ReadStorage, SystemData, WriteStorage};
+
+use super::marker::Marker;
+use error::NoError;
+
+#[derive(Serialize, Deserialize)]
+#[serde(bound = "")]
+pub struct EntityData<M: Marker, E, T: Components<M::Identifier, E>> {
```
torkleyy: No need for the `Marker` bound

src/saveload/ser.rs
omni-viral: `Entry` API is not there yet. PR #274 
It is possible to call `allocator.allocate(entity, None)` here and insert new marker into storage.
But I'd like it to be this way so the allocator is responsible for the marker attached to entity was allocated for this entity.

omni-viral: I can rewrite it as 
```rust
let mut ids = |entity| -> Option<M::Identifier> {
    let (marker, new) = allocator.mark(entity, markers);
    if new { add.push(entity); }
    Some(marker.id())
}
```

modifying `MarkerAllocator::mark` to report if new marker was created or not.

omni-viral: `MarkerAllocator::mark` then can be rewritten to use `Entry` API

src/saveload/details.rs
omni-viral: Either bound or adding another type parameter for use instead of `M::Identifier`.
And bound them equal in all impls

src/saveload/de.rs
kvark: I just find it inconsistent that some of the types in the same namespace are referred directly while others - via `fmt::Type`. I believe it reads better through the code if all the imports from the same namespace are consistent. In case of `fmt`, I find `fmt::Display` and `fmt::Formatter` quite nice, but that's a matter of taste.

src/error.rs
kvark: Well, first, because you are implementing `Error` for `NoError`, which already looks like a hack. Secondly, because the "phantom" is rather established for such things that can't be instantiated, so `PhantomError` is crystal clear here.

kvark:
@omni-viral thanks for addressing my notes!

Re:naming. How do you feel about "loadable" component/feature/trait? In my understanding, this is the make thing you are trying to achieve. Serialization/deserialization are just means to be able to load components from disk. Even saving is rather useless if you can't load it.

Also note:
> saveload feature

This is not a cargo feature, so we could have a less ambiguous PR name.

src/saveload/de.rs
```diff
@@ -0,0 +1,147 @@
...
+use saveload::marker::{Marker, MarkerAllocator};
+
+
+/// Wrapper for `Entity` and tuple of `WriteStorage`s that implements `serde::Deserialize`
+struct DeserializeEntity<'a, 'b: 'a, M: Marker, E, T: Components<M::Identifier, E>> {
+    entities: &'a Entities<'b>,
+    storages: &'a mut <T as Storages<'b>>::WriteStorages,
+    markers: &'a mut WriteStorage<'b, M>,
+    allocator: &'a mut FetchMut<'b, M::Allocator>,
+    pd: PhantomData<(M, E, T)>,
```
WaDelma: `M` shouldn't be needed in the `PhantomData` as it's already used in `&'a mut WriteStorage<'b, M>`.

src/saveload/de.rs
```diff
@@ -0,0 +1,147 @@
...
+    }
+}
+
+/// Wrapper for `Entities` and tuple of `WriteStorage`s that implements `serde::de::Visitor`
+struct VisitEntities<'a, 'b: 'a, M: Marker, E, T: Components<M::Identifier, E>> {
+    entities: &'a Entities<'b>,
+    storages: &'a mut <T as Storages<'b>>::WriteStorages,
+    markers: &'a mut WriteStorage<'b, M>,
+    allocator: &'a mut FetchMut<'b, M::Allocator>,
+    pd: PhantomData<(M, E, T)>,
```
WaDelma: Same here.

src/saveload/details.rs
WaDelma: Maybe add comment for explaining this.

src/saveload/marker.rs
```diff
@@ -0,0 +1,213 @@
...
+use std::hash::Hash;
+
+use shred::Resource;
+use {Component, DenseVecStorage, Entities, Entity, Join, ReadStorage, WriteStorage};
+
+use serde::ser::Serialize;
+use serde::de::DeserializeOwned;
+
+
+/// This trait should be implemetened by component which is gonna be used as marker.
```
WaDelma: s/gonna/going to be/

WaDelma: s/implemetened by/implemented by a/

src/saveload/marker.rs
```diff
@@ -0,0 +1,213 @@
...
+/// This marker should be set to entity that should be serialized.
+/// If serialization strategy needs to set marker to some entity it should use
+/// new marker allocated for `Marker::Allocator`.
```
WaDelma: s/it should use new marker allocated for/then it should use newly allocated marker from/

src/saveload/marker.rs
```diff
@@ -0,0 +1,213 @@
...
+
+use shred::Resource;
+use {Component, DenseVecStorage, Entities, Entity, Join, ReadStorage, WriteStorage};
+
+use serde::ser::Serialize;
+use serde::de::DeserializeOwned;
+
+
+/// This trait should be implemetened by component which is gonna be used as marker.
+/// This marker should be set to entity that should be serialized.
```
WaDelma: This sentense is unclear to me. Does it mean that implementing the trait marks it to be serialized?

src/saveload/marker.rs
```diff
@@ -0,0 +1,213 @@
...
+    fn id(&self) -> Self::Identifier;
+
+    /// Update marker with new value.
+    /// It must preserve internal `Identifier`.
+    ///
+    /// ## Panics
+    ///
+    /// Allowed to panic if `self.id() != update.id()`.
+    /// But usually implementer may ignore `update.id()` value
+    /// as deserialization algorithm ensures `id()`s match.
```
WaDelma: s/ensures/ensures that/
also I prefer not having parenthesis like this, but I am not sure if we have policy for it.

src/saveload/marker.rs
```diff
@@ -0,0 +1,213 @@
...
+    type Identifier: Copy + Debug + Eq + Hash;
+
+    /// Allocator for this `Marker`
+    type Allocator: MarkerAllocator<Self>;
+
+    /// Get this marker internal id
+    fn id(&self) -> Self::Identifier;
+
+    /// Update marker with new value.
+    /// It must preserve internal `Identifier`.
```
WaDelma: s/internal/the internal/

src/saveload/marker.rs
```diff
@@ -0,0 +1,213 @@
...
+    ///
+    /// ## Panics
+    ///
+    /// Allowed to panic if `self.id() != update.id()`.
+    /// But usually implementer may ignore `update.id()` value
```
WaDelma: s/\`update.id()\` value/the value of the \`update.id()\`/

src/saveload/marker.rs
```diff
@@ -0,0 +1,213 @@
...
+    /// as deserialization algorithm ensures `id()`s match.
+    fn update(&mut self, update: Self) {
+        ::std::mem::drop(update);
+    }
+}
+
+/// This allocator is used with `Marker` trait.
```
WaDelma: s/with/with the/

src/saveload/marker.rs
```diff
@@ -0,0 +1,213 @@
...
+/// It provides method for allocation of `Marker`s.
```
WaDelma: s/allocation of/allocating new/

src/saveload/marker.rs
```diff
@@ -0,0 +1,213 @@
...
+/// And also should provide `Marker -> Entity` mapping
```
WaDelma: Maybe "It should also provide a \`Marker -> Entity\` mapping."

src/saveload/marker.rs
```diff
@@ -0,0 +1,213 @@
...
+/// `maintain` method can be implemented for cleanup and actualization.
```
WaDelma: s/\`maintain\`/The \`maintain\`/

src/saveload/marker.rs
```diff
@@ -0,0 +1,213 @@
...
+/// See docs for `Marker` for example.
```
WaDelma: s/example/an example/

src/saveload/marker.rs
```diff
@@ -0,0 +1,213 @@
...
+pub trait MarkerAllocator<M: Marker>: Resource {
+    /// Allocate new `Marker`.
+    /// Stores mapping `Marker` -> `Entity`.
+    /// If _id_ argument is `Some(id)` then new marker will have this `id`.
```
WaDelma: s/new/the new/

src/saveload/marker.rs
```diff
@@ -0,0 +1,213 @@
...
+    /// Otherwise allocator creates marker with new unique id.
```
WaDelma: s/new/a new/

src/saveload/marker.rs
```diff
@@ -0,0 +1,213 @@
...
+    fn allocate(&mut self, entity: Entity, id: Option<M::Identifier>) -> M;
+
+    /// Get `Entity` by `Marker::Identifier`
+    fn get(&self, id: M::Identifier) -> Option<Entity>;
+
+    /// Create new unique marker `M` and attach it to entity.
+    /// Or get old marker if already marked.
```
WaDelma: s/if/if the entity is/

src/saveload/marker.rs
```diff
@@ -0,0 +1,213 @@
...
+            Some(marker) => marker,
+            None => {
+                let marker = self.allocate(entity, None);
+                storage.insert(entity, marker);
+                marker
+            }
+        }
+    }
+
+    /// Find `Entity` by `Marker` with same id and update `Marker` attached instance.
```
WaDelma: s/\`Entity\` by/an \`Entity\` by a/
s/\`Marker\` attached/the \`Marker\` attached to the/

src/saveload/ser.rs
```diff
@@ -0,0 +1,166 @@
...
+
+use serde::ser::{self, Serialize, SerializeSeq, Serializer};
+
+use {Entities, FetchMut, Join, ReadStorage, WriteStorage};
+
+use saveload::{Components, EntityData, Storages};
+use saveload::marker::{Marker, MarkerAllocator};
+
+/// Serialize components from specified storages via `SerializableComponent::save`
+/// of all marked entities with provided serializer.
```
WaDelma: I am not sure what this sentence is trying to say. The "of all marked entities" seems to be dangling. "Serialize all marked entities with provided serializer from specified storages" perhaps?

src/saveload/ser.rs
```diff
@@ -0,0 +1,166 @@
...
+/// When the component gets serialized with `SerializableComponent::save` method
+/// closure passed in `ids` argument mark unmarked `Entity` and they get serialized recursively.
```
WaDelma: s/closure/the closure/
s/argument mark/argmuent marks/
s/\`Entity\`/Entities/?
s/they get/they will get/

src/saveload/ser.rs
```diff
@@ -0,0 +1,166 @@
...
+            }
+        }
+        to_serialize = add;
+    }
+    serseq.end()
+}
+
+
+/// Serialize components from specified storages via `SerializableComponent::save`
+/// of all marked entities with provided serializer.
```
WaDelma: Same as before.

src/saveload/ser.rs
```diff
@@ -0,0 +1,166 @@
...
+/// When the component gets serialized with `SerializableComponent::save` method
+/// closure passed in `ids` arguemnt returns `None` for unmarked `Entity`.
+/// In this case `SerializableComponent::save` may perform workaround (forget about `Entity`)
+/// or fail.
+/// So function doesn't recursively mark referenced entities.
```
WaDelma: s/So/So the/

src/saveload/ser.rs
```diff
@@ -0,0 +1,166 @@
...
+        serseq.serialize_element(&EntityData::<M, E, T> {
+            marker: *marker,
+            components: T::save(entity, storages, &ids).map_err(ser::Error::custom)?,
+        })?;
+    }
+    serseq.end()
+}
+
+/// This type implements `Serialize` so that it may be used in generic environment
+/// where `Serialize` implementer is expected.
```
WaDelma: s/implementer/implementation/

src/saveload/ser.rs
```diff
@@ -0,0 +1,166 @@
...
+/// It may be constructed manually with `WorldSerialize::new`.
+/// Or fetched by `System` as `SystemData`.
```
WaDelma: s/by/from/

src/saveload/ser.rs
```diff
@@ -0,0 +1,166 @@
...
+
+world_serialize_new_functions!(A,B,C,D,E,F,G,H,I,J,K,L,M,N,O);
+
+impl<'a, M, E, T> WorldSerialize<'a, M, E, T>
+where
+    M: Marker,
+    T: Components<M::Identifier, E>,
+{
+    /// Remove all marked entities
+    /// Use it if you want to delete entities that was just serialized
```
WaDelma: s/Use it/Use this/
s/was/were/

src/saveload/de.rs
omni-viral: Types went serious changes and now it is indeed unnecessary 😄

src/saveload/marker.rs
omni-viral: Implementing the trait makes the component a valid marker. Marked entities get serialized.
Which marker to use and which components to serialize is for user to decide.

src/saveload/ser.rs
omni-viral: components ... of all marked entities

src/saveload/ser.rs
omni-viral: I'm not sure here. It's not fetched from `System` as it isn't in `System`. `System` fetches `SystemData` from `World`.

src/error.rs
omni-viral: `PhantomData` is unit type and can be instantiated. And it is an example from `std`.
IDK any other examples.

src/error.rs
omni-viral: added

src/saveload/mod.rs
omni-viral: It would be better if someone with better writing skills do this

src/error.rs
torkleyy: I don't have a strong opinion here, @kvark do you insist on this request?

src/saveload/marker.rs
```diff
@@ -0,0 +1,214 @@
...
+
+
+/// This trait should be implemetened by a component which is going to be used as marker.
+/// This marker should be set to entity that should be serialized.
+/// If serialization strategy needs to set marker to some entity
+/// then it should use newly allocated marker from `Marker::Allocator`.
+///
+/// ## Example
+///
+/// ```rust,no_run
```
torkleyy: Again, why `no_run`? Please remove that.

src/saveload/ser.rs
torkleyy: Would you like to rewrite this to use the Entry API in this PR or in a follow-up?

src/saveload/ser.rs
torkleyy: This is not resolved, could you use the latest version of rustfmt-nightly and format this please?

src/error.rs
kvark: Nope. I think I might have confused a bit the semantics of phantom types. Pretty sure I saw them used in the context where one couldn't create instances (`enum MyPhantomType {}`), but I don't want to push hard on that here. It's fine to ship as is.

torkleyy:
@omni-viral Thank you for this big PR!

bors r+

PR #283, opened 2017-10-07
Improve saveload docs and add example

torkleyy:


torkleyy:
@omni-viral Thanks for the review!

bors r+

Issue #145, opened 2017-05-05
[meta] follow Rust API guidelines


kvark:
https://github.com/brson/rust-api-guidelines

torkleyy:
Already on it.

Aceeri:
Checklist:
- **Organization** *(crate is structured in an intelligible way)*
  - [ ] Crate root re-exports common functionality ([C-REEXPORT])
  - [ ] Modules provide a sensible API hierarchy ([C-HIERARCHY])
- **Naming** *(crate aligns with Rust naming conventions)*
  - [ ] Casing conforms to RFC 430 ([C-CASE])
  - [ ] Ad-hoc conversions follow `as_`, `to_`, `into_` conventions ([C-CONV])
  - [ ] Methods on collections that produce iterators follow `iter`, `iter_mut`, `into_iter` ([C-ITER])
  - [ ] Iterator type names match the methods that produce them ([C-ITER-TY])
  - [ ] Ownership suffixes use `_mut` and `_ref` ([C-OWN-SUFFIX])
  - [ ] Single-element containers implement appropriate getters ([C-GETTERS])
- **Interoperability** *(crate interacts nicely with other library functionality)*
  - [ ] Types eagerly implement common traits ([C-COMMON-TRAITS])
    - `Copy`, `Clone`, `Eq`, `PartialEq`, `Ord`, `PartialOrd`, `Hash`, `Debug`,
      `Display`, `Default`
  - [ ] Conversions use the standard traits `From`, `AsRef`, `AsMut` ([C-CONV-TRAITS])
  - [ ] Collections implement `FromIterator` and `Extend` ([C-COLLECT])
  - [ ] Data structures implement Serde's `Serialize`, `Deserialize` ([C-SERDE])
  - [ ] Crate has a `"serde"` cfg option that enables Serde ([C-SERDE-CFG])
  - [ ] Types are `Send` and `Sync` where possible ([C-SEND-SYNC])
  - [ ] Error types are `Send` and `Sync` ([C-SEND-SYNC-ERR])
  - [ ] Error types are meaningful, not `()` ([C-MEANINGFUL-ERR])
  - [ ] Binary number types provide `Hex`, `Octal`, `Binary` formatting ([C-NUM-FMT])
- **Macros** *(crate presents well-behaved macros)*
  - [ ] Input syntax is evocative of the output ([C-EVOCATIVE])
  - [ ] Macros compose well with attributes ([C-MACRO-ATTR])
[...]

PR #285, opened 2017-10-07
Add examples and explanations to several item docs

torkleyy:
Partially addresses #145

### DIFF
--- a/src/world.rs
+++ b/src/world.rs
@@ -298,6 +298,37 @@ impl Entity {
 
 /// The entity builder, allowing to
 /// build an entity together with its components.
+///
+/// ## Examples
+///
+/// ```
+/// use specs::{Component, DenseVecStorage, HashMapStorage, World};
+///
+/// struct Health(f32);
+///
+/// impl Component for Health {
+///     type Storage = HashMapStorage<Self>;
+/// }
+///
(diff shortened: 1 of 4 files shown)


src/world.rs
```diff
@@ -722,6 +774,37 @@ impl World {
...
+    ///
+    /// There are two built-in resources:
+    ///
+    /// * `LazyUpdate` and
+    /// * `EntitiesRes`
+    ///
+    /// Both of them should only be fetched immutably, which is why
+    /// the latter one has a type def for convenience: `Entities` which
+    /// is just `Fetch<EntitiesRes>`. Both resources are special and need
+    /// to execute code at the end of the frame, which is done in `World::maintain`.
```
WaDelma: The text is fine, but it reminds me the fact that calling `World::maintain` is not enforced. Is there any way to encode this in type system?

torkleyy: I wouldn't know how, unless we change our architecture :(

torkleyy:
@WaDelma Thanks for your review, I really appreciate that! Regarding the above comments, I have several ideas how to improve Specs, however most of them require specialization or would introduce massive boilerplate otherwise.

bors +

torkleyy:
Oops

bors r+

PR #288, opened 2017-10-12
Replace let delta with let mut delta

torkleyy:
Sorry, the code was wrong. If you also want to dispatch you'll need to use a scope like so

```rust
let mut world = World::new();
world.add_resource(DeltaTime(0.04));

{
    let mut delta = world.write_resource::<DeltaTime>();
    delta.0 = 0.02;
    // The borrow ends here
}

dispatcher.dispatch(&mut world.res); // so we can borrow from world here again
```


SamHH:
That works, thanks!

torkleyy:
Thanks for confirming!

bors r+

PR #290, opened 2017-10-20
Introduce nightly feature

Aceeri:
Adds a new feature flag `nightly` for unstable, but useful features.

Also adds first feature using the flag which allows clearer panic messages when not adding/registering resources/components.

Examples of new error message when enabled:
```
thread 'main' panicked at 'Component `Pos` was unregistered. Did you forget to register the component with `World::register::<Pos>()`', src\libcore\option.rs:839:4
note: Run with `RUST_BACKTRACE=1` for a backtrace.
error: process didn't exit successfully: `target\debug\examples\cluster_bomb.exe` (exit code: 101)
```
```
thread 'main' panicked at 'No resource `specs::saveload::U64MarkerAllocator` exists in the world. Did you forget to add the resource with `World::add_resource::<specs::saveload::U64MarkerAllocator>(...)`', src\libcore\option.rs:839:4
note: Run with `RUST_BACKTRACE=1` for a backtrace.
error: process didn't exit successfully: `target\debug\examples\saveload.exe` (exit code: 101)
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,9 +36,10 @@ rudy = { version = "0.1", optional = true }
 
 [features]
 common = ["futures"]
+nightly = []
 
 [package.metadata.docs.rs]
-features = ["common", "serde"]
+features = ["common", "serde", "nightly"]
 
 [dev-dependencies]
 cgmath =  { version = "0.14", features = ["eders"] }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -39,7 +39,7 @@ common = ["futures"]
 nightly = []
 
 [package.metadata.docs.rs]
-features = ["common", "serde", "nightly"]
+features = ["common", "serde"]
 
 [dev-dependencies]
 cgmath =  { version = "0.14", features = ["eders"] }
--- a/src/world/mod.rs
+++ b/src/world/mod.rs
@@ -17,13 +17,15 @@ mod lazy;
 mod tests;
 
 #[cfg(not(feature = "nightly"))]
-/// Note: Enable nightly feature to get the component type
-/// printed instead of just a vague catch all message.
 const COMPONENT_NOT_REGISTERED: &str = "No component with the given id. Did you forget to register \
-                                        the component with `World::register::<ComponentName>()`?";
+                                        the component with `World::register::<ComponentName>()`? \
+                                        Note: Enable `nightly` feature to get the exact component \
+                                        type printed out.";
 #[cfg(not(feature = "nightly"))]
 const RESOURCE_NOT_ADDED: &str = "No resource with the given id. Did you forget to add \
-                                  the resource with `World::add_resource(resource)`?";
+                                  the resource with `World::add_resource(resource)`?
+                                  Note: Enable `nightly` feature to get the exact resource \
+                                  type printed out.";
 
 #[cfg(feature = "nightly")]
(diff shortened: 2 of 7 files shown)


src/world/mod.rs
```diff
@@ -16,11 +16,49 @@ mod lazy;
...
 const COMPONENT_NOT_REGISTERED: &str = "No component with the given id. Did you forget to register \
                                         the component with `World::register::<ComponentName>()`?";
+#[cfg(not(feature = "nightly"))]
 const RESOURCE_NOT_ADDED: &str = "No resource with the given id. Did you forget to add \
                                   the resource with `World::add_resource(resource)`?";
 
+#[cfg(feature = "nightly")]
+fn nightly_component_message<T>(id: usize) -> String {
+    let type_name = unsafe { ::std::intrinsics::type_name::<T>() };
+    let (message, function) = if id == 0 {
```
torkleyy: Hm, this formatting looks weird

src/world/mod.rs
```diff
@@ -16,11 +16,49 @@ mod lazy;
 #[cfg(test)]
 mod tests;
 
+#[cfg(not(feature = "nightly"))]
```
torkleyy: Could we add something like `"Note: enable nightly feature to get the component type printed out"`?

src/world/mod.rs
Aceeri: Not sure how I could make it better without making the lines super long though. `cargo +nightly fmt` doesn't seem to do anything to it either.

Aceeri:
@torkleyy This should be ready I believe

Cargo.toml
```diff
@@ -36,9 +36,10 @@ rudy = { version = "0.1", optional = true }
 
 [features]
 common = ["futures"]
+nightly = []
 
 [package.metadata.docs.rs]
-features = ["common", "serde"]
+features = ["common", "serde", "nightly"]
```
torkleyy: Does docs.rs build with the nightly toolchain?

src/world/mod.rs
```diff
@@ -16,11 +16,67 @@ mod lazy;
...
+/// Note: Enable nightly feature to get the component type
+/// printed instead of just a vague catch all message.
```
torkleyy: I'm not sure how this doc comment helps a user getting a panic? I thought we could include this note in the string.

src/world/mod.rs
```diff
@@ -311,7 +369,13 @@ impl World {
         let entities = self.entities();
         let resource = self.res.try_fetch::<MaskedStorage<T>>(id);
 
-        Storage::new(entities, resource.expect(COMPONENT_NOT_REGISTERED))
+
+        #[cfg(not(feature = "nightly"))]
+        let expect_message = COMPONENT_NOT_REGISTERED;
+        #[cfg(feature = "nightly")]
+        let expect_message = &nightly_component_message::<T>(id);
```
WaDelma: I would move these messages to function and guard that with `cfg`:
````rust
#[cfg(not(feature = "nightly"))]
fn component_message<T>(id: usize) -> &'static str {
    COMPONENT_NOT_REGISTERED
}
#[cfg(feature = "nightly")]
fn component_message<T>(id: usize) -> String {
    // Create message based on type
}
````

torkleyy:
Thank you!

bors r+

Issue #168, opened 2017-06-07
Component Wrappers


Aceeri:
Storages like `FlaggedStorage` and `ChangedStorage` (I'm assuming most "wrapper" storages as well) could be improved if it was possible to return wrapper types instead of just components themselves:
```rust
pub trait UnprotectedStorage<T> {
    type Wrapper<'a>;
    type WrapperMut<'a>;
    ...
    fn get<'a>(&'a self) -> Self::Wrapper<'a>;
    fn get_mut<'a>(&'a mut self) -> Self::WrapperMut<'a>;
    ...
}

impl<T> UnprotectedStorage<T> for ChangedStorage {
    type Wrapper<'a> = &'a T;
    type WrapperMut<'a> = Tracked<&'a mut T>;
    ...
    fn get<'a>(&'a self) -> &'a T { ... }
    fn get_mut<'a>(&'a mut self) -> Tracked<&'a mut T> { ... }
    ...
}
```

Alternatively:
```rust
pub trait UnprotectedStorage<'a, T> {
    type Wrapper: 'a;
    type WrapperMut: 'a;
    ...
    fn get(&self) -> Self::Wrapper;
    fn get_mut(&mut self) -> Self::WrapperMut;
    ...
}

impl<'a, T> UnprotectedStorage<'a, T> for ChangedStorage {
    type Wrapper = &'a T;
    type WrapperMut = Tracked<'a, &'a mut T>;
    ...
    fn get<'a>(&'a self) -> &'a T { ... }
    fn get_mut<'a>(&'a mut self) -> Tracked<'a, &'a mut T> { ... }
    ...
}
```

Since it would eliminate the need for things like `maintain` methods and ease overall usage of the storages to be used similar to any other.

Sadly, I don't think it is possible in current Rust's type system until either we get lifetime ATCs or a non-'static `TypeId` (since this is what prevents just adding a lifetime to the `UnprotectedStorage<T>`.

omni-viral:
It is possible with little boilerplate.
```rust
trait Wrapper<'a> {
  type Mut;
  type Ref;
}

trait UnprotectedStorage<T> {
  type Wrapper: for<'a> Wrapper<'a>;
  fn get<'a>(&'a self) -> <Self::Wrapper as Wrapper<'a>>::Ref;
  fn get_mut<'a>(&'a mut self) -> <Self::Wrapper as Wrapper<'a>>::Mut;
}
``|

torkleyy:
[...]

PR #291, opened 2017-10-21
Add `TrackedStorage`

torkleyy:
Phew, finally got an initial implementation.

## Still TODO:

* ~~Docs~~
* ~~Examples~~
* ~~Tests~~

## Why this is useful

* Systems can maintain their local, sorted cache. By using the change event system, the cache can be updated easily.
* It's much easier for the user to use components with `TrackedStorage` than `FlaggedStorage`, since for the latter one you have to check if the component would change yourself.

## Unresolved

* Technically, having partial equality is not 100% sufficient to detect changes. This means that a component containing float with the value `NaN` would always be marked as `Modified` even though it's not the case

* If a component is removed from entity `a` and also added to it again, this implementation would just mark it as modified (ignoring the case where the inserted component is the exact same as the previous). The problem is that changing this would be non-trivial. (we'd have to store removed components in a separate storage, add another bit set, ..)


src/storage/tracked.rs
```diff
@@ -0,0 +1,327 @@
...
+    Inserted,
+    Modified,
+    Removed,
+}
+
+impl Change {
+    fn add(&mut self, change: Change) {
+        let new = match (*self, change) {
+            (Change::None, new) => new,
+            (Change::Inserted, _) => Change::Inserted,
```
omni-viral: What if it `Inserted` and them `Removed`?

src/storage/tracked.rs
```diff
@@ -0,0 +1,316 @@
...
+        let TrackedStorage {
+            ref old,
+            ref storage,
+            ref mut changes,
+            ..
+        } = *self;
+
+        set.iter()
+            .filter(|id| old.get(*id) != storage.get(*id))
+            .for_each(|id| Self::insert_change(changes, id, Change::Modified))
```
Xaeroxe: This works, but also might get a bit expensive for large components.  We could probably make this more efficient by using something like this: [Dirty](https://bfrog.github.io/dirty/dirty/struct.Dirty.html) then implementing Component for it where T: Component, and using that as the internal component for TrackedStorage.  

From here we have two choices:

- Prioritize TrackedStorage efficiency, by assuming if the component is dirty then it was modified
- Prioritize TrackedStorage accuracy, by still checking the component against an old version but only when the component is dirty.

torkleyy: Nice idea, but if you look at `get_mut`, you can see that we need to return a mutable reference to the component there, which means that once we iterate over components mutably we have a set dirty flag.

torkleyy: And also this isn't expensive ^

omni-viral: And for small components it will introduce unnecessary overhead.
I think user should have an option to choose if he wants accuracy or efficiency and if he wants to use `Dirty` wrapper.

torkleyy: I only check equality for indices of `set` which only contains changed component indices.

torkleyy: Which means this optimization is already implemented, just a little bit different

torkleyy: Oh I meant to use `changes` here, not the `set` parameter. Sorry!

src/storage/tracked.rs
torkleyy: Sorry not pushed that yet. It would be `None`

src/storage/tracked.rs
torkleyy: I should probably go sleep because what I've been commenting here was nonsense.

torkleyy:
Pushed now, will finish this PR tomorrow

src/storage/tracked.rs
Xaeroxe: If my understanding is correct, then `changes` should contain an index for all components in question right?  Which means we're still doing this for all of them.  One thing I failed to mention though was that my idea with Dirty also included returning Dirty inside the Storage implementation, and using Dirty<T> as your component type.  So the end user who is iterating over the components will have to explicitly call `.write()` and set the dirty flag when they want to modify a component.

Xaeroxe: Except my idea falls apart as soon as I remember that query specs by component type, not by storage type, so this wouldn't work.  My bad.

torkleyy: I thought about that, but this is impossible without breakage: #168

src/storage/tracked.rs
```diff
@@ -0,0 +1,378 @@
...
+
+impl Change {
+    fn add(&mut self, change: Change) {
+        let new = match (*self, change) {
+            (Change::None, new) => new,
+            (Change::Inserted, Change::Modified) => Change::Inserted,
+            (Change::Inserted, Change::Removed) => Change::None,
+            (Change::Modified, Change::Modified) => Change::Modified,
+            (Change::Modified, Change::Removed) => Change::Removed,
+            (Change::Removed, Change::Inserted) => Change::Modified,
```
WaDelma: I would `use` the variants of the enum for this method to make it cleaner, but using them would be confusing for the `None` variant.
````rust
let new = match (*self, change) {
    (None, new) => new,
    (Inserted, Modified) => Inserted,
    (Inserted, Removed) => None,
    (Modified, Modified) => Modified,
    (Modified, Removed) => Removed,
    (Removed, Inserted) => Modified,
    (old, new) => panic!("Didn't expect change from {:?} to {:?}", old, new),
};
````
Also could this use `Add` trait?

File: src/storage/tracked.rs [hunk 4]
```diff
@@ -0,0 +1,378 @@
...
+    pub fn change_events<'a>(&'a self) -> ChangeEvents<'a> {
+        let inner = self.changes.iter().enumerate();
+
+        ChangeEvents { inner }
+    }
+
+    unsafe fn reset(&mut self) {
+        for id in &self.changed {
+            let elem = self.old.get_mut(id);
+            *elem = self.storage.get(id).clone();
```
WaDelma: For some reason this I had hard time understanding this code the first time I was reading it. Propably because there is `let` instead of `let mut` here. Dunno how to make it better though.

src/storage/tracked.rs
```diff
@@ -0,0 +1,378 @@
...
+        }
+
+        self.changed.clear();
+        self.changes.iter_mut().for_each(|c| *c = Change::None);
```
WaDelma: First time read `changed` and `changes` as the same. Maybe rename `changed` to `modified`.

WaDelma: Also do you actually need to reset elements of `changes` to `None`? I would think that access to it would be always guarded by `changed`.

WaDelma: Or scratch that. `insert_change` doesn't check it.

If you used `changed` as guard everywhere you didn't need the `None` at all, but having it might be clearer.

src/storage/tracked.rs
```diff
@@ -0,0 +1,378 @@
...
+use std::slice;
+
+use hibitset::{BitSet, BitSetLike};
+
+use {Component, DenseVecStorage, Index, MaskedStorage, Storage, UnprotectedStorage};
+
+#[derive(Copy, Clone, Debug, PartialEq, PartialOrd)]
+pub enum Change {
+    #[doc(hidden)]
+    None,
```
WaDelma: Why is the `None` hidden? Also did you think just wrapping `Change` to `Option` or does that make it too verbose?

src/storage/tracked.rs
```diff
@@ -0,0 +1,378 @@
...
+pub struct ChangeEvents<'a> {
+    inner: ChangeEventsInner<'a>,
+}
+
+impl<'a> Iterator for ChangeEvents<'a> {
+    type Item = (Index, Change);
+
+    fn next(&mut self) -> Option<Self::Item> {
+        self.inner
+            .find(|&(_, &c)| c != Change::None)
```
WaDelma: Why doesn't this use the bitset?

torkleyy:
@WaDelma `changed` should be removed, it doesn't make any sense.

src/storage/tracked.rs
torkleyy: Didn't document but basically that's because `None`s are filtered and just the internal layout. Using `Option` would need more space.

src/storage/tracked.rs
torkleyy: The bit set is broken, it has to be removed.

src/storage/tracked.rs
torkleyy: Me too, I think we're both used to the mutable binding because of ECS resources which require mutability for `DerefMut`.

src/storage/tracked.rs
torkleyy: > Also could this use Add trait?

This has a very different meaning than mathematical addition and I also don't want to expose it.

src/world/mod.rs
```diff
@@ -387,8 +387,10 @@ impl World {
         self.write_resource()
     }
 
-    /// Allows building an entity with its
-    /// components.
+    /// Allows building an entity with its components.
+    ///
+    /// **You have to ensure that the component storages you're
```
kvark: or what? I'd expect this to be something enforced by the type system

src/storage/tracked.rs
```diff
@@ -0,0 +1,413 @@
...
+/// The insertion has the bigger priority (`Inserted`).
+///
+/// ### A component has been modified and then removed
+///
+/// The removal has priority over the modification (`Removed`).
+#[derive(Copy, Clone, Debug, PartialEq, PartialOrd)]
+pub enum Change {
+    /// `None`s are filtered out and the user should never need to deal with it.
+    #[doc(hidden)]
+    None,
```
kvark: if it shouldn't affect the user, why is it exposed to the user?

src/storage/tracked.rs
```diff
@@ -0,0 +1,413 @@
...
+            (Removed, Inserted) => Modified,
+            (old, new) => panic!("Didn't expect change from {:?} to {:?}", old, new),
+        };
+
+        *self = new;
+    }
+}
+
+/// Events iterator for all events between the last reset and the last,
+/// eventual maintain. See `Change` docs for more explanations.
```
kvark: "last, eventual maintain" ?

src/storage/tracked.rs
```diff
@@ -0,0 +1,413 @@
...
+    storage: S,
+}
+
+impl<C, S> TrackedStorage<C, S>
+where
+    C: Clone,
+    S: UnprotectedStorage<C>,
+{
+    /// Returns an iterator for all change events.
+    pub fn change_events<'a>(&'a self) -> ChangeEvents<'a> {
```
kvark: `'a` can be totally omitted

src/storage/tracked.rs
```diff
@@ -0,0 +1,413 @@
...
+        let inner = self.changes.iter().enumerate();
+
+        ChangeEvents { inner }
+    }
+
+    /// Resets the `TrackedStorage`.
+    pub unsafe fn reset(&mut self, set: &BitSet) {
```
kvark: why is it unsafe?

src/storage/tracked.rs
torkleyy: True, kept that from a previous iteration

src/storage/tracked.rs
torkleyy: If the `set` is wrong, this would access uninitialized memory. This is an `UnprotectedStorage`, I'm glad there are safe methods at all :smi

src/world/mod.rs
torkleyy: Or it would panic, which is very unfortunate. The problem is that you can't borrow e.g. a resource otherwise. So we have the choice between taking `&mut self` and not allowing resource access (which is a common use case) or taking `&self` and risking that someone has borrowed that storage.

src/storage/tracked.rs
torkleyy: It's explained in `Change` docs. If the component doesn't implement `PartialEq`, you can also use the `TrackedStorage`, you just don't get any modification events but just insertions and removals.

src/storage/tracked.rs
torkleyy: Would that be worth it to create another enum?

torkleyy: I think we do indeed need to expose that, because otherwise the user doesn't have an id -> change mapping.

src/world/mod.rs
torkleyy: The core problem of all this is that the borrow checker cannot track our resource entries :( (which makes sense, but still)

src/storage/tracked.rs
torkleyy: If it's the grammar that's broken please suggest a better wording.

src/storage/tracked.rs
WaDelma: Can you elaborate bit why the bitset was broken? For me the only problem was that it wasn't being used in everywhere.

torkleyy: It would have contained every component, no matter if it has changed or not. Thus, it was just an additional overhead and wasted memory.

WaDelma: Ah, so you were adding to it even the `None` changes?

torkleyy: The problem is that `get_mut` often gets called without an actual change. So I could only mark all eventually changed components by marking all components retrieved by `get_mut`, which in the worst case is then equal to MaskedStorage::bitset`.

src/world/mod.rs
torkleyy: I'll solve this before I release a new version by renaming this method to `create_entity_unchecked` and just having a `create_entity` accept a mutable world.

torkleyy:
Thanks for all the reviews!

bors r+

PR #296, opened 2017-11-01
Add trait for easy registration with the World

Rhuagh:


WaDelma:
So this is basically simpler version of #155? Should we just resurrect that one or simplify it to something like this?

torkleyy:
The main motivation of @Aceeri's version was serialization and deserialization AFAIK (which has been solved by @omni-viral ), so I think a simpler solution is both sufficient and more practical.

src/common.rs
```diff
@@ -150,6 +150,11 @@ impl Errors {
     }
 }
 
+/// Trait used to bundle up resources/components for easy registration with the `World`.
+pub trait WorldResourceRegistrator {
```
torkleyy: I got a better name, today ;) You're already using "bundle" in the documentation, maybe that'd justify calling the trait `Bundle`?

src/common.rs
```diff
@@ -150,6 +150,11 @@ impl Errors {
...
+    fn register_with_world(self, world: &mut World);
```
torkleyy: Register implies a restriction to components, could you rename it to `add*`, please?

src/common.rs
```diff
@@ -337,4 +342,24 @@ mod test {
...
+
+        pub struct TestRegistrator;
+
+        impl WorldResourceRegistrator for TestRegistrator {
+            fn register_with_world(self, world: &mut World) {
+                world.add_resource(SomeResource { v : 12 });
+            }
+        }
+
+        TestRegistrator.register_with_world(&mut world);
```
torkleyy: We could add an impl for `World` here which adds an `add_bundle` method. Do you think that's useful @Rhuagh @WaDelma ?

src/world/mod.rs
```diff
@@ -529,3 +534,8 @@ impl Default for World {
         }
     }
 }
+
+/// Trait used to bundle up resources/components for easy registration with `World`.
+pub trait Bundle {
+    fn add_to_world(self, world: &mut World);
```
torkleyy: Missing docs

Rhuagh: Yeah, I didn't know what to write :/

src/world/tests.rs
```diff
@@ -101,3 +101,23 @@ fn delete_twice() {
...
+
+    impl Bundle for TestBundle {
+        fn add_to_world(self, world: &mut World) {
+            world.add_resource(SomeResource { v : 12 });
+        }
+    }
+
+    world.add_bundle(TestBundle);
+    assert_eq!(12, world.read_resource::<SomeResource>().v);
+}
```
Aceeri: New line

Rhuagh: It's missing .. fixed now.

Xaeroxe:
What's blocking merging here?  This looks ready to go.

torkleyy:
@kvark is on vacation and wants to approve it before merging, otherwise it's ready. Besides, we can't release anyways since there are still soundness issues with saveload stuff.

PR #299, opened 2017-11-14
Adds way of getting components of specific entity

WaDelma:
After randomly browsing through [rhusics](https://github.com/Rhuagh/rhusics) I noticed [piece of code](https://github.com/Rhuagh/rhusics/blob/259ff9e4ea61e98919472ea9ac77005177301c1d/src/ecs/collide/systems/spatial_collision.rs#L178) and thought about API that would make it cleaner/more efficient and this is the result.

The code in question after this PR:
```rust
let (left_shape, left_pose, left_next_pose) =
    (&shapes, &poses, &next_poses).join().get(left_entity).uwnrap();

let (right_shape, right_pose, right_next_pose) =
    (&shapes, &poses, &next_poses).join().get(right_entity).uwnrap();
```

Depends on slide-rs/hibitset#20

---
This change is [Reviewable](https://reviewable.io/reviews/slide-rs/specs/299)

### DIFF
--- a/src/join.rs
+++ b/src/join.rs
@@ -250,12 +250,15 @@ impl<J: Join> JoinIter<J> {
         }
     }
 
-    /// Allows getting joined values for specific entity.
+    /// Allows getting joined values for specific raw index.
     /// 
-    /// This method doesn't check if the entity is alive, so it should only be used if it actually is.
-    pub fn get_unchecked(&mut self, entity: Entity) -> Option<J::Type> {
-        if self.keys.contains(entity.id()) {
-            Some(unsafe { J::get(&mut self.values, entity.id()) })
+    /// The raw index for an `Entity` can be retrieved using `Entity::id` method.
+    ///
+    /// As this method operates on raw indices, there is no check to see if the entity is still alive,
+    /// so the caller should ensure it instead.
+    pub fn get_unchecked(&mut self, index: Index) -> Option<J::Type> {
+        if self.keys.contains(index) {
+            Some(unsafe { J::get(&mut self.values, index) })
         } else {
             None
         }
--- a/src/join.rs
+++ b/src/join.rs
@@ -194,8 +194,67 @@ impl<J: Join> JoinIter<J> {
 
 impl<J: Join> JoinIter<J> {
     /// Allows getting joined values for specific entity.
+    /// 
+    /// ## Example
+    ///
+    /// ```
+    /// # use specs::*;
+    /// # #[derive(Debug, PartialEq)]
+    /// # struct Pos; impl Component for Pos { type Storage = VecStorage<Self>; }
+    /// # #[derive(Debug, PartialEq)]
+    /// # struct Vel; impl Component for Vel { type Storage = VecStorage<Self>; }
+    /// let mut world = World::new();
+    ///
+    /// world.register::<Pos>();
+    /// world.register::<Vel>();
+    /// 
+    /// // This entity could be stashed anywhere (into `Component`, `Resource`, `System`s data, etc.) as it's just a number.
+    /// let entity = world
+    ///     .create_entity()
+    ///     .with(Pos)
+    ///     .with(Vel)
+    ///     .build();
+    /// 
+    /// // Later
+    /// {
+    ///     let mut pos = world.write::<Pos>();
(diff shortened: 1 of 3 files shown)


src/join.rs
```diff
@@ -120,6 +121,17 @@ impl<J: Join> JoinIter<J> {
     }
 }
 
+impl<J: Join> JoinIter<J> {
+    /// Allows getting joined values for specific entity.
+    pub fn get(&mut self, e: Entity) -> Option<J::Type> {
+        if self.keys.contains(e.id()) {
```
Aceeri: Thinking about it again, maybe we should have a note saying you need to check if the entity is alive prior to doing this?

torkleyy: Yes, that's something I missed but spotted in your code later @Aceeri :P Somehow I'm more picky on Gitter I guess ^^

@WaDelma I think it would be preferable to just use `Index` as parameter instead of `Entity`.

WaDelma: Why would `Index` be preferred? Only thing that uses it is `UnprotectedStorage` which is not something normal library user would touch and all user facing API uses `Entity`.

So the problem with dead entities is that if the index is reused it would get that one instead of returning `None`? I would change it to check the liveliness then.

torkleyy: > I would change it to check the liveliness then.

But that would require to pass `Entities`.

WaDelma: `Entities` or `World` depending where it's being called.

I just don't know if it's good idea to leave this for the user to check.

torkleyy: You can get the entities from the world.

If it's a good idea, not sure, but it's better than using an entity without upholding the imposed guarantee.

src/join.rs
```diff
@@ -123,9 +123,9 @@ impl<J: Join> JoinIter<J> {
 
 impl<J: Join> JoinIter<J> {
     /// Allows getting joined values for specific entity.
-    pub fn get(&mut self, e: Entity) -> Option<J::Type> {
-        if self.keys.contains(e.id()) {
-            Some(unsafe { J::get(&mut self.values, e.id()) })
+    pub fn get(&mut self, entity: Entity, entities: &Entities) -> Option<J::Type> {
+        if self.keys.contains(entity.id()) && entities.is_alive(entity) {
```
torkleyy: Could we at least also provide an `_unchecked` variant?

torkleyy:
Also, please resolve the `Cargo.toml` conflict.

WaDelma:
What I should change `Cargo.toml` to? The change to `hibitset` that this needs hasn't landed yet.

torkleyy:
Reviewed 1 of 3 files at r4.
Review status: 1 of 3 files reviewed at latest revision, 3 unresolved discussions.

---

*[src/join.rs, line 256 at r4](https://reviewable.io:443/reviews/slide-rs/specs/299#-L0EqjH2CjcGMJgAY6wo:-L0EqjH3Epx3_TmH-Oea:byxtxsm) ([raw file](https://github.com/slide-rs/specs/blob/7b92618ed8ee76eab0f0509786374ce06da1ffd7/src/join.rs#L256)):*
> ```Rust
>     /// 
>     /// This method doesn't check if the entity is alive, so it should only be used if it actually is.
>     pub fn get_unchecked(&mut self, entity: Entity) -> Option<J::Type> {
> ```

Logically this still gets the joined value for a given index, not an entity. Please change the parameter type and the documentation accordingly.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/slide-rs/specs/299)*

torkleyy:
:lgtm: Thank you, I didn't see that you added another commit. Sorry for the delay!


bors r+

---

Reviewed 2 of 3 files at r4.
Review status: 2 of 3 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/slide-rs/specs/299#-:-L0zA0c8g08ibgLqmN7C:b-zhzuro)*

torkleyy:
Reviewed 1 of 1 files at r5.
Review status: all files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/slide-rs/specs/299)*

Issue #232, opened 2017-07-26
Addon Storages?


Aceeri:
The "wrapper" storages we have now with `FlaggedStorage` and `ChangedStorage` aren't very flexible. Since I'm sure someone down the line will want to check for changes for some component that doesn't have the wrapper storage. (Since I've run into this plenty of times already when trying to implement UI stuff with specs to the point of where I just add the `FlaggedStorage` for all the UI components so I don't have to worry about it anymore.)

I was thinking we might be able to add some `Vec<Box<AddonTrait>>` and abstract away the "real" storage (`DenseVecStorage`, `HashMapStorage`, etc.) so they can't be modified except through some interface.

Aceeri:
Coming back to this I have something working and actually kind of nice: https://gist.github.com/Aceeri/39b23eb18bf05848ee44c21c5e34e14a

Rhuagh:
Should this be closed as the PR was rejected in favor or `Tracked`?.

PR #301, opened 2017-11-15
Metadata

Aceeri:
Moves the wrapper storages into their own separate "observing" structures. This is a breaking change due to no default associated types, but I think it might be better for ergonomics.

Checks won't pass right now, still need to update the docs and such. Somewhat of a proof of concept right now.

fixes https://github.com/slide-rs/specs/issues/232

---
This change is [Reviewable](https://reviewable.io/reviews/slide-rs/specs/301)


WaDelma:
Tracking issue for associated type defaults rust-lang/rust#29661

specs-derive/src/lib.rs
```diff
@@ -3,14 +3,16 @@
 //!
 //! [sp]: https://slide-rs.github.io/specs-website/
```
torkleyy: Didn't review this file yet

src/storage/flagged.rs
```diff
@@ -67,55 +70,40 @@ use world::EntityIndex;
...
 ///
 ///         // To iterate over the flagged/modified components:
-///         for flagged_comp in ((&comps).open().1).join() {
+///         for flagged_comp in (comps.meta().flagged).join() {
 ///             // ...
 ///         }
 ///
 ///         // Clears the tracked storage every frame with this system.
-///         (&mut comps).open().1.clear_flags();
+///         comps.find_mut::<Flagged>().clear_flags();
```
torkleyy: Inconsistent with the above style

src/storage/meta.rs
```diff
@@ -0,0 +1,39 @@
...
+
+use storage::{MaskedStorage, Storage, WrappedStorage, UnprotectedStorage};
+
+use Index;
+
+/// Observes interactions with the component's storage and stores state alongside it.
+///
+/// Useful for things like tracking modifications to components, storing sorted lists related
+/// to the storage, etc.
+pub trait Metadata<T>: Default {
```
torkleyy: When you have a `TrackedStorage`, doesn't that allow to do basically 99% of the meta stuff you need?

src/storage/meta.rs
```diff
@@ -0,0 +1,39 @@
...
+    fn insert(&mut self, _: Index, _: &T) { }
+    fn remove(&mut self, _: Index, _: &T) { }
+}
+
+/// Exposes "sub"-metadata by allowing metadata structs to forward the types
+/// via this trait.
+///
+/// Useful for if you are using a tuple metadata and don't want to index each
+/// sub-metadata using `metadata.0`, `metadata.1`, etc. as well as helping with
+/// modelling generic APIs where something needs a specific metadata structure from
```
torkleyy: Not sure about that, since the type does not really have a logical meaning.

torkleyy: Also, it's probably a bit inefficient

src/storage/mod.rs
```diff
@@ -148,6 +150,62 @@ impl<T: Component> Drop for MaskedStorage<T> {
     }
 }
 
+#[derive(Derivative)]
+#[derivative(Default(bound = ""))]
+/// Associates the component's storage with its metadata.
```
torkleyy: Please place docs before attribs

src/storage/mod.rs
```diff
@@ -148,6 +150,62 @@ impl<T: Component> Drop for MaskedStorage<T> {
...
+pub struct WrappedStorage<T: Component> {
```
torkleyy: I don't see why we need such a type.

src/storage/mod.rs
```diff
@@ -192,6 +250,16 @@ where
     pub fn check(&self) -> BitSet {
         self.data.mask.clone()
     }
+
+    pub fn meta(&self) -> &T::Metadata {
+        &self.data.wrapped.meta
+    }
+
+    pub fn find<M>(&self) -> &M
```
torkleyy: Any reason why this is immutable-only?

src/storage/tracked.rs
```diff
@@ -118,6 +118,7 @@ where
     /// impl Component for Comp {
     ///     // Will use `DenseVecStorage` by default.
     ///     type Storage = TrackedStorage<Self>;
+    ///     type Metadata = ();
```
torkleyy: So `TrackedStorage` doesn't work with `Metadata`?

src/storage/mod.rs
Aceeri: There is a mutable version `find_mut`

src/storage/meta.rs
Aceeri: I doubt it is that inefficient, all it does is forward to `self.field` which is very likely to be inlined.

src/storage/meta.rs
Aceeri: Sorted lists could be useful as well, as well as wanting to use something other than `TrackedStorage`.

torkleyy: Sorted lists can be built using `TrackedStorage`, that's one of my main motivations for adding it.

src/storage/meta.rs
torkleyy: Sorry, you're right, I misunderstood first. I think the identifier `find` confused me.

src/storage/tracked.rs
Aceeri: It could work, I just haven't ported it over like the FlaggedStorage yet.

src/storage/mod.rs
Aceeri: I think it is better for implementation, otherwise I'd have to repeat myself for every usage of `get`, `get_mut`, `insert`, `remove`, etc.

src/storage/flagged.rs
Aceeri: Fixed this

specs-derive/src/lib.rs
```diff
@@ -61,6 +63,90 @@ fn impl_component(ast: &MacroInput) -> Tokens {
...
+                    field.ident = Some(Ident::new(counter.to_string()));
+                    field
+                }).collect::<Vec<_>>()
+            }
+            VariantData::Unit => Vec::new(),
+        }
+    };
+
+    let count = field.iter().count();
+    let name_list = (0..count).map(|_| name).cloned().collect::<Vec<_>>();
```
kvark: would `vec![name; count]` work instead?

src/lib.rs
```diff
@@ -1,5 +1,6 @@
-#![deny(missing_docs)]
+//#![deny(missing_docs)]
```
kvark: any reason this is commented out?

src/lib.rs
```diff
@@ -216,10 +220,10 @@ pub use shred::{Dispatcher, DispatcherBuilder, Fetch, FetchId, FetchIdMut, Fetch
 pub use shred::AsyncDispatcher;
 
 pub use storage::{BTreeStorage, Change, ChangeEvents, DenseVecStorage, DistinctStorage, Entry,
-                  FlaggedStorage, HashMapStorage, InsertResult, MaskedStorage, NormalRestriction,
+                  Flagged, HashMapStorage, HasMeta, InsertResult, MaskedStorage, Metadata, NormalRestriction,
```
kvark: uh, that formatting doesn't work good for lots of uses

src/storage/meta.rs
```diff
@@ -0,0 +1,36 @@
+
+use std::marker::PhantomData;
+
+use Index;
+
+/// Observes interactions with the component's storage and stores state alongside it.
```
kvark: I think `Metadata` is too broad for this. Perhaps, something like `Hook` would do better? Or maybe `Observer`.

src/storage/meta.rs
```diff
@@ -0,0 +1,36 @@
...
+/// Exposes "sub"-metadata by allowing metadata structs to forward the types
+/// via this trait.
+///
+/// Useful for if you are using a tuple metadata and don't want to index each
+/// sub-metadata using `metadata.0`, `metadata.1`, etc. as well as helping with
+/// modelling generic APIs where something needs a specific metadata structure from
+/// a component.
+pub trait HasMeta<M> {
+    /// Immutably gets the sub-metadata.
+    fn find(&self) -> &M;
```
kvark: can we rename the methods to `meta` and `meta_mut`? Will need to be adjusted if `Metadata` is renamed too.

src/storage/meta.rs
```diff
@@ -0,0 +1,36 @@
...
+}
+
+/// Exposes "sub"-metadata by allowing metadata structs to forward the types
+/// via this trait.
+///
+/// Useful for if you are using a tuple metadata and don't want to index each
+/// sub-metadata using `metadata.0`, `metadata.1`, etc. as well as helping with
+/// modelling generic APIs where something needs a specific metadata structure from
+/// a component.
+pub trait HasMeta<M> {
```
kvark: Actually, wouldn't this effectively be `AsRef + AsMut`?

src/storage/mod.rs
```diff
@@ -401,6 +473,19 @@ where
             data: &mut self.data,
         }
     }
+
+    /*
+    pub fn meta_mut(&mut self) -> &mut T::Metadata {
```
kvark: any reason this is commented out?

src/storage/tracked.rs
kvark: I don't think we should proceed until `TrackedStorage` is ported

src/lib.rs
Aceeri: Just commented it out because I was removing/adding different structures often, I'll uncomment and document stuff once I figure out the final design for it.

src/storage/meta.rs
Aceeri: I suppose so yes, I was initially worried about it being confusing that you got the inner portions through `.as_ref()` but I guess that doesn't matter for a majority of usecases anyways.

src/storage/mod.rs
Aceeri: Oop, that was a mistake, I wanted to make sure I changed most of the documentation to use `find::<T>();`.

src/storage/meta.rs
Aceeri: Actually, we kind of need this, I'm changing it very slightly though so it isn't really identical to AsRef/AsMut anymore.

torkleyy:
I think it would be sufficient if there would be three "levels" for observing:

* flagged
* tracked
* changed (meaning the `tracked` would only check insertions, changed stuff moved out)

where each item includes the information provided by the former. If there's anything we can't do with this, please tell me.

Aceeri:
I'll close this, since I like the `Tracked` api and think it is going to work out the way we want.

PR #302, opened 2017-11-20
Clean up

Aceeri:
Various little portions of clean up and making the code a bit more consistent.

---
This change is [Reviewable](https://reviewable.io/reviews/slide-rs/specs/302)

### DIFF
--- a/src/storage/restrict.rs
+++ b/src/storage/restrict.rs
@@ -7,7 +7,6 @@ use hibitset::BitSet;
 
 use {Component, Entities, Entity, Index, Join, ParJoin, Storage, UnprotectedStorage};
 use storage::MaskedStorage;
-use world::EntityIndex;
 
 /// Specifies that the `RestrictedStorage` cannot run in parallel.
 pub enum NormalRestriction {}
@@ -91,7 +90,7 @@ where
     /// the storage has it or not.
     pub fn get_unchecked(&self, entry: &Entry<'rf, T>) -> &T {
         entry.assert_same_storage(self.data.borrow());
-        unsafe { self.data.borrow().get(entry.index()) }
+        unsafe { self.data.borrow().get(entry.id) }
     }
 }
 
@@ -105,7 +104,7 @@ where
     /// the storage has it or not.
     pub fn get_mut_unchecked(&mut self, entry: &Entry<'rf, T>) -> &mut T {
         entry.assert_same_storage(self.data.borrow());
-        unsafe { self.data.borrow_mut().get_mut(entry.index()) }
+        unsafe { self.data.borrow_mut().get_mut(entry.id) }
     }
 }
 
@@ -191,7 +190,7 @@ where
         let (mask, data) = self.data.open_mut();
         RestrictedStorage {
             bitset: mask,
-            data,
+            data: data,
             entities: &self.entities,
             phantom: PhantomData,
         }
@@ -206,7 +205,7 @@ where
         let (mask, data) = self.data.open_mut();
         RestrictedStorage {
             bitset: mask,
-            data,
+            data: data,
             entities: &self.entities,
             phantom: PhantomData,
         }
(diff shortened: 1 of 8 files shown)


src/storage/restrict.rs
```diff
@@ -191,7 +190,7 @@ where
         let (mask, data) = self.data.open_mut();
         RestrictedStorage {
             bitset: mask,
-            data,
+            data: data,
             entities: &self.entities,
```
WaDelma: I would actually rename mask to bitset and use that syntax for both.

torkleyy:
Reviewed 8 of 8 files at r1.
Review status: all files reviewed at latest revision, 5 unresolved discussions.

---

*[src/storage/restrict.rs, line 193 at r1](https://reviewable.io:443/reviews/slide-rs/specs/302#-L-QR_duLH7icXfAHtyK:-L-QR_duLH7icXfAHtyL:box5wcg) ([raw file](https://github.com/slide-rs/specs/blob/93bfc753bea7c8ec30cb15a254c354a298d13956/src/storage/restrict.rs#L193)):*
> ```Rust
>         RestrictedStorage {
>             bitset: mask,
>             data: data,
> ```

Why did you make this change?

---

*[src/storage/restrict.rs, line 208 at r1](https://reviewable.io:443/reviews/slide-rs/specs/302#-L-QRfFh6KpGitpczTwd:-L-QRfFh6KpGitpczTwe:buo9nmt) ([raw file](https://github.com/slide-rs/specs/blob/93bfc753bea7c8ec30cb15a254c354a298d13956/src/storage/restrict.rs#L208)):*
> ```Rust
>         RestrictedStorage {
>             bitset: mask,
>             data: data,
> ```

Same here

---

*[src/storage/restrict.rs, line 260 at r1](https://reviewable.io:443/reviews/slide-rs/specs/302#-L-QT1TkDpMdMawhNjPp:-L-QT1TkDpMdMawhNjPq:b-ieb9x6) ([raw file](https://github.com/slide-rs/specs/blob/1d0a8997741956e3d6445de09527fe7bce3408fc/src/storage/restrict.rs#L260)):*
> ```Rust
> }
> 
> impl<'rf, T> EntityIndex for Entry<'rf, T>
> ```

I don't see how `Borrow` is better here. I would prefer if one could just get the `Entity` from `Entry`, so we don't have to be generic over `EntityIndex`.

---

*[src/world/entity.rs, line 181 at r1](https://reviewable.io:443/reviews/slide-rs/specs/302#-L-QTA0rMKr2BcICXrSm:-L-QTA0rMKr2BcICXrSn:b-ybu8tx) ([raw file](https://github.com/slide-rs/specs/blob/1d0a8997741956e3d6445de09527fe7bce3408fc/src/world/entity.rs#L181)):*
> ```Rust
> ///
> /// e.g. Entry, Entity, etc.
> pub trait EntityIndex {
> ```

Same here, I don't see a reason for removing it.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/slide-rs/specs/302)*

Aceeri:
*[src/storage/restrict.rs, line 193 at r1](https://reviewable.io:443/reviews/slide-rs/specs/302#-L-QR_duLH7icXfAHtyK:-L-T3ewEWkCv_PnE9CaW:b-my9v4l) ([raw file](https://github.com/slide-rs/specs/blob/93bfc753bea7c8ec30cb15a254c354a298d13956/src/storage/restrict.rs#L193)):*
<details><summary><i>Previously, torkleyy (Thomas Schaller) wrote…</i></summary><blockquote>

Why did you make this change?
</blockquote></details>

Consistency with the other fields mostly

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/slide-rs/specs/302)*

torkleyy:
If I lint that code, I get the suggestion to revert your change.

Aceeri:
What is the lint saying?

torkleyy:
"Expression can be simplified: field initialization shorthand can be used"

Additionally, I don't think it's any more consistent than it was before.

Aceeri:
I'll just close this PR for now then and open a separate one for combining `Entry` and `RestrictedStorage`.

PR #303, opened 2017-11-30
Allow consuming of BitSets, adds AtomicBitset to Joinable bitsets.

Aceeri:
Also adds a bunch of the missing changes to the change log.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 ## [Unreleased]
 
+* Implementations of Join for owned BitSets, including AtomicBitset. ([#303])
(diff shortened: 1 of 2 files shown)


torkleyy:
What is this useful for?

Aceeri:
Just a bit of ergonomics so you could do something like 
```
for (entity, _) in (&*entities, storage.check()).join() {
    // ...
}
```
Without using a reference, it cleans up a lot of code where you want to combine bitsets with operators too.

Comparison with only references:
```
for (entity, _) in (&*entities, &(&comps.check() ^ &comps2.check())).join() { }
```
Without:
```
for (entity, _) in (&*entities, comps.check() ^ comps2.check()).join() { }
```

PR #313, opened 2017-12-09
Improve Join and Entities docs

### LINKED ISSUE #307: Improve docs for `Join`

torkleyy:
Fixes #307

---
This change is [Reviewable](https://reviewable.io/reviews/slide-rs/specs/313)

### DIFF
--- a/src/join.rs
+++ b/src/join.rs
@@ -124,7 +124,9 @@ bitset_and!{A, B, C, D, E, F, G, H, I, J, K, L, M, N, O, P}
 ///
 ///     let entities = world.read_resource::<EntitiesRes>();
 ///     // note: `EntitiesRes` is the fetched resource; we get back
-///     // `Fetch<EntitiesRes>` which has a typedef as `Entities`.
+///     // `Fetch<EntitiesRes>`.
+///     // `Fetch<EntitiesRes>` can also be referred to by `Entities` which
+///     // is a shorthand type definition to the former type.
 ///
 ///     let joined: Vec<_> = (&*entities, &pos, &vel).join().collect(); // note the `&*entities`
 ///     assert_eq!(joined, vec![(ent, &Pos, &Vel)]);
--- a/src/join.rs
+++ b/src/join.rs
@@ -66,6 +66,75 @@ bitset_and!{A, B, C, D, E, F, G, H, I, J, K, L, M, N, O, P}
 /// The purpose of the `Join` trait is to provide a way
 /// to access multiple storages at the same time with
 /// the merged bit set.
+///
+/// Joining component storages means that you'll only get values where
+/// for a given entity every storage has an associated component.
+///
+/// ## Example
+///
+/// ```
+/// # use specs::*;
+/// # #[derive(Debug, PartialEq)]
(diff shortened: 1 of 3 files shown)


src/join.rs
```diff
@@ -66,6 +66,75 @@ bitset_and!{A, B, C, D, E, F, G, H, I, J, K, L, M, N, O, P}
 /// The purpose of the `Join` trait is to provide a way
 /// to access multiple storages at the same time with
 /// the merged bit set.
+///
+/// Joining component storages means that you'll only get values where
+/// for a given entity every storage has an associated component.
```
WaDelma: Should we draw parallel to SQL? Or is that just confusing?

torkleyy: I don't find it very useful tbh

src/join.rs
```diff
@@ -66,6 +66,75 @@ bitset_and!{A, B, C, D, E, F, G, H, I, J, K, L, M, N, O, P}
...
+///     // Now there is one entity that has both a `Vel` and a `Pos`.
+///     let joined: Vec<_> = (&pos, &vel).join().collect();
+///     assert_eq!(joined, vec![(&Pos, &Vel)]);
+///
+///     // If we want to get the entity the components are associated to,
+///     // we need to join over `Entities`:
+///
+///     let entities = world.read_resource::<EntitiesRes>();
+///     // note: `EntitiesRes` is the fetched resource; we get back
+///     // `Fetch<EntitiesRes>` which has a typedef as `Entities`.
```
Aceeri: I feel like the `Entities` reference here is somewhat xonfusinf, maybe: "`Fetch<EntitiesRes>` can also be referred to by `Entities` which is a shorthand type definition to the former `Fetch` type."

src/world/mod.rs
```diff
@@ -229,17 +230,42 @@ impl<'a> EntityBuilder<'a> {
...
 ///     .with(Pos { x: 0.0, y: 1.0 })
 ///     .with(Vel { x: 0.0, y: 1.0 })
 ///     .build();
+///
+/// {
+///     // `World::read` returns a component storage.
+///     let pos_storage = world.read::<Pos>();
+///     let vel_storage = world.read::<Vel>();
+///
+///     // `World::get` allows to get a component from it:
```
Aceeri: `Storage::get`

src/lib.rs
```diff
@@ -251,7 +249,8 @@ pub use storage::{MergeError, PackedData};
 /// Please note that you should call `World::maintain`
 /// after creating / deleting entities with this resource.
 ///
-/// When `.join`ing on `Entities`, you will need to do it like this:
+/// When `.join`ing on `Entities`, you will need to do it like this
```
Aceeri: "When joining `Entities` you will need to first dereference `Entities`/`Fetch<EntitiesRes>` to get the underlying `EntitiesRes`,  then you will need to re-reference it since only the referenced `Entities` has an implementation of `Join`."

torkleyy:
Thanks, please review again.

torkleyy:
Thanks for reviewing!

bors r+

PR #320, opened 2017-12-25
Add `LazyBuilder`

### LINKED ISSUE #319: EntitiesRes should provide a (lazy) entity builder
For convenience and readability, I recommend Entities (EntityRes) provide an entity builder that looks to the outside world to be the same as the one used for World::create_entity(). Internally it would call the lazy updater to link the components.

This would allow a similar pattern to be used to create entities inside and outside Systems.

The user's code would look something like this:

```
entities.create_entity(&lazy_update)
        .with(mesh.clone())
        .with(background.clone())
        .with(component.clone())
        .with(local_transform)
        .with(Transform::default())
        .build()
```

torkleyy:
Fixes #319

---
This change is [Reviewable](https://reviewable.io/reviews/slide-rs/specs/320)

### DIFF
--- a/src/world/lazy.rs
+++ b/src/world/lazy.rs
@@ -1,6 +1,36 @@
 use crossbeam::sync::TreiberStack;
 
-use {Component, Entity, World};
+use {Component, EntitiesRes, Entity, World};
+
+/// Like `EntityBuilder`, but inserts the component
+/// lazily, meaning on `maintain`.
+/// If you need those components to exist immediately,
+/// you have to insert them into the storages yourself.
+pub struct LazyBuilder<'a> {
+    /// The entity that we're inserting components for.
+    pub entity: Entity,
+    /// The lazy update reference.
+    pub lazy: &'a LazyUpdate,
+}
+
+impl<'a> LazyBuilder<'a> {
+    /// Inserts a component using `LazyUpdate`.
+    pub fn with<C>(self, component: C) -> Self
+    where
+        C: Component + Send + Sync,
+    {
+        self.lazy.insert(self.entity, component);
+
+        self
+    }
+
+    /// Finishes the building and returns the built entity.
+    /// Please note that no component is associated to this
+    /// entity until you call `World::maintain`.
+    pub fn build(self) -> Entity {
+        self.entity
+    }
+}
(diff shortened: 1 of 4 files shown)


trsoluti:
:lgtm:

I would have added it under Entities (where there's already a create(), but under lazy_update is fine.

---

Reviewed 3 of 3 files at r1.
Review status: all files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/slide-rs/specs/320#-:-L1Bu2tsCuW3XKDjcN-j:bvdo1ub)*

torkleyy:
Thanks for the review. I'll add a doc comment there to make this more visible.

src/world/lazy.rs
```diff
@@ -1,6 +1,36 @@
...
+impl<'a> LazyBuilder<'a> {
+    /// Inserts a component using `LazyUpdate`.
+    pub fn with<C>(self, component: C) -> Self
+    where
+        C: Component + Send + Sync,
+    {
+        self.lazy.insert(self.entity, component);
+
+        self
+    }
```
Aceeri: Maybe add a `with_id` variant for the dynamic components?

src/world/lazy.rs
```diff
@@ -29,6 +59,38 @@ pub struct LazyUpdate {
...
+    /// ```
+    /// # use specs::*;
+    /// # let mut world = World::new();
+    /// struct Pos(f32, f32);
+    ///
+    /// impl Component for Pos {
+    ///     type Storage = VecStorage<Self>;
+    /// }
+    ///
+    /// # let lazy = world.read_resource::<LazyUpdate>();
```
Aceeri: I feel like this might be worth exposing in the docs, just for people to understand where LazyUpdate comes from if they haven't seen it elsewhere.

torkleyy: I usually hide the fetching since you'll mostly fetch it from a system, whereas in docs examples it's easier to fetch from `World`.

torkleyy:
Please review again, @Aceeri

torkleyy:
Thanks!

bors r+

PR #321, opened 2017-12-25
Replace specs_website with just the book

### LINKED ISSUE #318: Website does not say what specs actually is
Sorry, but it is really terrible when that happens...

torkleyy:
Fixes #318

---
This change is [Reviewable](https://reviewable.io/reviews/slide-rs/specs/321)


icefoxen:
Not sure why I'm ping'ed on this, but it seems like a reasonable solution to me.

torkleyy:
Thanks, I pinged you because you created the issue and I wanted to know if you're content with this before I close the issue.

bors r+

PR #322, opened 2017-12-25
Document aliasing rules for resources

### LINKED ISSUE #314: Document that fetching both ReadStorage and WriteStorage isn't allowed
@Xaeroxe wrote in a previous PR review:

> We should probably also add a more formalized way of saying: If you're retrieving both a WriteStorage and a ReadStorage for a component at the same time you're doing it wrong and your code won't work. (That's a common mistake I see people making)

torkleyy:
Fixes #314

---
This change is [Reviewable](https://reviewable.io/reviews/slide-rs/specs/322)

### DIFF
--- a/book/src/06_system_data.md
+++ b/book/src/06_system_data.md
@@ -1,5 +1,29 @@
 # [DRAFT] System Data
 
+Every system can request data which it needs to run. This data can be specified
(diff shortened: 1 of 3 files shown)


torkleyy:
Thanks for the review!

bors r+

PR #328, opened 2018-01-02
Add drop method to UnprotectedStorage trait.

omni-viral:
Default-implemented via remove.
Can be overridden for some special components (I'd use this in amethyst renderer).
`AnyStorage::remove` is replaced with `AnyStorage::drop` which doesn't return anything thus avoiding unnecessary boxing.

---
This change is [Reviewable](https://reviewable.io/reviews/slide-rs/specs/328)

### DIFF
--- a/src/storage/mod.rs
+++ b/src/storage/mod.rs
@@ -14,8 +14,7 @@ use std;
 use std::marker::PhantomData;
 use std::ops::{Deref, DerefMut, Not};
 
-use hibitset::{BitSet, BitSetNot};
-use mopa::Any;
+use hibitset::{BitSet, BitSetNot, BitSetLike};
 use shred::Fetch;
 
 use self::drain::Drain;
@@ -55,16 +54,18 @@ unsafe impl<'a> DistinctStorage for AntiStorage<'a> {}
 
 /// A dynamic storage.
 pub trait AnyStorage {
-    /// Remove the component of an entity with a given index.
-    fn remove(&mut self, id: Index) -> Option<Box<Any>>;
+    /// Drop components of given entities.
+    fn drop(&mut self, entities: &[Entity]);
 }
 
 impl<T> AnyStorage for MaskedStorage<T>
 where
     T: Component,
 {
-    fn remove(&mut self, id: Index) -> Option<Box<Any>> {
-        MaskedStorage::remove(self, id).map(|x| Box::new(x) as Box<Any>)
+    fn drop(&mut self, entities: &[Entity]) {
+        for entity in entities {
+            MaskedStorage::drop(self, entity.id());
+        }
     }
 }
 
@@ -106,7 +107,7 @@ pub enum InsertResult<T> {
 /// The `UnprotectedStorage` together with the `BitSet` that knows
 /// about which elements are stored, and which are not.
 #[derive(Derivative)]
-#[derivative(Default(bound = ""))]
+#[derivative(Default(bound = "T::Storage: Default"))]
 pub struct MaskedStorage<T: Component> {
     mask: BitSet,
     inner: T::Storage,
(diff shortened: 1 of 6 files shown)


omni-viral:
@torkelyy Maybe `discard` instead of `drop`? Because in my storage wrapper where I override the method component doesn't get dropped but cached instead.

src/storage/storages.rs
```diff
@@ -174,15 +175,13 @@ unsafe impl<T> DistinctStorage for NullStorage<T> {}
 pub struct VecStorage<T>(Vec<T>);
 
 impl<T> UnprotectedStorage<T> for VecStorage<T> {
-    unsafe fn clean<F>(&mut self, has: F)
```
torkleyy: Did you benchmark this?

torkleyy:
@omni-viral Which method are we talking about here? `AnyStorage`'s or `UnprotectedStorage`'s?

src/storage/storages.rs
omni-viral: Nope.
I'll try to compare old way (but checking index directly in bitset instead of function)
and iteration over bitset.

omni-viral:
> @omni-viral Which method are we talking about here? AnyStorage's or UnprotectedStorage's?

@torkleyy both❓

torkleyy:
My confusion arose because a storage wrapper can not 'override' `AnyStorage::whatever`. I wouldn't change `clean` without a clear motivation as to not make unnecessary breaking changes, `AnyStorage::drop` can be renamed to `discard`, I'm fine with that.

omni-viral:
I need to change `clean` so it can be implemented by calling `remove` with indices from bitset.
I need to implement it that way in my storage wrapper because it need to take all removed components instead of dropping them.

torkleyy:
bors try

omni-viral:
I've rebased and squashed it. Should be fine now.

src/storage/mod.rs
```diff
@@ -485,12 +495,12 @@ where
 }
 
 /// Used by the framework to quickly join components.
-pub trait UnprotectedStorage<T>: Default + Sized {
+pub trait UnprotectedStorage<T> {
     /// Clean the storage given a check to figure out if an index
     /// is valid or not. Allows us to safely drop the storage.
```
torkleyy: Outdated docs

src/storage/storages.rs
torkleyy: Did you do this?

src/storage/storages.rs
```diff
@@ -174,13 +175,13 @@ unsafe impl<T> DistinctStorage for NullStorage<T> {}
...
+    unsafe fn clean<B>(&mut self, has: B)
     where
-        F: Fn(Index) -> bool,
+        B: BitSetLike,
     {
         use std::ptr;
         for (i, v) in self.0.iter_mut().enumerate() {
```
torkleyy: I'm unsure if this is faster than iterating over the bitset itself.

omni-viral: It depends

torkleyy: I'm sorry, but this answer really doesn't help. Could you elaborate, please?

omni-viral: If most entities has the component - enumerating and check may be faster than iterating over the bitset. Otherwise it must be faster to iterate over bitset.

src/storage/storages.rs
omni-viral: There was no significant difference as expected.

src/storage/storages.rs
omni-viral: Anyway. Changing nothing can't make things to worse.

torkleyy:
It's an improvement, that's true.

bors r+

PR #337, opened 2018-01-13
Saveload overhaul

### LINKED ISSUE #138: Serializing/Deserializing the `Entity` struct
Some components might rely on "pointing" to another entity and store that entity inside them which can become annoying when dealing with serialization and deserialization.

The main problem is making sure that the entity that the deserialized component is pointing to is actually alive. This is not as much of a  problem when deserializing entire worlds since the world deserialization can just use the entities as offsets into the entity list it was given, but when deserializing a singular entity then there are no guarantees.

We might just be able to assume that the user checks for a correct entity (Or it uses `Option<Entity>` and defaults to `None` if the entity does not exist.)

### LINKED ISSUE #282: saveload tracking issue
* `WorldDeserialize` is missing a `new` function
* Both `World*` structs should only contain generic storages, e.g. `Deref<Target = MaskedStorage>`

cc @omni-viral

torkleyy:
## TODO

* [x] Commit messages need to be improved
* [x] There is some weird lifetime error that I don't really understand

## Motivation for this PR

* Rename some items to make the code easier to understand
* Remove a couple of unnecessary bounds
* Allow passing references of storages (previously you could not reuse the `WriteStorage`)

---

Fixes #138 
Fixes #282 

---
This change is [Reviewable](https://reviewable.io/reviews/slide-rs/specs/337)


torkleyy:
This is the error I'm getting:

```
&'b mut self,
   |         ------------ these two types are declared with different lifetimes...
36 |         entities: &'b EntitiesRes,
37 |         markers: &'b mut WriteStorage<'a, M>,
   |                          -------------------
...
47 |             markers,
   |             ^^^^^^^ ...but data from `self` flows into `markers` here
```

omni-viral:
@torkleyy the problem is in "de.rs" line 129

torkleyy:
OMG this is such a stupid mistake... Thanks for your help!

omni-viral:
The error message is confusing. I've spotted it by chance.

torkleyy:
@Xaeroxe is going to continue this PR, so here's what I planned:

It turned out that the current code doesn't really allow multiple allocators and has some weird interaction with the marker component storage. Also, you can not have multiple allocators for the same marker type which should be possible.

Therefore, I've been thinking that it would make sense to put the entity -> marker and marker -> entity mapping both into the allocator (also see [these messages on Gitter](https://gitter.im/slide-rs/specs?at=5a5b20a51dcb91f1774acbc1)). You'll likely be using a `BitSet` in order to allow joining over the markers.

That's all I remember so far, please feel free to ask if something doesn't make sense.

Xaeroxe:
Currently looking into simplifying the marker type by stripping away what I feel may be metadata not appropriate to store here.  Some examples:

- `Marker::Identifier` and `Marker::id()`: This feels extraneous, a marker's primary purpose should be to identify, rather than it being just one property of them.  We're cloning these things all the time so allowing additional metadata to be associated with them may result in confusing behavior as users lose track of where their metadata actually is.

- `Marker::update()`: Similar to the above, this feels excessive and easy to get confused on.  This kind of meta data also doesn't exactly fulfill its proposed use case imo as the actual update signature doesn't provide enough context to give a meaningful audit of changes.  In addition to this, the constant cloning of these markers means this data might even be misleading in some instances.

I've removed the `Component` impl for the `Marker` as well, since it was preventing use of multiple allocators with the same marker type for varying purposes.  I'm also making the `MarkerAllocator` trait able to map in the opposite direction in order to make up for this removal.

torkleyy:
@Xaeroxe I was also confused by this at the beginning, but @omni-viral said that it can be useful for e.g. a generational id or something. I'm not quite sure about it.

Xaeroxe:
This metadata may be useful in the right contexts, I just don't think this is the right place to be trying to store and generate it.  It should be handled by something with a more clear picture of what exactly it's trying to track.  Typically the user, with their own data structures.

torkleyy:
I'm also in favor of simplifying it, I just left it as is because @omni-viral told me so. Let's wait for them to respond.

omni-viral:
I'm all for simplifying. But would it be possible to implement otherwise?

Xaeroxe:
Can you clarify what you're referring to when you say "it"?

omni-viral:
Network synchronization that checks for sequence id and rejects if entity in packet comes with less sequence id then one already in the world.

Xaeroxe:
Oh, yes that will definitely be possible.  That kind of meta data could be stored in another component that the network system itself queries and writes to.  It can exist completely separate from the marker system.

omni-viral:
Ok. And how deserializer would know how to deal with it?

Xaeroxe:
I don't believe it's the deserializer's job to perform that kind of filtering.  It should be handled in the network layer.

omni-viral:
But if deserializer will overwrite components with outdated data you won't be able to restore them.

Xaeroxe:
I'm finally getting back to this.  I see your point now @omni-viral .  

If a marker's type is meant to be a semantic detail that controls how the underlying identifier is used during loading I'm no longer sure what we gain from making these not components @torkleyy .  Can you please provide a use case where we needed a single marker type to have multiple storages?

Xaeroxe:
Rebased, review comments addressed, example fixed, formatted with rustfmt 0.3.4-nightly.  I think this is ready for review.

src/common.rs
```diff
@@ -161,8 +161,7 @@ impl Errors {
 #[derive(Derivative)]
 #[derivative(Default(bound = ""))]
 pub struct Merge<F> {
-    #[derivative(Default(value = "PhantomData"))]
-    future_type: PhantomData<F>,
+    #[derivative(Default(value = "PhantomData"))] future_type: PhantomData<F>,
```
torkleyy: Oh, is this some new rustfmt setting again?

torkleyy:
I think it's good as-is, so feel free to merge @WaDelma

src/common.rs
omni-viral: Is this attribute necessary at all?

torkleyy: Yes, otherwise `F` would have to implement `Default`.

omni-viral: No. PhantomData implements default for any type parameter.

torkleyy: It does now but I think it was a rather recent fix which I don't want to rely on yet.

examples/saveload.rs
```diff
@@ -75,14 +102,25 @@ fn main() {
...
+        fn run(&mut self, (ents, pos, vel, markers): Self::SystemData) {
+            let mut ser = ron::ser::Serializer::new(Some(Default::default()), true);
+            SerializeComponents::<NoError, U64Marker>::serialize(
+                &(&pos, &vel),
+                &ents,
+                &markers,
+                &mut ser,
+            ).unwrap_or_else(|e| eprintln!("Error: {}", e));
+            // TODO: Specs should return an error which combines serialization
+            // and component errors.
```
WaDelma: Is this TODO to be left to future PR?

examples/saveload.rs
```diff
@@ -75,14 +102,25 @@ fn main() {
...
-
-            world.remove_serialized();
+        type SystemData = (
+            Entities<'a>,
+            ReadStorage<'a, Pos>,
+            ReadStorage<'a, Mass>,
+            ReadStorage<'a, U64Marker>,
+        );
+
+        fn run(&mut self, (ents, pos, vel, markers): Self::SystemData) {
```
WaDelma: s/vel/mass/

examples/serialize.rs
```diff
@@ -48,7 +48,7 @@ fn main() {
             let entity_list: Vec<_> = (&*data.entities).join().collect();
 
             // Remove all components
-            for (entity, _) in (&*data.entities, &data.comp.check()).join() {
+            for (entity, _) in (&*data.entities, data.comp.mask().clone()).join() {
```
WaDelma: Can't you use `drain` here?

src/saveload/marker.rs
```diff
@@ -109,49 +110,92 @@ impl<'a> EntityBuilder<'a> {
...
     /// Id of the marker
-    type Identifier: Copy + Debug + Eq + Hash;
-
+    type Identifier;
     /// Allocator for this `Marker`
     type Allocator: MarkerAllocator<Self>;
 
-    /// Get this marker internal id
+    /// Get this marker internal id.
+    /// This value may never change.
```
WaDelma: As the update method says that changes in return value doesn't cause UB, I think the wording should be relaxed.
Maybe something like: "The value of this method should be constant".

src/saveload/marker.rs
```diff
@@ -161,63 +205,73 @@ pub trait Marker: Component + DeserializeOwned + Serialize + Copy {
...
         storage.insert(entity, marker);
         entity
     }
 
+    /// Create new unique marker `M` and attach it to entity.
+    /// Or get old marker if this entity is already marked.
+    fn mark<'m>(&mut self, entity: Entity, storage: &'m mut WriteStorage<M>) -> (&'m M, bool) {
+        let mut new = false;
+
+        let marker = storage.entry(entity).unwrap().or_insert_with(|| {
```
WaDelma: What is with this `unwrap` here? Is it expected to not panic? I would change it to `expect` that contains explanation.

src/saveload/ser.rs
```diff
@@ -1,166 +1,176 @@
...
+        E: Display,
+        S: Serializer,
+    {
+        let mut serseq = serializer.serialize_seq(None)?;
+        let ids = |entity| -> Option<M> { markers.get(entity).cloned() };
+        for (entity, marker) in (&*entities, &*markers).join() {
+            serseq.serialize_element(&EntityData::<M, Self::Data> {
+                marker: marker.clone(),
+                components: self.serialize_entity(entity, &ids)
+                    .map_err(ser::Error::custom)?, // TODO: revise
```
WaDelma: What does the TODO here mean? If it's still relevant, it should be made more informative.

src/saveload/ser.rs
```diff
@@ -1,166 +1,176 @@
...
+                };
+                for (entity, marker) in to_serialize {
+                    serseq.serialize_element(&EntityData::<M, Self::Data> {
+                        marker,
+                        components: self.serialize_entity(entity, &mut ids)
+                            .map_err(ser::Error::custom)?,
+                    })?;
+                }
+            }
+            to_serialize = add;
```
WaDelma: This code is bit hard to decipher, but I don't know how to improve it.
Also you could reuse `to_serialize` and only do two allocations, but I don't know if it's worth it.

src/saveload/storages.rs
```diff
@@ -0,0 +1,106 @@
...
+        WriteStorage::insert(self, entity, comp);
+    }
+
+    fn remove(&mut self, entity: Entity) {
+        WriteStorage::remove(self, entity);
+    }
+
+    //    fn join(&'b mut self) -> Self::Join {
+    //        self
+    //    }
```
WaDelma: Remove commented code

src/saveload/storages.rs
```diff
@@ -0,0 +1,106 @@
...
+        WriteStorage::insert(*self, entity, comp);
+    }
+
+    fn remove(&mut self, entity: Entity) {
+        WriteStorage::remove(*self, entity);
+    }
+
+    //    fn join(&'c mut self) -> Self::Join {
+    //        &mut **self
+    //    }
```
WaDelma: Remove commented code

src/common.rs
omni-viral: It does since 1.0.0

torkleyy:
Not sure why I'm doing this everywhere then.. I'll check later.

src/common.rs
torkleyy: Ahhh, the attribute on `PhantomData` isn't required, but we need `derivative` here.

LinearZoetrope:
I had an issue with saveload that crashed on deserialize with `NullStorage`, any chance the overhaul fixes this? I was going to open a separate issue, but figured this may be a better place to report it since it's a pretty big overhaul.

torkleyy:
@LinearZoetrope What kind of crash was that? I never experienced any crashes, so can't say if I fixed them.

LinearZoetrope:
I'll have to retest, but if I recall correctly, it was claiming anything stored in NullStorage wasn't present when deserializing.

torkleyy:
@LinearZoetrope I'll check it and add a test to ensure it's working.

examples/saveload.rs
Xaeroxe: Yes

Xaeroxe:
@torkleyy I believe this is ready to be merged.

Xaeroxe:
Ok I've just rebased this, it should be ready to merge.  Recommend one more review prior to merging, as the rebase kind of aggressively preferred to keep code from this PR where there were conflicts.

src/lib.rs
```diff
@@ -1,4 +1,4 @@
-#![deny(missing_docs)]
+#![warn(missing_docs)]
```
WaDelma: So there are missing documentations?

src/saveload/ser.rs
```diff
@@ -1,166 +1,179 @@
...
-    for (entity, marker) in (&**entities, &*markers).join() {
-        serseq.serialize_element(&EntityData::<M, E, T> {
-            marker: *marker,
-            components: T::save(entity, storages, &ids).map_err(ser::Error::custom)?,
-        })?;
+    /// The data representation of the components.
+    type Data: Serialize;
+
+    /// Serialize the components of a single entiy using a entity -> marker mapping.
+    fn serialize_entity<F>(&self, entity: Entity, ids: F) -> Result<Self::Data, E>
```
WaDelma: s/entiy/entity/

torkleyy:
Errors for missing docs make it rather hard to experiment with new features.

PR #343, opened 2018-02-10
Upgrade syn and quote, rename component attrib to storage

torkleyy:
Not meant to merge yet.

---
This change is [Reviewable](https://reviewable.io/reviews/slide-rs/specs/343)

### DIFF
--- a/specs-derive/src/lib.rs
+++ b/specs-derive/src/lib.rs
@@ -6,61 +6,55 @@
 extern crate proc_macro;
 #[macro_use]
 extern crate quote;
+#[macro_use]
 extern crate syn;
 
 use proc_macro::TokenStream;
 use quote::Tokens;
-use syn::{Ident, MacroInput, MetaItem, NestedMetaItem};
+use syn::{Path, DeriveInput};
+use syn::synom::Synom;
 
 /// Custom derive macro for the `Component` trait.
 ///
 /// ## Example
 ///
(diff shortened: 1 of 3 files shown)


specs-derive/src/lib.rs
```diff
@@ -6,61 +6,59 @@
...
-            _ => None,
-        })
-        .and_then(|attr| match *attr {
-            MetaItem::Word(ref ident) => Some(ident),
-            _ => None,
-        })
-        .cloned()
-        .unwrap_or(Ident::new("::specs::storage::DenseVecStorage"));
+        .map(|attr| attr.tts.clone())
+        //.unwrap_or_else(|| quote!("(DenseVecStorage)").into());
```
torkleyy: @dtolnay What's the recommended way for providing a default path here?

dtolnay: Maybe something like this:

```rust
let storage = ast.attrs.iter()
    .find(|attr| attr.path.segments[0].ident == "storage")
    .map(|attr| syn::parse2::<StorageAttribute>(attr.tts.clone()).unwrap().storage)
    .unwrap_or_else(|| parse_quote!(DenseVecStorage));
```

torkleyy: Thank you!

torkleyy:
Thanks @Rhuagh 

bors r+

PR #353, opened 2018-02-21
Use "parallel" feature flag

BartAdv:
This allows it to be built for targets such as wasm32


BartAdv:
Hm, it still doesn't build when referenced in my project via `specs = { git = "https://github.com/bartadv/specs", rev = "b43e8623165685b78b3b8a87a7fc52faa41c8612", default-features = false }`. Ehh those features, I still don't feel like grokking it :)

Anyway, I feel like there's some cleanup to be done in features in Cargo.toml here. I've added `default` feature, but I'm not sure what's the reason behind `common` feature - was it meant to be default one?

Cargo.toml
```diff
@@ -21,13 +21,12 @@ travis-ci = { repository = "slide-rs/specs" }
 crossbeam = "0.3.0"
 derivative = "1"
 fnv = "1.0"
-hibitset = { version = "0.4.1", features = ["parallel"] }
+hibitset = { version = "0.4.1", default-features = false }
 mopa = "0.2"
 shred = "0.6.0"
 shrev = "0.8.0"
-shred-derive = "0.4"
```
torkleyy: I think you accidentally deleted this line.

torkleyy:
Also, maybe we can add a Travis job to ensure it builds on wasm32 in the future.

Cargo.toml
BartAdv: Oops, restored

BartAdv:
Note - I'd still like to check what is the cause of the build problem in my project, whether it requires some attention in specs Cargo.toml or not. Hadn't got time for this lately though ...

BartAdv:
OK, it was just missing `default-features = false` on shred dependency...

src/join.rs
```diff
@@ -163,6 +168,7 @@ pub trait Join {
 /// The purpose of the `ParJoin` trait is to provide a way
 /// to access multiple storages in parallel at the same time with
 /// the merged bit set.
+#[cfg(feature = "parallel")]
```
torkleyy: Meh, this isn't really ideal. One should be able to compile just the same code for wasm, but this disallows `ParJoin`. Instead we'd need a fallback for it. But I'm not sure if we can use rayon at all on wasm :(

WaDelma: It seems that `rayon` should build with wasm (rayon-rs/rayon#503), but as wasm doesn't support threads it just panics when using it.

BartAdv: Heh, what's the point of being able to compile something only for it to blow up later? I thought this behaviour of wasm32-unknown-unknown target is something that's just done so that it can be shipped earlier. Anyhow, I can change it if you wish

BartAdv: Unless the fallback would be to define ParJoin as Join?

WaDelma: I wasn't implying we should use it, but just providing info.

torkleyy:
I'm planning on getting out a new Specs version soon, is anybody still interested in fixing this? It's not really a priority for me right now, although I'd like to target wasm in the future.

torkleyy:
## How to fix this issue

* update rayon to fall back to sequential iteration on wasm
* update this PR

PR #354, opened 2018-02-22
Tests, DistinctStorage bounds, narrower generics for better inference.

Aceeri:
Solves a soundness issue with `RestrictedStorage` that disallows storages that do not implement `DistinctStorage` to not be allowed to be run in parallel.

Adds some basic tests (mainly for compile time guarantees, but with a bit of sanity checking).

Adds `ImmutableParallelRestriction` as a concrete bound on the `Join`/`ParJoin` implementations of `&'a RestrictedStorage` so that Rust can infer `storage.restrict_mut().join()` to be a mutable restricted storage, instead of having to specify `(&mut storage.restrict_mut()).join()`.

---
This change is [Reviewable](https://reviewable.io/reviews/slide-rs/specs/354)

### DIFF
--- a/src/storage/restrict.rs
+++ b/src/storage/restrict.rs
@@ -86,6 +86,8 @@ where
 {
 }
 
+// For explanation of why `ImmutableParallelRestriction` instead of `RT: ImmutableAliasing`, see
+// the `Join` implementation of `RestrictedStorage`.
 unsafe impl<'rf, 'st: 'rf, B, T, R> ParJoin
     for &'rf RestrictedStorage<'rf, 'st, B, T, R, ImmutableParallelRestriction>
 where
@@ -166,6 +168,15 @@ where
     }
 }
 
+// This is actually fine for `SequentialRestriction` (or any restriction with
+// `ImmutableAliasing` as well, however Rust then requires differentiation
+// between the two. Since there is no reason to use `SequentialRestriction`
+// and only use it as an `ImmutableRestriction`, might as well improve ergonomics
+// for common-case calls.
+//
+// e.g.
+// With `ImmutableAliasing`: `(&mut storage.restrict_mut())`
+// With `ImmutableParallelRestriction`: `storage.restrict_mut()`
 impl<'rf, 'st: 'rf, B, T, R> Join for &'rf RestrictedStorage<'rf, 'st, B, T, R, ImmutableParallelRestriction>
 where
     T: Component,
--- a/src/storage/restrict.rs
+++ b/src/storage/restrict.rs
@@ -6,7 +6,7 @@ use std::ops::{Deref, DerefMut};
 use hibitset::BitSet;
 
 use join::{Join, ParJoin};
-use storage::{MaskedStorage, Storage, UnprotectedStorage};
+use storage::{DistinctStorage, MaskedStorage, Storage, UnprotectedStorage};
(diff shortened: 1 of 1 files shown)


src/storage/restrict.rs
```diff
@@ -165,7 +166,7 @@ where
     }
 }
 
-impl<'rf, 'st: 'rf, B, T, R, RT> Join for &'rf RestrictedStorage<'rf, 'st, B, T, R, RT>
+impl<'rf, 'st: 'rf, B, T, R> Join for &'rf RestrictedStorage<'rf, 'st, B, T, R, ImmutableParallelRestriction>
```
WaDelma: Shouldn't this be `ImmutableAliasing` or am I understanding it wrong?

WaDelma: Actually I am not sure what's wrong with allowing it for `MutableParallelRestriction` too...
The `ParJoin` is implemented for mutable reference of it so you cannot use `join` to break anything as `Entry`s are tied to `rf` lifetime.
(Not that it's that useful for it)

WaDelma: Or is this what you mean be better inference? If that's the case it's fine but I would like to have comment here indicating it.

Aceeri: In all technicality yes it can be any restriction, but it is all but useless except with `ImmutableParallelRestriction` and confuses Rust since it defaults to immutable referenced implementations.

torkleyy:
> (Side note: It would be nice if we had compiletest to check that certain patterns don't work. This way we could ensure we don't accidentally allow things that aren't safe.)

:+1:

WaDelma:
I haven't used `RestrictedStorage` and I am not sure are you supposed to be able to use it have mutable optional components in parallel join. I don't see way how to use the API like that as nothing that uses `Entry` returns optional.

I actually cannot see what `par_restrict_mut` can be used for that isn't available without it. Am I missing something here?

Also it would be good to have examples how in the parallel restrict docs how to use them.

These are bit outside of the scope of this PR, but they came up while reading the code.

WaDelma:
Only thing I require from this PR is to have comment on the `Join` implementation that adds `ImmutableParallelRestriction` that explains it's for type inference. Other stuff should probably come with different PR.

Aceeri:
> I actually cannot see what par_restrict_mut can be used for that isn't available without it. Am I missing something here?

It restricts the storage to only getting the component unchecked later on. It is unsafe to get other entities components mutably or immutably since they could be getting accessed in another thread. The purpose is mainly for deferring access for FlaggedStorage honestly.

WaDelma:
I thought it could be used to have optional components in parallel iteration. Currently there is no way to do it.

torkleyy:
Can you elaborate @WaDelma ? I'm not sure I quite understand. I'd love to hear about missing features before the next release (which is very soon :smile:).

WaDelma:
Basically I want to do equivalent in parallel:
```rust
fn run(&mut self, (entities, positions): Self::SystemData) {
    for (entity) in (&*entities).join() {
        if let Some(position) = positions.get(entity) {
            // Do something when entity has component
        }
        // Do something always
    }
}
```

I haven't been able to figure out how to do this with the current API.

torkleyy:
Unfortunately, this PR got stale and I think this API needs some more work to be fully functional. I'm working on a solution for all this, thus I'll close this PR for now.

PR #356, opened 2018-02-25
Improve parallel join chapter in the book.

WaDelma:
This change is [Reviewable](https://reviewable.io/reviews/slide-rs/specs/356)

### DIFF
--- a/book/src/08_parallel_join.md
+++ b/book/src/08_parallel_join.md
@@ -2,8 +2,8 @@
 
 As mentioned in the chapter dedicated on how to [dispatch][c3] systems,
 Specs automatically parallelizes system execution when there are non-conflicting
-system data requirements (each resource in the system data may only be accessed either
-once mutably or many times immutably from multiple systems).
+system data requirements (Two `System`s conflict if their `SystemData` needs access
+to the same resource where at least one of them needs write access to it).
 
 [c3]: ./03_dispatcher.html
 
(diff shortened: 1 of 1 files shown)


book/src/08_parallel_join.md
```diff
@@ -2,11 +2,13 @@
 
 As mentioned in the chapter dedicated on how to [dispatch][c3] systems,
 Specs automatically parallelizes system execution when there are non-conflicting
-system data requirements (every system data may only be accessed
-mutably once xor immutably from multiple systems).
+system data requirements (each resource in the system data may only be accessed either
```
torkleyy: Maybe the text in brackets should be: "two `System`s conflict if their `SystemData` needs access to the same resource where at least one of both needs write access"

book/src/08_parallel_join.md
```diff
@@ -45,8 +47,10 @@ fn run(&mut self, (vel, mut pos): Self::SystemData) {
 }
 ```
 
-The `par_join` method produces a type implementing [`rayon`s `ParallelIterator`][ra]
+There is always overhead in parallelization, so you should always carefully profile to see if there are benefits in the switch. If you have only few things to iterate over then sequential join is faster.
```
torkleyy: Please remove the second "always". I'd probably put this into a little "box" by placing "\>" before the text.

book/src/08_parallel_join.md
```diff
@@ -2,11 +2,13 @@
...
 As mentioned in the chapter dedicated on how to [dispatch][c3] systems,
 Specs automatically parallelizes system execution when there are non-conflicting
-system data requirements (every system data may only be accessed
-mutably once xor immutably from multiple systems).
+system data requirements (Two `System`s conflict if their `SystemData` needs access
+to the same resource where at least one of them needs write access to it).
 
 [c3]: ./03_dispatcher.html
 
+## Basic parallization
```
itmuckel: Typo in parallelization.

WaDelma: I constantly make that typo, thanks for spotting!

PR #357, opened 2018-02-25
Vec vs VecStorage vs WriteStorage over VecStorage iteration benchmark

theor:
This change is [Reviewable](https://reviewable.io/reviews/slide-rs/specs/357)

### DIFF
--- a/benches/storage_perf.rs
+++ b/benches/storage_perf.rs
@@ -1,0 +1,87 @@
+use criterion::{Bencher, Criterion};
+use specs::prelude::*;
+
+#[derive(Clone, Debug)]
+pub struct GridVelocity {
+    pub dx: i16,
+    pub dy: i16,
+}
+impl GridVelocity {
+    pub fn new() -> Self { GridVelocity { dx:0, dy:0 } }
+}
+impl Component for GridVelocity {
+    type Storage = VecStorage<Self>;
+}
+
+fn bench_vec(b: &mut Bencher) {
+    let mut vels: Vec<GridVelocity> = (0..1000*1000).map(|_i| { GridVelocity::new() }).collect();
+    b.iter(|| {
+        vels.iter_mut().for_each(|vel| {
+        // for vel in  {
+            vel.dx += 1;
+        });
+    });
+} 
+
+pub fn bench_vecstorage(b: &mut Bencher) {
+    use std::default::Default;
+    use specs::storage::UnprotectedStorage;
+
+    let mut vel_storage: VecStorage<GridVelocity> = Default::default();
+
+    for i in 0..1000*1000 {
+        unsafe { vel_storage.insert(i, GridVelocity::new()); }
+    }
+
+    b.iter(|| {
+        for i in 0..1000*1000 {
+            let mut vel = unsafe { vel_storage.get_mut(i) };
+            vel.dx += 1;
+        }
+    });
+}  
+
+pub fn bench_storage_entity_prefetch(b: &mut Bencher) {
+    let mut world = World::new();
+    world.register::<GridVelocity>();
+    for _i in 0..1000*1000 {
+         world.create_entity().with(GridVelocity::new()).build();
+    }
+
+    let mut vel_storage = world.write::<GridVelocity>();
+    let ents = world.entities();
+    let entities: Vec<Entity> = (0..1000*1000).map(|i| ents.entity(i)).collect();
+
+    b.iter(|| {
+        for e in entities.iter() {
+            let mut vel = vel_storage.get_mut(*e).unwrap();
+            vel.dx += 1;
+        }
+    });
+}  
+
+#[cfg_attr(rustfmt, rustfmt_skip)]
+fn raw_vect_perf(c: &mut Criterion) {
+    c.bench_function("Vec iteration", bench_vec);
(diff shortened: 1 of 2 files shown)


theor:
Results on my machine:
```
Vec iteration           time:   [381.70 us 390.14 us 399.00 us]
                        change: [-3.1627% -0.4722% +2.0912%] (p = 0.73 > 0.05)
                        No change in performance detected.
Found 1 outliers among 100 measurements (1.00%)
  1 (1.00%) high mild

VecStorage iteration    time:   [416.57 us 422.85 us 429.46 us]
                        change: [-2.2356% +0.8992% +3.8720%] (p = 0.56 > 0.05)
                        No change in performance detected.
Found 5 outliers among 100 measurements (5.00%)
  3 (3.00%) high mild
  2 (2.00%) high severe

WriteStorage iteration  time:   [4.1544 ms 4.1694 ms 4.1859 ms]
                        change: [-3.2153% -2.0943% -0.9862%] (p = 0.00 < 0.05)
                        Change within noise threshold.
Found 11 outliers among 100 measurements (11.00%)
  1 (1.00%) low mild
  8 (8.00%) high mild
  2 (2.00%) high severe
```

benches/storage_perf.rs
```diff
@@ -0,0 +1,73 @@
...
+
+    let mut vel_storage = world.write::<GridVelocity>();
+    let ents = world.entities();
+    let entities: Vec<Entity> = (0..1000 * 1000).map(|i| ents.entity(i)).collect();
+
+    b.iter(|| {
+        for e in entities.iter() {
+            let mut vel = vel_storage.get_mut(*e).unwrap();
+            vel.dx += 1;
+        }
```
WaDelma: I am not sure what is the idea behind this prefetch thing. Why don't directly iterate entities? Collecting them to vector is expensive operation which I presume would be performed whenever new things are added to world.

WaDelma: I also don't know why are avoiding `join` which is the usual way to use storages. Getting them like that is more expensive as you need to check liveness. (There is some overhead on `join` due to iterating `BitSet`)

benches/storage_perf.rs
```diff
@@ -0,0 +1,73 @@
...
+    type Storage = VecStorage<Self>;
+}
+
+fn bench_vec(b: &mut Bencher) {
+    let mut vels: Vec<GridVelocity> = (0..1000 * 1000).map(|_i| GridVelocity::new()).collect();
+    b.iter(|| {
+        vels.iter_mut().for_each(|vel| {
+            vel.dx += 1;
+        });
+    });
```
WaDelma: You are basically comparing here iterating vector vs doing manually iteration with indices and using `get_unchecked` (the bench_vecstorage).
Have you tried to benchmark it directly on `Vec`?

WaDelma: The difference might be because there are no `#[inline]` annotations on the `UnprotectedStorage` methods.

benches/storage_perf.rs
theor: In my actual use case, I have a different component containing a list of entities, which is built by looking for other entities having this component. So in my case, I can guarantee they will have this component, and I guess I'd handle removal in a separate system running later.

About prefetching: basically I have a Vec<Entity>. I guess I could create a hibitset with that and join it ? otherwise I tried not to bench the entity fetching, given that it is expensive.

Btw joining/iterating on hibitset seems to be crazy expensive too - I merged two systems at some points, because joining twice was slow.

benches/storage_perf.rs
theor: I'm not sure get what you want me to benchmark ? if you're talking about indexing Vecs, it was quite similar to using iterators. Even using `as_mut_ptr()` didn't give the huge perf boost I expected.

WaDelma: I was just wondering why `VecStorage` would differ at all from `Vec`.

theor: VecStorage and Vec have quite similar performance. Read/WriteStorage over VecStorage introduces the overhead

benches/storage_perf.rs
torkleyy: > I can guarantee they will have this component

If this is too slow for you and you can guarantee it, then maybe you just want to access the raw storage?

torkleyy:
@theor How do you want us to proceed?

torkleyy:
Not sure about this PR. Usually I'd comment "closing due to inactivity", but it's a benchmark that we could just merge, even though the benchmarks are rather unfair (see the points made above). I don't really care, we can either close this or merge it.

torkleyy:
Err oops, that was somehow a reflexive button press ^^ I'll leave it up to you to decide @WaDelma

WaDelma:
I am not sure. More benchmarks is good, but these three are doing conceptually different things (first iterates vector, second gets every single index of `VecStorage`, third one resolves all valid entities to vector and gets them from `VecStorage`) resulting to the same iteration.

First and third tests are generic enough use cases, but they are only testing case where every entity has the component. Second one isn't generalizable at all (Well the generalization is basically the third test).

It's just a weird mix of benchmarks. I would be more in favour merging these benchmarks if there were tests like first and third where the storages were partially filled. Also test for just iterating the `VecStorage` with join.

PR #374, opened 2018-04-24
Add contains function to Storage

Xaeroxe:
Just adds a slightly easier and better way to check for the existence of a component on an entity.

---
This change is [Reviewable](https://reviewable.io/reviews/slide-rs/specs/374)

### DIFF
--- a/src/storage/mod.rs
+++ b/src/storage/mod.rs
@@ -209,6 +209,11 @@ where
         }
     }
 
+    /// Returns true if the storage has a component for this entity.
+    pub fn contains(&self, e: Entity) -> bool {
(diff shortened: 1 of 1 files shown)


Xaeroxe:
@torkleyy I personally feel like that digs a bit too deep into how specs works internally.  More likely to confuse users than educate them on how to use the function imo.

torkleyy:
That entities can be dead is far from an internal, users have to check livenes as soon as they store an entity somewhere.

Xaeroxe:
@torkleyy I figured out a way to say it briefly, so I added the note.

PR #375, opened 2018-04-25
Rename world::read and world::write, re-export shred::Resources

Xaeroxe:
PR based on conversation here: https://gitter.im/slide-rs/specs?at=5ae0cb975d7286b43a698dbc

Basically these functions now have slightly confusing names, so this PR makes them less ambiguous.  Additionally it re-exports `shred::Resources` to make the `System::setup` function easier to use.

---
This change is [Reviewable](https://reviewable.io/reviews/slide-rs/specs/375)

### DIFF
--- a/src/prelude.rs
+++ b/src/prelude.rs
@@ -4,8 +4,8 @@
 
 pub use hibitset::BitSet;
 pub use join::{Join, ParJoin};
-pub use shred::{Dispatcher, DispatcherBuilder, Read, ReadExpect, RunNow, System, SystemData,
-                Write, WriteExpect};
+pub use shred::{Dispatcher, DispatcherBuilder, Read, ReadExpect, Resources, RunNow, System,
+                SystemData, Write, WriteExpect};
 pub use shrev::ReaderId;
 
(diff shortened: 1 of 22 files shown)


Xaeroxe:
re-submitted as I missed a spot.

Should be good to merge as soon as travis passes

torkleyy:
I'd like to have it reviewed by @WaDelma before.

src/prelude.rs
```diff
@@ -4,8 +4,8 @@
 
 pub use hibitset::BitSet;
 pub use join::{Join, ParJoin};
-pub use shred::{Dispatcher, DispatcherBuilder, Read, ReadExpect, RunNow, System, SystemData,
-                Write, WriteExpect};
+pub use shred::{Dispatcher, DispatcherBuilder, Read, ReadExpect, Resources, RunNow, System,
+                SystemData, Write, WriteExpect};
```
torkleyy: I wonder if we should simply pub crate shred.

Xaeroxe: Honestly given how integral shred is to specs it might not be a bad idea.

PR #378, opened 2018-04-27
Doc fixes

Xaeroxe:
As @torkleyy requested I reviewed the docs for specs.

I found a few sores that I've corrected.  Let me know if you can think of anything that I missed.

By the way the specs::bitset module was removed from the documentation because it's doc page was useless and had only two sentences on it.

---
This change is [Reviewable](https://reviewable.io/reviews/slide-rs/specs/378)

### DIFF
--- a/book/src/06_system_data.md
+++ b/book/src/06_system_data.md
@@ -2,7 +2,7 @@
 
 Every system can request data which it needs to run. This data can be specified
 using the `System::SystemData` type. Typical implementors of the `SystemData` trait
-are `ReadStorage`, `WriteStorage`, `Fetch`, `FetchMut` and `Entities`.
+are `ReadStorage`, `WriteStorage`, `Read`, `Write` and `Entities`.
 A tuple of types implementing `SystemData` automatically also implements `SystemData`.
 This means you can specify your `System::SystemData` as follows:
 
@@ -18,8 +18,8 @@ impl<'a> System<'a> for Sys {
 }
 ```
 
-It is very important that you don't fetch both a `ReadStorage` and a `WriteStorage`
-for the same component or a `Fetch` and a `FetchMut` for the same resource.
+It is very important that you don't request both a `ReadStorage` and a `WriteStorage`
+for the same component or a `Read` and a `Write` for the same resource.
 This is just like the borrowing rules of Rust, where you can't borrow something
 mutably and immutably at the same time. In Specs, we have to check this at
 runtime, thus you'll get a panic if you don't follow this rule.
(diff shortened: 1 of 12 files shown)


src/bitset.rs
```diff
@@ -1,3 +1,4 @@
+#![doc(hidden)]
```
torkleyy: Just remove pub from the module maybe?

book/src/06_system_data.md
```diff
@@ -18,8 +18,8 @@ impl<'a> System<'a> for Sys {
 }
 ```
 
-It is very important that you don't fetch both a `ReadStorage` and a `WriteStorage`
-for the same component or a `Fetch` and a `FetchMut` for the same resource.
+It is very important that you don't Read both a `ReadStorage` and a `WriteStorage`
```
WaDelma: First of all read should either be de-capitalized or made code.
After that the sentence is super awkward: It's implying that there is reading of a `WriteStorage`.
Not sure how to fix it.

book/src/07_join.md
```diff
@@ -1,7 +1,7 @@
 # Joining components
 
 In the last chapter, we learned how to access resources using `SystemData`.
-To access our components with it, we can just fetch a `ReadStorage` and use
+To access our components with it, we can just get a `ReadStorage` and use
 `Storage::get` to get the component associated to an entity. This works quite
```
WaDelma: It's kinda unfortunate that there is three different uses of 'get' here, but it's not that big of a problem.

src/lib.rs
```diff
@@ -214,7 +214,7 @@ pub mod saveload;
 /// Implementations and structures related to bitsets.
 ///
 /// Normally used for `Join`s and filtering entities.
-pub mod bitset;
+mod bitset;
```
WaDelma: Did you verify that there is no uses for this? (In benchmarks and examples as those are not automatically tested afaik).

src/saveload/mod.rs
```diff
@@ -4,25 +4,25 @@
...
 //!
-//! Fetching those makes it very easy to serialize or deserialize
-//! components. However, be aware that you cannot fetch storages
+//! Reading those makes it very easy to serialize or deserialize
+//! components. However, be aware that you cannot read storages
 //! used in `WorldDeserialize` with the same system. E.g.
 //! `type SystemData = (WorldDeserialize<'a, Marker, MyError, (Pos, Vel)>, WriteStorage<'a, Vel>)`
-//! is not valid since both `WorldDeserialize` and `WriteStorage` would fetch the same component
+//! is not valid since both `DeserializeComponents` and `WriteStorage` would fetch the same component
 //! storage mutably.
```
WaDelma: Is this documentation up to date on saveload stuff?

src/lib.rs
Xaeroxe: @WaDelma There's no way that there could be usages.  This module only contains trait implementations and nothing else.  The docs for it don't generate anything except for the doc comments.

src/saveload/mod.rs
Xaeroxe: @WaDelma Yes, do you have reason to believe otherwise?

WaDelma: Ah, so there are just renamings missing?
```
 //! used in `WorldDeserialize` with the same system. E.g.
 //! `type SystemData = (WorldDeserialize<'a, Marker, MyError, (Pos, Vel)>, WriteStorage<'a, Vel>)`
```

Also the the saveload example looks totally different.

torkleyy: I think World* does not exist anymore.

torkleyy: I wanted to eventually add them back but not sure now and don't have much time, I'll let you two decide.

Rhuagh:
Once this is finished, I'll go through the book and try to correct anything that's broken, incomplete or outright wrong.

Xaeroxe:
@WaDelma I fixed all the problems you pointed out, thanks for catching those!

Rhuagh:
Ready for merge?

PR #386, opened 2018-04-30
Add must_use to InsertResult

### LINKED ISSUE #385: Unused `InsertResult` does not generate a compiler warning.
I've found code that wrongly assumes that after an insert has been called for an Entity that calling `storage.get(entity).unwrap()` is not going to panic.  This may panic because the Entity could have been dead for the insert.  Is it possible for us to generate a compiler warning for unused `InsertResult` similar to how rustc generates a compiler warning for unused `Result`?

Xaeroxe:
Fixes #385

---
This change is [Reviewable](https://reviewable.io/reviews/slide-rs/specs/386)

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -60,7 +60,8 @@ pub enum Error {
     /// Wrong generation error.
     WrongGeneration(WrongGeneration),
 
-    #[doc(hidden)] __NonExhaustive,
+    #[doc(hidden)]
+    __NonExhaustive,
 }
 
 impl Display for Error {
@@ -104,7 +105,7 @@ impl StdError for Error {
 }
 
 /// Wrong generation error.
-#[derive(Debug)]
+#[derive(Debug, PartialEq, Eq)]
 pub struct WrongGeneration {
     /// The action that failed because of the wrong generation.
     pub action: &'static str,
@@ -135,7 +136,7 @@ impl StdError for WrongGeneration {
 /// An error type which cannot be instantiated.
 /// Used as a placeholder for associated error types if
 /// something cannot fail.
-#[derive(Debug)]
+#[derive(Debug, PartialEq, Eq)]
 pub enum NoError {}
 
 impl Display for NoError {
--- a/src/storage/mod.rs
+++ b/src/storage/mod.rs
@@ -20,7 +20,7 @@ use hibitset::{BitSet, BitSetLike, BitSetNot};
 use shred::{CastFrom, Fetch};
 
 use self::drain::Drain;
-use error::WrongGeneration;
+use error::{Error, WrongGeneration};
 use join::{Join, ParJoin};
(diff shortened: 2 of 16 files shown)


src/world/mod.rs
```diff
@@ -85,7 +85,10 @@ impl<'a> EntityBuilder<'a> {
     pub fn with<T: Component>(self, c: T) -> Self {
         {
             let mut storage = self.world.write_storage();
-            storage.insert(self.entity, c);
+            if let EntityIsDead(_) = storage.insert(self.entity, c) {
+                // This is technically possible, however quite unlikely.
```
torkleyy: I don't think it is, try yourself. It should not fail before the deletion is maintained.

src/storage/mod.rs
```diff
@@ -109,6 +109,7 @@ where
 pub unsafe trait DistinctStorage {}
 
 /// The status of an `insert()`ion into a storage.
+#[must_use]
```
torkleyy: Actually, couldn't this simply be a `Result<Option<T>>`?

torkleyy: I think I wanted to do this before but decided against it because of breakage.

torkleyy: My suggestion: Use the error from `error.rs` (create one if there is no such variant yet) and make this simply a type def.

src/world/mod.rs
Xaeroxe: Oh hm, I forgot the EntityBuilder has a lifetime in it.

src/storage/mod.rs
Xaeroxe: Can you clarify what that type def would look like?  I'm a bit confused.

torkleyy: Sorry, I shouldn't be comenting here given my limited time.

```
pub type InsertResult<T> = Result<Option<T>, MyError>;
```

Xaeroxe:
@torkleyy I've changed `InsertResult` for the typedef, and formatted the codebase.

src/error.rs
```diff
@@ -60,7 +60,8 @@ pub enum Error {
     /// Wrong generation error.
     WrongGeneration(WrongGeneration),
 
-    #[doc(hidden)] __NonExhaustive,
+    #[doc(hidden)]
```
torkleyy: :tada:

Xaeroxe:
@torkleyy done!

src/storage/tests.rs
```diff
@@ -107,7 +119,7 @@ mod map_test {
         let mut w = World::new();
         let mut c = create(&mut w);
 
-        c.insert(ent(1 << 25), Comp(7));
+        let _ = c.insert(ent(1 << 25), Comp(7));
```
Xaeroxe: I left the `let _ =` here because the panic we're trying to catch shouldn't come from the failed insert.

PR #395, opened 2018-05-08
book: Add a chapter about advanced component strategies

### LINKED ISSUE #332: book: explain how to point to another entity / component

### LINKED ISSUE #334: Access components of specific Entity
I created an entity with a "Visible(bool)" component that is false by default.
In a specific System I need to set this component to true.

So basically I need to get a specific Entity and set it's component. Can I do this in a System? Should the Entity be a resource? How does the "Entities" Storage work? I couldn't find answers in closed issues or the docs.

The naive approach would be to give this entity a component that's unique to it (IsTheComponentImLookingFor) and query that in a system, but this seems sooo wrong. 😆

Rhuagh:
Fixes #334 
Fixes #332 

This change is [Reviewable](https://reviewable.io/reviews/slide-rs/specs/395)

### DIFF
--- a/book/src/11_advanced_component.md
+++ b/book/src/11_advanced_component.md
@@ -1,0 +1,140 @@
+# Advanced strategies for components
+
+So now that we have a fairly good grasp on the basics of Specs, 
+it's time that we start experimenting with more advanced patterns!
+
+## Marker components
+
+Say we want to add a drag force to only some entities that have velocity, but
+let other entities move about freely without drag?
+
+The most common way is to use a marker component for this. A marker component
+is a component without any data that can be added to entities to "mark" them
+for processing, and can then be used to narrow down result sets using `Join`.
+
+Some code for the drag example to clarify:
+
+```rust,ignore
+#[derive(Component)]
+#[storage(NullStorage)]
+pub struct Drag;
+
+#[derive(Component)]
+pub struct Position {
+    pub pos: [f32; 3],
+}
+
+#[derive(Component)]
+pub struct Velocity {
+    pub velocity: [f32; 3],
+}
+
+struct Sys {
+    drag: f32,
+}
+
+impl<'a> System<'a> for Sys {
+    type SystemData = (
+        ReadStorage<'a, Drag>,
+        ReadStorage<'a, Velocity>,
+        WriteStorage<'a, Position>,
+    );
+    
+    fn run(&mut self, (drag, velocity, mut position): Self::SystemData) {
+        // Update positions with drag
+        for (pos, vel, _) in (&mut position, &velocity, &drag) {
+            pos += vel - self.drag * vel * vel;
+        }
+        // Update positions without drag
+        for (pos, vel, _) in (&mut position, &velocity, !&drag) {
+            pos += vel;
+        }
+    } 
+}
+```
+
+Using `NullStorage` is recommended for marker components, since they don't contain
+any data and as such will not consume any memory. This means we can represent them using 
+only a bitset. Note that `NullStorage` will only work for components that are ZST (i.e. a 
+struct without fields). 
+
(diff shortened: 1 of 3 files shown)


book/src/11_advanced_component.md
```diff
@@ -0,0 +1,137 @@
...
+        for (pos, vel, _) in (&mut position, &velocity, !&drag) {
+            pos += vel;
+        }
+    } 
+}
+```
+
+Using `NullStorage` is recommended for marker components, since they don't contain
+any data and as such will not consume any memory. This means we can represent them using 
+only a bitset.
```
torkleyy: Please add something like this:

> Please note that `NullStorage` only works if your component is a ZST (meaning a struct without fields).

book/src/11_advanced_component.md
```diff
@@ -0,0 +1,137 @@
...
+        }
+    }
+}
+```
+
+We could also model this as a resource (more about that in the next section), but it could 
+be useful to be able to have multiple entities following targets, so modeling this with
+a component makes sense. This could in extension be used to model large scale hierarchical
+structure (scene graphs). For a generic implementation of such a hierarchical system, check
+out the crate `specs-hiearchy`.
```
torkleyy: missing link

Rhuagh:
Fixed.

book/src/11_advanced_component.md
```diff
@@ -0,0 +1,140 @@
+# Advanced strategies for components
+
+So, now we have a fairly good grasp of the basics of Specs, so this is 
+where we start experimenting with more advanced patterns!
```
WaDelma: This sentence doesn't flow well for me. Maybe it could be rewritten to something like: "So now that we have a fairly good grasp on the basics of Specs, it's time that we start to experiment with more advanced patterns!"

book/src/11_advanced_component.md
```diff
@@ -0,0 +1,140 @@
...
+        for (pos, vel, _) in (&mut position, &velocity, !&drag) {
+            pos += vel;
+        }
+    } 
+}
+```
+
+Using `NullStorage` is recommended for marker components, since they don't contain
+any data and as such will not consume any memory. This means we can represent them using 
+only a bitset. Note that `NullStorage` will only work components that are ZST (i.e. a
```
WaDelma: s/work components/work for components/

book/src/11_advanced_component.md
```diff
@@ -0,0 +1,140 @@
...
+    type SystemData = (
+        Read<'a, ActiveCamera>,
+        ReadStorage<'a, Camera>,
+        ReadStorage<'a, Transform>,
+        ReadStorage<'a, Mesh>,
+    );
+    
+    fn run(&mut self, (active_cam, camera, transform, mesh) : Self::SystemData) {
+        let camera = camera.get(active_cam.0);
+        let view_matrix = transform.get(active_cam.0).invert();
```
WaDelma: Shouldn't there be `unwrap`s on these gets?

book/src/11_advanced_component.md
```diff
@@ -0,0 +1,140 @@
...
+            // Set world transform matrix
+            // Render mesh
+        }
+    }
+}
+```
+
+By doing this, whenever the spectator chooses a new player to follow, we simply change
+what `Entity` is referenced in the `ActiveCamera` resource, and the scene will be
+rendered from the viewpoint instead.
```
WaDelma: s/the/that/

Xaeroxe:
Cancelled to give time to address @WaDelma 's comments

Rhuagh:
Fixed.

Xaeroxe:
@WaDelma Can you review resolution and r+ if you deem appropriate?

PR #407, opened 2018-05-13
fix: Reexport ParallelIterator so users don't have to add a rayon dependency directly to use par_join

Rhuagh:
This change is [Reviewable](https://reviewable.io/reviews/slide-rs/specs/407)


Rhuagh:
Only prelude? Or put it somewhere further down ?

Issue #414, opened 2018-05-15
Allow saveload to serialize/deserialize more than 8 component types


andreivasiliu:
I'm not sure if I'm doing something horribly wrong or commendably right, but I've hit the size limit of tuples that implement the `SerializeComponents`/`DeserializeComponents` traits, which is currently 8 component types.

This will not compile:
```
 SerializeComponents::<Error, U64Marker>::serialize(
            &(positions, sizes, shapes, rooms, in_rooms, player_controllers, velocities,
              forces, aims, collision_sets, revolute_joints, jumps, animations),
            &entities,
            &markers,
            &mut serializer
        )
```

I think the proper solution here would be to implement `SerializeComponents` for any structure with `#[derive(SystemData)]`, but I don't know enough Rust to figure out how to do it myself.

Short-term I'd be happy with increasing the tuple size limit too.

omni-viral:
I'd add another custom derive (`SerializeComponents`, `DeserializeComponents`) instead of implementing everything in one.

torkleyy:
@WaDelma @Xaeroxe Do you think we should add another custom derive?

jojolepro:
Hi. Is this still an issue?

andreivasiliu:
I'm still using tuples, and have to modify 6 different places whenever adding a new component (declaration+parameter+tuple for both serialization+deserialization), since I couldn't figure out how to implement `SerializeComponents`:
https://github.com/andreivasiliu/stacked-worlds/blob/master/src/saveload.rs#L46-L48

Documenting an example of `SerializeComponents` implementation or adding an automatic derive for it would help.

My biggest issue (the limit of 8, as per the issue title) was fixed though, so feel free to close this; I assume the saveload module will eventually get better documentation/examples even without this issue.

expenses:
@andreivasiliu you can bring that down to 2 if you change it to
```rust
type SystemData = (
        Entities<'a>,
        (
                ReadStorage<'a, Position>,
                ...
        ),
        ReadStorage<'a, U64Marker>,
    );

    fn run(&mut self, (entities, components, markers): Self::SystemData)
    {
        let mut serializer = ron::ser::Serializer::new(Some(Default::default()), true);
        SerializeComponents::<Error, U64Marker>::serialize(
            &components,
            &entities,
            &markers,
            &mut serializer
        ).unwrap_or_else(|e| {
            // FIXME: handle this
            eprintln!("Error: {}", e);
});
```

or equivalent.

[...]

PR #415, opened 2018-05-15
Increase tuple size for SerializeComponents/DeserializeComponents

andreivasiliu:
The short-term solution for https://github.com/slide-rs/specs/issues/414

I tried adding more, but it looks like 16 is the maximum.

---
This change is [Reviewable](https://reviewable.io/reviews/slide-rs/specs/415)


torkleyy:
You can also just implement `SerializeComponents` / `DeserializeComponents` for your own struct by the way.

PR #419, opened 2018-05-19
Expose the wrapped storage from Read/WriteStorage

andrewhickman:
This is already possible through `Join::open` (although not very ergonomic), so there shouldn't be any soundness issues here. This is useful for using custom storages with additional methods.

---
This change is [Reviewable](https://reviewable.io/reviews/slide-rs/specs/419)

### DIFF
--- a/src/storage/mod.rs
+++ b/src/storage/mod.rs
@@ -192,6 +192,11 @@ where
     T: Component,
     D: Deref<Target = MaskedStorage<T>>,
 {
+    /// Gets the wrapped storage.
+    pub fn storage(&self) -> &T::Storage {
+        &self.data.inner
+    }
+
     /// Tries to read the data associated with an `Entity`.
     pub fn get(&self, e: Entity) -> Option<&T> {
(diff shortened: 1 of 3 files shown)


src/storage/mod.rs
```diff
@@ -192,6 +192,11 @@ where
     T: Component,
     D: Deref<Target = MaskedStorage<T>>,
 {
+    /// Gets the wrapped storage.
+    pub fn storage(&self) -> &T::Storage {
```
torkleyy: Could we name this `unprotected_storage`? I usually prefer short names, but in this case it's not what most people are looking to use, so I think it's good to have some kind of warning.

torkleyy:
Ah please also do a bump of the patch version so I can release after it's merged.

torkleyy:
Okay, this time my review comes in three parts, I have one last request: Please add your change to the changelog. That should be it now :)

torkleyy:
Specs 0.11.2 is uploaded now :)

PR #424, opened 2018-05-23
Add EntityResBuilder

Xaeroxe:
This allows building of entities within a system so long as the necessary Component storages are also available.

---
This change is [Reviewable](https://reviewable.io/reviews/slide-rs/specs/424)

### DIFF
--- a/src/world/entity.rs
+++ b/src/world/entity.rs
@@ -294,7 +294,7 @@ impl EntitiesRes {
         let entity = self.create();
         EntityResBuilder {
             entity,
-            res: self,
+            entities: self,
             built: false,
         }
     }
@@ -353,7 +353,7 @@ pub struct EntityResBuilder<'a> {
     pub entity: Entity,
     /// The active borrow to `EntitiesRes`, used to delete the entity if the
     /// builder is dropped without called `build()`.
-    pub res: &'a EntitiesRes,
+    pub entities: &'a EntitiesRes,
     built: bool,
 }
 
@@ -374,7 +374,7 @@ impl<'a> EntityResBuilder<'a> {
 impl<'a> Drop for EntityResBuilder<'a> {
     fn drop(&mut self) {
         if !self.built {
-            self.res.delete(self.entity).unwrap();
+            self.entities.delete(self.entity).unwrap();
         }
     }
 }
--- a/src/world/entity.rs
+++ b/src/world/entity.rs
@@ -5,6 +5,8 @@ use shred::Read;
 
 use error::WrongGeneration;
 use join::{Join, ParJoin};
+use storage::WriteStorage;
+use world::Component;
 
 /// An index is basically the id of an `Entity`.
 pub type Index = u32;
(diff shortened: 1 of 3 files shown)


src/world/entity.rs
```diff
@@ -331,6 +346,39 @@ impl<'a> Join for &'a EntitiesRes {
...
 unsafe impl<'a> ParJoin for &'a EntitiesRes {}
 
+/// An entity builder from `EntitiesRes`.  Allows building an entity with its
+/// components if you have mutable access to the component storages.
+pub struct EntityResBuilder<'a> {
+    /// The entity being built
+    pub entity: Entity,
+    /// The active borrow to `EntitiesRes`, used to delete the entity if the
+    /// builder is dropped without called `build()`.
+    pub res: &'a EntitiesRes,
```
torkleyy: `entities`

torkleyy:
bors delegate=Rhuagh

rmbreak:
Reviewed 3 of 3 files at r1, 1 of 1 files at r2.
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from [Reviewable](https://reviewable.io/reviews/slide-rs/specs/424)*

rmbreak:
:lgtm:

---







*Comments from [Reviewable](https://reviewable.io/reviews/slide-rs/specs/424#-:-LDCt_yL1usw6TIwk2gy:bnfp4nl)*

src/world/entity.rs
```diff
@@ -331,6 +346,39 @@ impl<'a> Join for &'a EntitiesRes {
 
 unsafe impl<'a> ParJoin for &'a EntitiesRes {}
 
+/// An entity builder from `EntitiesRes`.  Allows building an entity with its
+/// components if you have mutable access to the component storages.
+pub struct EntityResBuilder<'a> {
+    /// The entity being built
+    pub entity: Entity,
```
WaDelma: Because this is public you can change it at will. So you could partially build entity, switch it another one, forget build it and then only the switched one gets removed.

There is nothing horribly wrong with this, but it seems to just lead to confusing situations. I would make entity private and create getter for it.

WaDelma: It also is bit confusing that building doesn't actually do anything. Normally with builders this cannot be observed, but because you can switch entity you can with this.

src/world/entity.rs
```diff
@@ -331,6 +346,39 @@ impl<'a> Join for &'a EntitiesRes {
...
+    /// The active borrow to `EntitiesRes`, used to delete the entity if the
+    /// builder is dropped without called `build()`.
+    pub entities: &'a EntitiesRes,
```
WaDelma: I am also not sure why making this public would be useful, but there is no harm in this case... Well there could be if you had multiple worlds and started mixing their types, but that's rare.

Xaeroxe:
@WaDelma I'm following the pattern of the other builders in specs.  At minimum the entity needs to be pub(crate) to make the marker implementation work.  Also since the builder eats your borrow of `EntitiesRes` I figured it'd be appropriate to expose it so the user doesn't arbitrarily lose capabilities during the build.  What would you have happen during the build?

torkleyy:
I think it should stay public. If you want to make things go wrong, we can't prevent people from that. Code that changes the entity during the building definitely smells, so.. I don't think we should restrict this. As for the entities resource, I don't see how it "eats your borrow". It's immutable.

WaDelma:
The borrow is shared so it's `Copy` and will not be moved. So there shouldn't be any loss of capabilities.
Okay. It's just that it feels weird that you would even be able to swap it.

Xaeroxe:
bors r=@Rhuagh

Issue #422, opened 2018-05-20
Implement common builder trait


sinesc:
It would be convenient to have a common builder trait for EntityBuilder and LazyBuilder, so entity-building code could be reused regardless of which builder is being used.

```rust
trait Builder {
    fn with<T: specs::Component + Send + Sync>(self, c: T) -> Self;
    fn build(self) -> specs::Entity;
}
```

The name should probably be a bit more specific.

torkleyy:
@sinesc Thanks for the suggestion, I think there are even more things we could expose via traits, like inserting a component or similar.

GraemeWilde:
You would need to `use` the new trait right? So would this be a breaking change?

Aceeri:
@torkleyy Can close this issue by the way :)

PR #426, opened 2018-05-24
Added common Builder trait for EntityBuilder and LazyBuilder

Xt-8:
PR for #422.

Hopefully I did it right.

---
This change is [Reviewable](https://reviewable.io/reviews/slide-rs/specs/426)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ unsoundness!
 * Add common `Builder` trait to `EntityBuilder` and `LazyBuilder` ([#426])
 * Add common `MarkedBuilder` trait to `EntityBuilder` and `LazyBuilder` ([#426])
 
+[#426]: https://github.com/slide-rs/specs/pull/426
+
 # 0.11.4
 
 * Add `build_entity()` to `EntitiesRes` so you can use builder syntax in a system.


LinearZoetrope:
There should also probably be a serde-gated `MarkedBuilder` with `marked`.

WaDelma:
This is breaking change indeed (as witnessed tests failing because they don't import the trait). I am not against doing one for this as it seems that it would be useful.

It probably needs the aforementioned `MarkedBuilder` which should be extension trait to the `Builder` trait.

Xaeroxe:
Also we need to add `EntityResBuilder` to this, which was merged yesterday.

GraemeWilde:
`EntityResBuilder` has an extra parameter in it's `with` (a storage), not sure what to there.

Xaeroxe:
@GraemeWilde ah yeah.  Ok, just ignore it then.

GraemeWilde:
I also updated the changelog and the specs version to 0.12.0 (I think thats what will need to happen?).

src/world/mod.rs
```diff
@@ -76,15 +90,15 @@ pub struct EntityBuilder<'a> {
...
 }
 
-impl<'a> EntityBuilder<'a> {
+impl<'a> Builder for EntityBuilder<'a> {
     /// Appends a component and associates it with the entity.
     ///
     /// # Panics
     ///
     /// Panics if the component hasn't been `register()`ed in the
     /// `World`.
```
torkleyy: No need for the duplicated docs here.

src/saveload/marker.rs
```diff
@@ -35,7 +61,7 @@ impl<'a> EntityBuilder<'a> {
     /// ## Panics
     ///
     /// Panics in case there's no allocator added to the `World`.
-    pub fn marked<M>(self) -> Self
+    fn marked<M>(self) -> Self
```
torkleyy: This still has duplicated docs

src/world/mod.rs
```diff
@@ -97,7 +105,7 @@ impl<'a> EntityBuilder<'a> {
 
     /// Finishes the building and returns the entity.
```
torkleyy: "As opposed to `LazyBuilder`, the components are available immediately."

src/world/mod.rs
```diff
@@ -34,6 +34,20 @@ impl<'a> Iterator for CreateIter<'a> {
     }
 }
 
+/// A common trait for EntityBuilder and LazyBuilder, allowing either to be used.
+pub trait Builder {
```
torkleyy: Might want to point out the guarantees here:

* entity is definitely alive
* components may or may not exist before a `maintain`

src/saveload/marker.rs
GraemeWilde: Should I be removing the docs from the trait or the implementation? Do the docs from traits fall through to the implementations?

torkleyy: Docs on the implementation of a trait only make sense if there's something special about the implementation. So for the lazy stuff it does make sense to document the delay of the insertions.

GraemeWilde:
Is it okay if I make all the doc mentions of `stuff` into links?

torkleyy:
I used to do that but I don't anymore because the links don't work with
reexports.

GraemeWilde:
Are you talking about when you `pub use` re-export something from a different crate?

torkleyy:
@GraemeWilde Yes, but it doesn't even have to be in a different crate. If we reexport it in e.g. the prelude of Specs, the link will break.

src/world/mod.rs
```diff
@@ -34,14 +34,16 @@ impl<'a> Iterator for CreateIter<'a> {
...
+/// Entity is definitely alive, but the components may or may not exist before a call to
+/// [`World::maintain`].
 pub trait Builder {
     /// Appends a component and associates it with the entity.
     ///
     /// # Panics
     ///
-    /// Panics if the component hasn't been `register()`ed in the
-    /// `World`.
+    /// Panics if the component hasn't been [`register()`](World::register)ed in the
```
torkleyy: Do these links really work? I never saw them like this.

GraemeWilde: Hmm. They work for nightly, but not stable apparently. The ``[`Thing`]`` ones without specifying the link later don't work without nightly either,. The re-export ones don't work (as you were saying) without `#[doc(no-inline)]`, but using that puts it in the re-exports section with no description...

If you would prefer, I'll just take out all the links.

PR #434, opened 2018-05-31
Saveload convenience (proc-)macros

LinearZoetrope:
This adds two major features:

1. A custom derive for `IntoSerialize` and `FromDeserialize` named `#[derive(Saveload)]`. It may seem odd to derive them together, but after spending significant time trying to do it the normal way (one derive per trait) it became clear that due to the requirement of "proxy" types that fill the `IntoSerialize::Data` and `FromDeserialize::Data` fields, it made sense to do it this way or you start getting into attribute hell.
2. A macro which automatically will create `SystemData` for the most common usecase of `SerializeComponents` and `DeserializeComponents`. Unlike the existing tuple auto-implementation, this generates arbitrarily large structs that you can readily plop straight into your serialization or deserialization system. It also doesn't use recursion, so it shouldn't hit any nasty macro expansion recursion limits.

I recommend reading the module comment for the `Saveload` derive, the doc comment for `saveload_components!`, and the new documentation all around for a greater overview of what this all looks like in the end. This makes using the `saveload` features much more pleasant and less stuttery. In the case of `saveload_components!` especially it lets people easily keep their Serialization and Deserialization component lists in sync which can be hugely important and has been a source of a lot of frustrating bugs for me in the past.

The current `derive` does have a major downside in that I couldn't figure out how to get it to work with generic structs, but due to the way specs works I don't find myself holding a `Foo<T>` too often anyway.

---
This change is [Reviewable](https://reviewable.io/reviews/slide-rs/specs/434)

### DIFF
--- a/specs-derive/src/impl_saveload.rs
+++ b/specs-derive/src/impl_saveload.rs
@@ -1,0 +1,496 @@
+//! Contains implementations for #[derive(Saveload)]
+//! which derives both `IntoSerialize` and `FromDeserialize`.
+//!
+//! Since this requires a proxy "Data" `struct`/`enum`, these two need
+//! to be derived together or it begins requiring unwieldy attribute explosion.
+//!
+//! Currently there is one major limitation to this implementation: it cannot handle
+//! generic types such as `struct Foo<T> { x: T, e: Entity }`. I'm not good enough at
+//! using `syn` to iron this out, but since specs requires explicit references to non-generic
+//! components in system data, it can generally be worked around by newtyping like
+//! `struct TypedFoo(Foo<u32>);` and the like.
+
+// NOTE: All examples given in the documentation below are "cleaned up" into readable Rust,
+// so it doesn't give an entirely accurate view of what's actually generated (for instance,
+// I'll write `Entity` instead of `::specs::Entity` which is actually what's generated).
+
+use quote::Tokens;
+use syn::{DataEnum, DataStruct, DeriveInput, Field, Generics, Ident, Type, TypeParamBound};
+
+// Handy collection since tuples got unwieldy and
+// unclear in purpose
+struct SaveloadDerive {
+    type_def: Tokens,
+    ser: Tokens,
+    de: Tokens,
+    saveload_name: Ident,
+    saveload_generics: Generics,
+}
+
+/// The main entrypoint, sets things up and delegates to the
+/// type we're deriving on.
+pub fn impl_saveload(ast: &mut DeriveInput) -> Tokens {
+    use syn::Data;
+
+    if !ast.generics.params.is_empty() {
+        panic!("Deriving `Saveload` does not, at the moment, support generic types (i.e. `struct Foo<T>`).
+        
+        This limitation may be fixed in the future, but for now you can usually work around this by newtyping
+        your generic data structure to a concrete-variant (e.g. `struct FooU32(Foo<u32>);`), or implementing `IntoSerialize` and `FromDeserialize` 
+        yourself, which is relatively straightforward if tedious (see the documentation for those traits).")
+    }
+
+    add_bound(
+        &mut ast.generics,
+        parse_quote!(::specs::saveload::IntoSerialize),
+    );
+    add_bound(
+        &mut ast.generics,
+        parse_quote!(::specs::saveload::FromDeserialize),
+    );
+    add_bound(&mut ast.generics, parse_quote!(Clone));
+
+    let derive = match ast.data {
+        Data::Struct(ref mut data) => saveload_struct(data, &mut ast.ident, &mut ast.generics),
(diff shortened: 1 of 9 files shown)


specs-derive/src/impl_saveload.rs
```diff
@@ -0,0 +1,496 @@
...
+        Type::TraitObject(_) => {}
+        Type::ImplTrait(_) => {}
+        Type::Slice(_) => {
+            panic!("Slices are unsupported, use owned types like Vecs or Arrays instead")
+        }
+        Type::Reference(_) => panic!("References are unsupported"),
+        Type::Ptr(_) => panic!("Raw pointer types are unsupported"),
+        Type::BareFn(_) => panic!("Function types are unsupported"),
+        /* We're in a struct so it doesn't matter */
+        Type::Never(_) => unreachable!(),
```
WaDelma: When never type is stable this can happen: You could have field with type `!` (Extremely unlikely with struct, more likely with enum. Even more unlikely without generics support).

specs-derive/src/impl_saveload.rs
```diff
@@ -0,0 +1,496 @@
...
+//! which derives both `IntoSerialize` and `FromDeserialize`.
+//!
+//! Since this requires a proxy "Data" `struct`/`enum`, these two need
+//! to be derived together or it begins requiring unwieldy attribute explosion.
+//!
+//! Currently there is one major limitation to this implementation: it cannot handle
+//! generic types such as `struct Foo<T> { x: T, e: Entity }`. I'm not good enough at
+//! using `syn` to iron this out, but since specs requires explicit references to non-generic
+//! components in system data, it can generally be worked around by newtyping like
+//! `struct TypedFoo(Foo<u32>);` and the like.
```
WaDelma: This is how I handled generics on the derive target: https://github.com/WaDelma/bob/blob/master/bob/src/lib.rs#L400
The helper function `add_ty_params` is at: https://github.com/WaDelma/bob/blob/master/bob/src/lib.rs#L844

The code is bit messy and I wouldn't write the helper like that again, but it should give some idea how to do it.

LinearZoetrope:
Hmm, the visibility modifiers are being a pain. If they're `pub` they want all components to be `pub`, but if they're `pub(crate)` or `pub(super)` they complain endlessly about leaking private types.

LinearZoetrope:
Crap, I forgot code like

`impl IntoSerialize for C where C: Serialize+Clone` vs `impl IntoSerialize for Vec<T> where T: IntoSerialize` would require specialization to resolve (and even though IDK if specialization would solve this).

Not sure how to solve this from a design standpoint, seems like there are only really two options:

1. Hand-implement `IntoSerialize` and `FromDeserialize` for every native type to avoid a blanket impl, and expect others to `#[derive(Saveload)]` for their own types.
2. Extend the derive to have special casing for things like `Vec<T>` and generate specialized code

I feel like 1 is more painful but probably ultimately more flexible, it seems closer to how `serde` works, but is also a massive undertaking (not in difficulty so much in time and scale).

torkleyy:
Phew, sounds complicated. I would love to add these macros, but I don't know what to do about this without using more time on this PR. I think I'll need about two weeks before I can take a closer look, @Xaeroxe @WaDelma feel free to decide on your own or wait, whatever you prefer.

specs-derive/src/impl_saveload.rs
LinearZoetrope: Ah yes, I think I didn't understand the never type. It's meant to be the "official" solution to things like `specs` `NoError` type is now, right? Like a `Result<Foo,!>` is a type that can exist, but its `Err` can never exist at runtime, as guaranteed by its type semantics. And also has the benefit of satisfying every trait because it's a bottom type, so you can use it as shorthand for "don't worry about this case, if we get here we've already panicked anyway."

I got confused because I know `panic!` and its cousins are of type `!` now and assumed it was solely the return type of those devices, I wasn't aware there was a communication to build that into the type system as a feature, neat!

WaDelma: Yeah exactly that!

torkleyy:
@LinearZoetrope Are you still interested in this PR?

Also pinging @WaDelma @Xaeroxe @jojolepro @Moxinilian @omni-viral in case any of you need this feature and would like to provide a solution.

LinearZoetrope:
Yes I am. Sorry, I had to switch tracks for work but I can poke at this some more soon, especially if anyone else has any ideas about how to handle some of the issues.

jojolepro:
While I do need this feature, I don't think I will be of much help. I'm not advanced enough in macro and generics yet to fully understand how to fix the implementation.

WaDelma:
I can look and see if I can improve it.

WaDelma:
I think it would be good idea to split the custom derive and the macro to separate prs.
I feel like there are too many moving parts as is.

WaDelma:
Closing in favor to #460 and #461

Issue #436, opened 2018-06-02
Use NonZeroU32 for Generation


andrewhickman:
As far as I can tell, there is no way for the user to get an entity with a generation of zero. Thus when `nonzero` is stabilised we can allow a more efficient representation of `Option<Entity>` by representing generation as `Generation(NonZeroU32)`. Unfortunately `NonZeroI32` has been removed in the current nightly or we could use that instead.

I've managed to get this working on [my branch](https://github.com/andrewhickman/specs/tree/non-zero-gen), I'll put in a pull request when this stuff stabilises.

torkleyy:
Ah, nice idea, thanks for submitting! Looking forward to your PR :)

Aceeri:
@Xaeroxe should close this now

PR #447, opened 2018-08-04
Use non-zero optimization for Generation

andrewhickman:
As promised in issue #436, this PR uses `NonZeroU32` in `Generation`. Note this bumps the minimum supported rust version to 1.28 - is this a concern? If so I could put this under a feature flag.

---
This change is [Reviewable](https://reviewable.io/reviews/slide-rs/specs/447)

### DIFF
--- a/src/world/entity.rs
+++ b/src/world/entity.rs
@@ -1,3 +1,4 @@
+use std::fmt;
 use std::num::NonZeroU32;
 use std::sync::atomic::{AtomicUsize, Ordering};
 
@@ -363,8 +364,15 @@ impl<'a> Drop for EntityResBuilder<'a> {
 /// Index generation. When a new entity is placed at an old index,
 /// it bumps the `Generation` by 1. This allows to avoid using components
 /// from the entities that were deleted.
-#[derive(Clone, Copy, Debug, Hash, Eq, Ord, PartialEq, PartialOrd)]
+#[derive(Clone, Copy, Hash, Eq, Ord, PartialEq, PartialOrd)]
 pub struct Generation(NonZeroU32);
+
+// Show the inner value as i32 instead of u32.
+impl fmt::Debug for Generation {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        f.debug_tuple("Generation").field(&self.id()).finish()
+    }
+}
 
 impl Generation {
     pub(crate) const ONE: Generation = Generation(unsafe { NonZeroU32::new_unchecked(1) });
@@ -416,7 +424,8 @@ impl ZeroableGeneration {
     /// Panics if it is alive.
     fn raised(&self) -> Generation {
         assert!(!self.is_alive());
-        Generation(unsafe { NonZeroU32::new_unchecked((1 - self.id()) as u32) })
+        let gen = 1i32.checked_sub(self.id()).expect("generation overflow");
+        Generation(unsafe { NonZeroU32::new_unchecked(gen as u32) })
     }
 
     /// Revives and increments a dead `ZeroableGeneration`.
--- a/src/world/entity.rs
+++ b/src/world/entity.rs
@@ -139,7 +139,6 @@ impl Allocator {
         self.raised.add_atomic(id);
         let gen = self
             .generation(id)
-            .map(|gen| if gen.is_alive() { gen } else { gen.raised() })
             .unwrap_or(Generation::ONE);
         Entity(id, gen)
     }
@@ -320,7 +319,6 @@ impl<'a> Join for &'a EntitiesRes {
         let gen = v
             .alloc
             .generation(idx)
-            .map(|gen| if gen.is_alive() { gen } else { gen.raised() })
             .unwrap_or(Generation::ONE);
         Entity(idx, gen)
     }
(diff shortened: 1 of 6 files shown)


torkleyy:
Hi! Unfortunately, this is a nightly-only feature, we cannot use this optimization.

Aceeri:
@torkleyy @andrewhickman 1.29 is released now, might be worth looking at merging this (unless we want a larger buffer size).

Xaeroxe:
Should probably be released in 0.13, I'm doing a 0.12.3 right now

src/world/entity.rs
```diff
@@ -383,20 +406,38 @@ impl Generation {
     /// Panics in debug mode if it's not alive.
     fn die(&mut self) {
         debug_assert!(self.is_alive());
-        self.0 = -self.0;
+        self.0 = NonZeroU32::new(-self.id() as u32).map(Generation);
```
pthariensflame: Can't this underflow and therefore panic in debug builds?

Xaeroxe: Sure can, so yeah that should be corrected.

andrewhickman: Which operation can overflow?

andrewhickman: The assert should ensure that `id()` is positive and the only overflow case is where `id()` is i32::MIN so I think this one is correct. AFAIK signed-ness casts are always no-ops in rust so that should be OK as well

src/world/entity.rs
```diff
@@ -383,20 +406,38 @@ impl Generation {
...
     /// # Panics
     ///
-    /// Panics in debug mode if it is alive.
-    fn raised(self) -> Generation {
-        debug_assert!(!self.is_alive());
-        Generation(1 - self.0)
+    /// Panics if it is alive.
+    fn raised(&self) -> Generation {
+        assert!(!self.is_alive());
+        Generation(unsafe { NonZeroU32::new_unchecked((1 - self.id()) as u32) })
```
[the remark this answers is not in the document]
andrewhickman: Note that wrapping will give the wrong behaviour here, sending id from `i32::MIN + 1` to `i32::MIN` and vice versa. I think it would be better to just unconditionally panic here. Potentially we could start wrapping the generations back around to 1 but that sounds risky to me.

andrewhickman:
The test failure seems to caused by ab43ce19b0d334b893a99e33102d50c31f8a4678

Xaeroxe:
I removed `is_alive()` and `raised()` from the generation because if a `Generation` can no longer be less than 0 then it's not possible for that structure to register as "dead".  Since a structure can never be dead what's the point of "check if this is alive and if so return it".  Perhaps I missed something but is that not what a `NonZeroU32` does?

src/world/entity.rs
pthariensflame: Ok; I can accept the `assert!` making the negation safe, but I'd like to see where casts from signed to unsigned are supposed to be defined as bitcasts.

pthariensflame: Never mind; found it: https://doc.rust-lang.org/reference/expressions/operator-expr.html#semantics

Xaeroxe: `id()` can only ever be a positive number by virtue of the fact that it is built from a `NonZeroU32`.  Won't this always panic?

Xaeroxe: Hm ok but it can only ever be positive or 0.  I think we still have a problem.

Xaeroxe:
I don't know, I'm not even sure it's worth the effort to be honest.  Specs is hard enough to read as is, and while this is technologically an improvement relying on overflow behavior to implement negatives makes it so much harder to maintain that I question the value of this.

Xaeroxe:
Ok, so I've been thinking about this a bit more and I actually see a way we can make this work.  You've already made sure that the only entry and exit points from the `Generation` structure output to i32 correctly, the next thing I feel this really needs to be perfect is that everywhere we're casting to a `NonZeroU32` or to a `i32` we should provide a comment explaining the overflow/underflow behavior.  This way we still have a `Generation` value with a nice interface, and specs maintainers understand what's going on.

andrewhickman:
@Xaeroxe not sure from your comments if you worked this out but the only reason I used `NonZeroU32` is that `NonZeroI32` was never stabilised. So the inner value can be negative, but the bits are stored as u32. I'll add a comment explaining this and fix this PR up tomorrow.

Xaeroxe:
Yeah I learned that from the original issue... probably should have read that first :P Sorry for all the trouble, and thanks for fixing my mistakes

pthariensflame:
If we wanted to go a little overboard, we could implement a `NonZeroI32` on stable as a `#[repr(transparent)]` wrapper around `NonZeroU32`, delegating methods wherever possible.

Xaeroxe:
Maybe we ought to make a crate ;)

Xaeroxe:
@pthariensflame K you're gonna laugh but...

I did it!  https://crates.io/crates/nonzero_signed

andrewhickman:
@Xaeroxe ooh nice! Shall we use it here?

Xaeroxe:
Might as well :)

On Sep 21, 2018 14:30, "Andrew Hickman" <notifications@github.com> wrote:

@Xaeroxe <https://github.com/Xaeroxe> ooh nice! Shall we use it here?

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/slide-rs/specs/pull/447#issuecomment-423661751>, or mute
the thread

Xaeroxe:
Thank you!

Issue #440, opened 2018-06-27
delete_all panics with some lazy update


thiolliere:
```rust
#[test]
fn delete_and_lazy() {
    let mut world = World::new();
    {
        let lazy_update = world.write_resource::<::specs::LazyUpdate>();
        lazy_update.exec(|world| {
            world.entities().create();
        })
    }
    world.maintain();
    {
        let lazy_update = world.write_resource::<::specs::LazyUpdate>();
        lazy_update.exec(|world| {
            world.entities().create();
        })
    }
    world.delete_all();
}
```

```
thread 'delete_and_lazy' panicked at 'index out of bounds: the len is 0 but the index is 0', /checkout/src/libcore/slice/mod.rs:2342:14
```

Aceeri:
Simplest fix for this is just adding a `world.maintain()` to the `world.delete_all()` method I believe.

PR #454, opened 2018-08-20
Check length of generations in allocator before killing

Aceeri:
Probably fixes https://github.com/slide-rs/specs/issues/440 but would probably be good to double check there isn't anything wrong here.

From issue on my reasoning why I think this should be enough:
> Since atomically created entities don't have a generation in the Allocator until merge  is called, so if you atomically create an entity and it is the first of  its generation, then it will be out of range and panic. And since we  remove the entity from raised a line before, then it should never be allocated in the future and should be safe to just ignore that entity on kill.

### DIFF
--- a/src/world/entity.rs
+++ b/src/world/entity.rs
@@ -66,7 +66,10 @@ impl Allocator {
 
             self.alive.remove(entity.id());
             self.raised.remove(entity.id());
-            self.generations[id].die();
+            if self.generations.len() > id {
+                self.generations[id].die();
+            }
+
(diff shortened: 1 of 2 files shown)


src/world/entity.rs
```diff
@@ -66,7 +66,10 @@ impl Allocator {
 
             self.alive.remove(entity.id());
             self.raised.remove(entity.id());
-            self.generations[id].die();
+            if self.generations.len() > id {
```
torkleyy: Please add a comment here.

Issue #358, opened 2018-03-02
Allow joining over possibly-existing components


LinearZoetrope:
Brought this up on the Gitter:

It would be nice to allow joining over components that **possibly** exist. You can do this now with `Entities` and `get`/`get_mut`/`entry`, but it buries the true input/output story of a given loop deep inside its code (or possibly other functions in it). I think allowing this makes it more clear what a piece of code wants to touch to allow all or most of the components a loop wants to modify in the loop header itself.

This could take multiple forms.

Idea 1,
-----
Allow a `maybe`/`maybe_mut` that returns an `Option<&T>`/`Option<&mut T>`:

```rust
for (foo,bar) in (&storage1, storage2.maybe()).join() {
    if let Some(bar) = bar {
        // do whatever
    } else {
        // do something else
    }
}
```

Idea 2 
------
Allow a `maybe`/`entries` that returns an `Option<&T>`/`Entry<&mut T>`

```rust
for (foo,bar) in (&storage1, storage2.entries()) {
    if foo == 5 {
        let bar = bar.entry().or_insert(Default::default());
    }
}
```

`Entries`, naturally, only works on `write_storage`

Idea 3
-------
Similar to idea 2, but join over `or_insert`

```rust
for (foo,bar) in (&storage1, storage2.or_insert(Default::default())) {
    
}
```

This has the benefit that for write storage outputs, everything is always `&mut T`, but it uglifies the loop header.

Overall
--------

I think my favorite is Option 2 (though there's no reason we couldn't combine 1 and 2), it gives you more control over when to insert a component rather than forcing you to insert if the component is touched at all.

Obviously for the entry system, in order for this to work with `par_join`, it would need to require an atomic `maintain`-like operation.

torkleyy:
Wow, this seems like a much wanted feature. I have a couple of higher priority things to do, but I'll try to get back to this issue in about 2 - 3 weeks. If somebody else wants to take this, I'd love to take a PR, too.

The idea with the entries (number 2) sounds nice, we could use those for retrieving multiple components mutably (#336), too.

torkleyy:
That sounds kinda complicated as we would need an enum then. And as soon as you `|` on three or more component that would get rather ugly, unless you know a good solution?

GraemeWilde:
[...]

PR #455, opened 2018-08-21
MaybeJoin for possibly existing components

Aceeri:
Related issue: https://github.com/slide-rs/specs/issues/358

Adds a generic `MaybeJoin` to all `Join` structures to optionally get a component.

I think we might want to add an `BitSetAll` BitSetLike to hibitset for a more specialized `BitSetNot(BitSet::new())` since I don't like the idea of allocation of some vecs on a join that won't actually be used. Ideally just have a BitSetLike that returns the max usize on all layers.

Dependent on https://github.com/slide-rs/hibitset/pull/36 for optimization (not sure if rust is smart enough to remove the unnecessary Vec allocations.)

---
This change is [Reviewable](https://reviewable.io/reviews/slide-rs/specs/455)

### DIFF
--- a/src/join.rs
+++ b/src/join.rs
@@ -3,9 +3,9 @@
 use std;
 use std::cell::UnsafeCell;
 
-use hibitset::{BitIter, BitProducer, BitSetAnd, BitSetLike};
-use rayon::iter::ParallelIterator;
+use hibitset::{BitIter, BitProducer, BitSet, BitSetAnd, BitSetLike, BitSetNot};
 use rayon::iter::plumbing::{bridge_unindexed, Folder, UnindexedConsumer, UnindexedProducer};
+use rayon::iter::ParallelIterator;
 use tuple_utils::Split;
 
 use world::{Entities, Entity, Index};
@@ -155,6 +155,13 @@ pub trait Join {
         JoinIter::new(self)
     }
 
+    fn maybe(self) -> MaybeJoin<Self>
+    where
+        Self: Sized,
+    {
+        MaybeJoin(self)
+    }
+
     /// Open this join by returning the mask and the storages.
     ///
     /// This is unsafe because implementations of this trait can permit
@@ -177,6 +184,28 @@ pub unsafe trait ParJoin: Join {
         Self: Sized,
     {
         JoinParIter(self)
+    }
+}
+
+pub struct MaybeJoin<J: Join>(pub J);
+
+impl<T> Join for MaybeJoin<T>
+where
+    T: Join,
+{
+    type Type = Option<<T as Join>::Type>;
+    type Value = (<T as Join>::Mask, <T as Join>::Value);
+    type Mask = BitSetNot<BitSet>;
+    unsafe fn open(self) -> (Self::Mask, Self::Value) {
+        let (mask, value) = self.0.open();
+        (BitSetNot(BitSet::new()), (mask, value))
+    }
+    unsafe fn get((mask, value): &mut Self::Value, id: Index) -> Self::Type {
+        if mask.contains(id) {
+            Some(<T as Join>::get(value, id))
+        } else {
+            None
+        }
     }
 }
 
@@ -336,8 +365,7 @@ where
     J::Type: Send,
     J::Value: 'a + Send,
     J::Mask: 'a + Send + Sync,
-{
-}
+{}
 
 impl<'a, J> UnindexedProducer for JoinProducer<'a, J>
 where
--- a/src/join.rs
+++ b/src/join.rs
@@ -3,7 +3,7 @@
 use std;
 use std::cell::UnsafeCell;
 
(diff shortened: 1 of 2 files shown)


Xaeroxe:
> (not sure if rust is smart enough to remove the unnecessary Vec allocations.)

By the way, it is.

torkleyy:
I don't think so Jacob. You might be mixing this with `Vec::new` not
allocating before pushing.

Xaeroxe:
@torkleyy the new constructor for the bitset doesn't push anything, therefore no heap allocations are made.  However an empty bitset does certainly occupy space, just not also a separate heap allocation.

I think we believe the same thing I just didn't quite communicate that adequately.

src/join.rs
```diff
@@ -155,6 +155,13 @@ pub trait Join {
         JoinIter::new(self)
     }
 
+    fn maybe(self) -> MaybeJoin<Self>
```
Xaeroxe: Can we document this method?

torkleyy:
Oh, I confused this with the atomic bit set, which does allocate. Sorry!

src/join.rs
Xaeroxe: We should also provide a usage example.

Xaeroxe:
Sorry I didn't see the failing Travis.  I take it we need a hibitset release first.

torkleyy:
What's the status here @Xaeroxe @Aceeri ?

Aceeri:
@torkleyy I think just waiting on a release of hibitset, I could try to do it but just forgot about this PR earlier.

torkleyy:
I'll publish it for you, hold on.

torkleyy:
Done, you can bump the version now.

src/join.rs
```diff
@@ -155,6 +155,61 @@ pub trait Join {
         JoinIter::new(self)
     }
 
+    /// Optionally gets a `Join` if it exists.
```
torkleyy: This sentence sounds confusing to me. Suggesting: `Returns a \`Join\`able` that yields for all indices, returning `None` for all missing elements.`

src/join.rs
```diff
@@ -155,6 +155,61 @@ pub trait Join {
...
+    ///         .build();
+    ///
+    ///     dispatcher.dispatch(&mut world.res);
+    ///
+    ///     let positions = world.read_storage::<Pos>();
+    ///     assert_eq!(positions.get(e1), Some(&Pos { x: 5, y: 2 }));
+    ///     assert_eq!(positions.get(e2), Some(&Pos { x: 0, y: 0 }));
+    /// }
+    /// ```
+    fn maybe(self) -> MaybeJoin<Self>
```
torkleyy: Iterating only over a `MaybeJoin` would iterate over 4,294,96,7296 `None` values. Don't know how to fix it, but this PR doesn't qualify for a merge before it's solved.

Xaeroxe: I'm not sure it needs to be solved to be honest.  This would only happen if every element in a join tuple was a MaybeJoin, in which case we should probably just tell people "don't do that."

Xaeroxe: As a slightly more user friendly alternative, we could store a variable in the `MaybeJoin` iterator called "count" and if count goes to max print a warning: "MaybeJoin iterated over all indices.  Do you have a join iterator consisting exclusively of MaybeJoin?  Consider refactoring this code to avoid performance problems."

Xaeroxe: The only other alternative to this I can think of is to require `MaybeJoin` to store a `EntitiesRes` reference somehow so that it will only iterate over valid indices.

Xaeroxe: @torkleyy Do you find any of my ideas here satisfactory?

torkleyy: I'm not quite sure. Let's use a warning in the docs for now.

Aceeri: A quicker way (and more completely covering way) is to just add 
```
if idx as usize == MAX_EID - 1 {
    println!("WARNING: Iterated through all indices ({} indices), you might've made a join with all `MaybeJoin`s, which is unbounded in length", MAX_EID);
}
```
But that would have a more significant cost.

src/join.rs
```diff
@@ -165,6 +226,13 @@ pub trait Join {
 
     /// Get a joined component value by a given index.
     unsafe fn get(value: &mut Self::Value, id: Index) -> Self::Type;
+
+    /// If this `Join` typically returns all indices in the mask, then iterating over only it
+    /// or combined with other joins that are also dangerous will cause the `JoinIter`/`ParJoin` to
+    /// go through all indices which is usually not what is wanted and will kill performance.
+    fn dangerous_iter() -> bool {
```
torkleyy: Could you do a rename to `is_unconstrained`?

torkleyy: Ah, can also add an inline attrib here.

PR #457, opened 2018-09-06
Reduced the complexity of the entity allocator

Flecheck:
This pull request uses a vector to store deleted entities to reduce the complexity of entity allocation to O(1)*.
The maximum memory footprint is 2x the maximum current amount but it can only happen in the worst case and is I believe acceptable for such a use case.

The current approach has the pitfall of often reusing the same entities, hence potentially making their generation quite high. Introducing a VecDequeue instead of the current Vec for the EntityCache could potentially solve this problem (with a cost in memory), so if you believe the change would be worth it I can add it relatively easily. However, this might be unnecessary as 2³¹-1 is the maximum generation.

---
This change is [Reviewable](https://reviewable.io/reviews/slide-rs/specs/457)

### DIFF
--- a/src/world/entity.rs
+++ b/src/world/entity.rs
@@ -51,7 +51,8 @@ pub(crate) struct Allocator {
     alive: BitSet,
     raised: AtomicBitSet,
     killed: AtomicBitSet,
-    start_from: AtomicUsize,
+    cache: EntityCache,
+    max_id: AtomicUsize,
 }
 
 impl Allocator {
@@ -65,20 +66,16 @@ impl Allocator {
             }
 
             self.alive.remove(entity.id());
-            self.raised.remove(entity.id());
 
-            // Since atomically created entities don't have a generation until merge
-            // is called, the first entity of a generation won't actually have a generation
-            // in the allocator. And it is fine for us to just ignore the entity since
-            // we clear `self.raised` before so it won't get allocated later.
-            if self.generations.len() > id {
-                self.generations[id].die();
-            }
+            self.update_generation_length(id);
 
-            if id < self.start_from.load(Ordering::Relaxed) {
-                self.start_from.store(id, Ordering::Relaxed);
+            if self.raised.remove(entity.id()) {
+                self.generations[id] = self.generations[id].raised();
             }
+            self.generations[id].die();
         }
+
+        self.cache.extend(delete.iter().map(|e| e.0));
 
         Ok(())
     }
@@ -122,66 +119,36 @@ impl Allocator {
         Entity(id, gen)
     }
 
-    /// Attempt to move the `start_from` value
-    pub fn update_start_from(&self, start_from: usize) {
-        loop {
-            let current = self.start_from.load(Ordering::Relaxed);
-
-            // if the current value is bigger then ours, we bail
-            if current >= start_from {
-                return;
-            }
-
-            if start_from
-                == self.start_from
-                    .compare_and_swap(current, start_from, Ordering::Relaxed)
-            {
-                return;
-            }
-        }
-    }
-
     /// Allocate a new entity
     pub fn allocate_atomic(&self) -> Entity {
-        let idx = self.start_from.load(Ordering::Relaxed);
-        for i in idx.. {
-            if !self.alive.contains(i as Index) && !self.raised.add_atomic(i as Index) {
-                self.update_start_from(i + 1);
+        let id = self.cache.pop_atomic().unwrap_or_else(|| {
+            atomic_increment(&self.max_id).expect("No entity left to allocate") as Index
+        });
 
-                let gen = self.generations
-                    .get(i as usize)
-                    .map(|&gen| {
-                        if gen.is_alive() {
-                            gen
-                        } else {
-                            gen.raised()
-                        }
-                    })
-                    .unwrap_or(Generation(1));
-
-                return Entity(i as Index, gen);
-            }
-        }
-        panic!("No entities left to allocate")
+        self.raised.add_atomic(id);
+        let gen = self
+            .generations
+            .get(id as usize)
+            .map(|&gen| if gen.is_alive() { gen } else { gen.raised() })
+            .unwrap_or(Generation(1));
+        Entity(id, gen)
     }
 
     /// Allocate a new entity
     pub fn allocate(&mut self) -> Entity {
-        let idx = self.start_from.load(Ordering::Relaxed);
-        for i in idx.. {
-            if !self.raised.contains(i as Index) && !self.alive.add(i as Index) {
-                // this is safe since we have mutable access to everything!
-                self.start_from.store(i + 1, Ordering::Relaxed);
+        let id = self.cache.pop().unwrap_or_else(|| {
+            let id = *self.max_id.get_mut();
+            *self.max_id.get_mut() = id.checked_add(1).expect("No entity left to allocate");
+            id as Index
+        });
 
-                while self.generations.len() <= i as usize {
-                    self.generations.push(Generation(0));
-                }
-                self.generations[i as usize] = self.generations[i as usize].raised();
+        self.update_generation_length(id as usize);
 
-                return Entity(i as Index, self.generations[i as usize]);
-            }
-        }
-        panic!("No entities left to allocate")
+        self.alive.add(id as Index);
+
+        self.generations[id as usize] = self.generations[id as usize].raised();
+
+        Entity(id as Index, self.generations[id as usize])
     }
 
     /// Maintains the allocated entities, mainly dealing with atomically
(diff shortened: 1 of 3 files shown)


OvermindDL1:
I've always used a 32-bit index and a 32-bit generation in my engine and I never assumed that the generation was linear either, if I had a rollover I never had issues from it.  I should only have had issues if there was for some reason an entity handle that survived with one generation ID without being touched (thus 0'd out due to it being invalid) for an entire 2^32 count of generations.  Even just a simple system that touched 100k entities stored in 'something' per update just to ensure they were valid would be simple enough to prevent the issue, but as stated, I'm not sure I've ever seen it in 15 years of my C++ engine?

For note, my 'deleted' entities were stored in a bag, essentially just (in C++) an std::vector where I just appended to the end and popped a new one off the end, so yes very few tended to be reused very rapidly, but that could be decently fixed by a circular queue with little extra cost, but even them I'm not sure the little extra cost is worth it?

So as this PR seems to match how my engine does it, I'm +1 on it.  :-)

torkleyy:
please ping me in 4 hours

src/world/entity.rs
```diff
@@ -432,3 +409,50 @@ impl Generation {
...
+
+impl Extend<Index> for EntityCache {
+    fn extend<T: IntoIterator<Item = Index>>(&mut self, iter: T) {
+        self.maintain();
+        self.cache.extend(iter);
+        self.len.store(self.cache.len(), Ordering::Relaxed);
+    }
+}
+
+fn atomic_add1(i: &AtomicUsize) -> Option<usize> {
```
torkleyy: Can use `fetch_add` instead.

Flecheck: The main issue with `fetch_add` is that it will wrap on overflow, whereas this will return an option preventing weird bugs.

OvermindDL1: I personally exploited overflow on my generations (assuming any even reached that high) to wrap back around.  For generations I just wanted 'enough' entropy to minimize the chance of a re-use collision with a stored ID untouched somewhere.

Now entity ID index's should not overflow of course, but if that were hit I'd think I'd have some memory issues because of the component load in 99.99999% of cases...

Flecheck: It is to prevent overflow on the entity indexes, which could allocate the first entities multiple times.

OvermindDL1: Yeah if that one is for entity indexes then it definitely shouldn't overflow, that should be checked for.

src/world/entity.rs
```diff
@@ -122,66 +121,41 @@ impl Allocator {
...
-
-                while self.generations.len() <= i as usize {
-                    self.generations.push(Generation(0));
-                }
-                self.generations[i as usize] = self.generations[i as usize].raised();
-
-                return Entity(i as Index, self.generations[i as usize]);
+        let id = self.cache.pop_atomic().unwrap_or_else(|| {
+            let id = self.max_id.load(Ordering::Relaxed);
+            if id as Index == <Index>::max_value() {
```
torkleyy: No need for UFCS here.

src/world/entity.rs
```diff
@@ -122,66 +121,41 @@ impl Allocator {
...
-                self.start_from.store(i + 1, Ordering::Relaxed);
-
-                while self.generations.len() <= i as usize {
-                    self.generations.push(Generation(0));
-                }
-                self.generations[i as usize] = self.generations[i as usize].raised();
-
-                return Entity(i as Index, self.generations[i as usize]);
+        let id = self.cache.pop_atomic().unwrap_or_else(|| {
+            let id = self.max_id.load(Ordering::Relaxed);
```
torkleyy: No need for an atomic operation (please use `get_mut`).

src/world/entity.rs
```diff
@@ -122,66 +121,41 @@ impl Allocator {
...
+            if id as Index == <Index>::max_value() {
+                panic!("No entity left to allocate");
             }
+            self.max_id.store(id as usize + 1, Ordering::Relaxed);
```
torkleyy: Same as above, no need for atomic operations.

src/world/entity.rs
```diff
@@ -432,3 +397,50 @@ impl Generation {
...
+            ) {
+                Ok(x) => return Some(self.cache[x - 1]),
+                Err(next_prev) => prev = next_prev,
+            }
+        }
+        None
+    }
+
+    fn maintain(&mut self) {
+        self.cache.truncate(self.len.load(Ordering::Relaxed));
```
torkleyy: Please use `get_mut`. An atomic operation is even confusing here.

src/world/entity.rs
```diff
@@ -432,3 +397,50 @@ impl Generation {
...
+    }
+}
+
+impl Extend<Index> for EntityCache {
+    fn extend<T: IntoIterator<Item = Index>>(&mut self, iter: T) {
+        self.maintain();
+        self.cache.extend(iter);
+        self.len.store(self.cache.len(), Ordering::Relaxed);
```
torkleyy: Same as above

src/world/entity.rs
```diff
@@ -432,3 +397,50 @@ impl Generation {
...
+
+#[derive(Default, Debug)]
+struct EntityCache {
+    cache: Vec<Index>,
+    len: AtomicUsize,
+}
+
+impl EntityCache {
+    fn pop_atomic(&self) -> Option<Index> {
+        let mut prev = self.len.load(Ordering::Relaxed);
```
torkleyy: Can you move this into `atomic_decrement`?

src/world/entity.rs
```diff
@@ -432,3 +397,50 @@ impl Generation {
...
+
+impl Extend<Index> for EntityCache {
+    fn extend<T: IntoIterator<Item = Index>>(&mut self, iter: T) {
+        self.maintain();
+        self.cache.extend(iter);
+        self.len.store(self.cache.len(), Ordering::Relaxed);
+    }
+}
+
+fn atomic_increment(i: &AtomicUsize) -> Option<usize> {
```
torkleyy: Please add a doc comment like this:

```
/// Increments `i` atomically without wrapping on overflow.
/// Resembles a `fetch_add(1, Ordering::Relaxed)` with
/// checked overflow, returning `None` instead.
```

src/world/entity.rs
```diff
@@ -65,21 +66,19 @@ impl Allocator {
...
-            // Since atomically created entities don't have a generation until merge
-            // is called, the first entity of a generation won't actually have a generation
-            // in the allocator. And it is fine for us to just ignore the entity since
-            // we clear `self.raised` before so it won't get allocated later.
-            if self.generations.len() > id {
-                self.generations[id].die();
+
+            while self.generations.len() <= id as usize {
+                self.generations.push(Generation(0));
             }
```
WaDelma: I think this is clearer:
```rust
if self.generations.len() <= id as usize {
    self.generations.resize(id as usize + 1, Generation(0));
}
```

src/world/entity.rs
```diff
@@ -122,66 +121,41 @@ impl Allocator {
...
-                    .unwrap_or(Generation(1));
-
-                return Entity(i as Index, gen);
-            }
-        }
-        panic!("No entities left to allocate")
+        let id = match self.cache.pop_atomic() {
+            Some(x) => x,
+            None => atomic_increment(&self.max_id).expect("No entity left to allocate") as Index,
+        };
```
WaDelma: This looks lot like `unwrap_or_else`

src/world/entity.rs
```diff
@@ -122,66 +121,41 @@ impl Allocator {
...
+            if id as Index == <Index>::max_value() {
+                panic!("No entity left to allocate");
             }
+            self.max_id.store(id as usize + 1, Ordering::Relaxed);
+            id as Index
+        });
+
+        while self.generations.len() <= id as usize {
+            self.generations.push(Generation(0));
         }
```
WaDelma: `resize` as above. Maybe make it a helper function?

Flecheck:
I addressed your comments and added a benchmark on the aspects this reduction of complexity was meant to address. There seems to be no regression on existing features (but please keep in mind that I experienced very inconsistent results, maybe because of my laptop).

However, the new benchmarked scenarios gained about 30-40% in speed between master and this PR.
Those two scenarios are : random deletions in a large pool of entities then creation, and deletion of the oldest entity in a large pool of entities then creation. Those scenarios seem realistic enough.

torkleyy:
Thank you! Maybe you can increase the sample size?

Flecheck:
What do you mean by sample size?

torkleyy:
> What do you mean by sample size?

@Flecheck The number of benchmarks you run to determine the performance.

torkleyy:
But I think this PR can be merged now. More reviews are welcome though @WaDelma @Xaeroxe ;)

torkleyy:
Theo did you check the performance change for an Amethyst demo (using the
entity allocator a lot).

Moxinilian:
Well there is no real situation in Amethyst where the allocator is the bottleneck. I did try it on amethyst-rhusics' boxes3d demo, and the FPS started dropping on my machine around 6000 entities because of the physics computations. I could not notice any performance difference on other Amethyst examples.

However I was there when the benchmarks on this PR were done, so let me expand a bit on the inconsistencies of the benchmarks. When Flecheck ran the world benchmarks, the new cache always outperformed master on the new bench cases. However, we also got results that did not make sense on operations that had nothing to do with the cache such as join parallelism (regressions or improvements of about 20%, randomly). There is no doubt this PR increases performances in some scenarios, and there are no reasons to believe it slows down other things, as the most stable benchmarks we had did not report any impact in the performances.

PR #460, opened 2018-09-08
Saveload custom derive

WaDelma:
Split from #434
A custom derive for IntoSerialize and FromDeserialize named #[derive(Saveload)]. It may seem odd to derive them together, but after spending significant time trying to do it the normal way (one derive per trait) it became clear that due to the requirement of "proxy" types that fill the IntoSerialize::Data and FromDeserialize::Data fields, it made sense to do it this way or you start getting into attribute hell.

---
This change is [Reviewable](https://reviewable.io/reviews/slide-rs/specs/460)

### DIFF
--- a/specs-derive/Cargo.toml
+++ b/specs-derive/Cargo.toml
@@ -9,8 +9,12 @@ keywords = ["gamedev", "parallel", "specs", "ecs", "derive"]
 license = "MIT/Apache-2.0"
 
 [dependencies]
-syn = "0.12"
+syn = {version="0.12", features=["extra-traits"]}
 quote = "0.4"
+
+[dev-dependencies]
+specs = { path = "..", features=["serde"] }
+serde = "1"
 
 [lib]
 proc-macro = true
--- a/specs-derive/Cargo.toml
+++ b/specs-derive/Cargo.toml
@@ -9,11 +9,11 @@ keywords = ["gamedev", "parallel", "specs", "ecs", "derive"]
 license = "MIT/Apache-2.0"
 
 [dependencies]
-syn = {version="0.12", features=["extra-traits"]}
+syn = { version = "0.12", features = ["extra-traits"] }
 quote = "0.4"
 
 [dev-dependencies]
-specs = { path = "..", features=["serde"] }
+specs = { path = "..", features = ["serde"] }
 serde = "1"
 
 [lib]
--- a/specs-derive/src/impl_saveload.rs
+++ b/specs-derive/src/impl_saveload.rs
@@ -169,7 +169,7 @@ fn saveload_struct(
 /// This generates a struct named `StructNameSaveloadData` such that all fields are their associated `Data` variants, as well as a bound on the required marker
 ///  e.g.
 ///
-/// ```
+/// ```nobuild
 /// struct FooSaveloadData<MA> where MA: Serialize+Marker, for<'de> MA: Deserialize<'de> {
 ///    e: <Entity as IntoSerialize<MA>>::Data
 /// }
@@ -178,7 +178,7 @@ fn saveload_struct(
 /// The generation for the `into` and `from` functions just constructs each field by calling `into`/`from` for each field in the input struct
 /// and placing it into the output struct:
 ///
-/// ```
+/// ```nobuild
 ///  fn into<F: FnMut(Entity) -> Option<MA>>(&self, mut ids: F) -> Result<Self::Data, Self::Error> {
 ///      FooSaveloadData {
 ///          e: IntoSerialize::into(&self.e, &mut ids)?
@@ -226,8 +226,8 @@ fn saveload_named_struct(
 /// This generates a struct named `StructNameSaveloadData` such that all fields are their associated `Data` variants, as well as a bound on the required marker
 ///  e.g.
 ///
-/// ```
-/// struct FooSaveloadData<MA>  (
+/// ```nobuild
+/// struct FooSaveloadData<MA> (
 ///    <Entity as IntoSerialize<MA>>::Data
 /// ) where MA: Serialize+Marker, for<'de> MA: Deserialize<'de>;
 /// ```
@@ -235,7 +235,7 @@ fn saveload_named_struct(
 /// The generation for the `into` and `from` functions just constructs each field by calling `into`/`from` for each field in the input struct
 /// and placing it into the output struct:
 ///
-/// ```
+/// ```nobuild
 ///  fn into<F: FnMut(Entity) -> Option<MA>>(&self, mut ids: F) -> Result<Self::Data, Self::Error> {
 ///      FooSaveloadData (
 ///          e: IntoSerialize::into(&self.0, &mut ids)?
@@ -290,7 +290,7 @@ fn saveload_tuple_struct(
 /// This generates a struct named `EnumNameSaveloadData` such that all fields are their associated `Data` variants, as well as a bound on the required marker
 ///  e.g.
 ///
-/// ```
+/// ```nobuild
 /// enum FooSaveloadData<MA> where MA: Serialize+Marker, for<'de> MA: Deserialize<'de> {
 ///    Bar(<Entity as IntoSerialize<MA>>::Data),
 ///    Baz { e: <Entity as IntoSerialize<MA>>::Data },
@@ -301,7 +301,7 @@ fn saveload_tuple_struct(
 /// The generation for the `into` and `from` functions just constructs each field of each variant by calling `into`/`from` for each field in the input
 /// and placing it into the output struct in a giant match of each possibility:
 ///
-/// ```
+/// ```nobuild
 ///  fn into<F: FnMut(Entity) -> Option<MA>>(&self, mut ids: F) -> Result<Self::Data, Self::Error> {
 ///      match *self {
 ///          Foo::Bar(ref field0) => FooSaveloadData::Bar(IntoSerialize::into(field0, &mut ids)? ),
--- a/specs-derive/src/impl_saveload.rs
+++ b/specs-derive/src/impl_saveload.rs
@@ -1,0 +1,496 @@
+//! Contains implementations for #[derive(Saveload)]
+//! which derives both `IntoSerialize` and `FromDeserialize`.
+//!
+//! Since this requires a proxy "Data" `struct`/`enum`, these two need
+//! to be derived together or it begins requiring unwieldy attribute explosion.
+//!
+//! Currently there is one major limitation to this implementation: it cannot handle
+//! generic types such as `struct Foo<T> { x: T, e: Entity }`. I'm not good enough at
+//! using `syn` to iron this out, but since specs requires explicit references to non-generic
+//! components in system data, it can generally be worked around by newtyping like
+//! `struct TypedFoo(Foo<u32>);` and the like.
+
+// NOTE: All examples given in the documentation below are "cleaned up" into readable Rust,
+// so it doesn't give an entirely accurate view of what's actually generated (for instance,
+// I'll write `Entity` instead of `::specs::Entity` which is actually what's generated).
+
+use quote::Tokens;
+use syn::{DataEnum, DataStruct, DeriveInput, Field, Generics, Ident, Type, TypeParamBound};
+
+// Handy collection since tuples got unwieldy and
+// unclear in purpose
+struct SaveloadDerive {
+    type_def: Tokens,
+    ser: Tokens,
+    de: Tokens,
+    saveload_name: Ident,
+    saveload_generics: Generics,
+}
+
+/// The main entrypoint, sets things up and delegates to the
+/// type we're deriving on.
+pub fn impl_saveload(ast: &mut DeriveInput) -> Tokens {
+    use syn::Data;
+
+    if !ast.generics.params.is_empty() {
+        panic!("Deriving `Saveload` does not, at the moment, support generic types (i.e. `struct Foo<T>`).
+        
+        This limitation may be fixed in the future, but for now you can usually work around this by newtyping
+        your generic data structure to a concrete-variant (e.g. `struct FooU32(Foo<u32>);`), or implementing `IntoSerialize` and `FromDeserialize` 
+        yourself, which is relatively straightforward if tedious (see the documentation for those traits).")
+    }
+
+    add_bound(
+        &mut ast.generics,
+        parse_quote!(::specs::saveload::IntoSerialize),
+    );
+    add_bound(
+        &mut ast.generics,
+        parse_quote!(::specs::saveload::FromDeserialize),
+    );
+    add_bound(&mut ast.generics, parse_quote!(Clone));
+
+    let derive = match ast.data {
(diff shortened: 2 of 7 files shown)


omni-viral:
There is no much sense in deriving only one trait anyway. We could merge them altogether.

torkleyy:
> There is no much sense in deriving only one trait anyway. We could merge them altogether.

I split them earlier, I'm sure there was a reason for that. Deriving two traits at once doesn't seem odd to me.

specs-derive/src/impl_saveload.rs
```diff
@@ -0,0 +1,487 @@
...
+    add_bound(
+        &mut ast.generics,
+        parse_quote!(::specs::saveload::FromDeserialize),
+    );
+    add_bound(&mut ast.generics, parse_quote!(Clone));
+
+    let derive = match ast.data {
+        Data::Struct(ref mut data) => saveload_struct(data, &mut ast.ident, &mut ast.generics),
+        Data::Enum(ref data) => saveload_enum(data, &ast.ident, &ast.generics),
+        Data::Union(_) => panic!("Unions cannot derive IntoSerialize/FromDeserialize at present"),
```
torkleyy: won't in the future (missing tag); missing backticks + spacing

specs-derive/src/impl_saveload.rs
```diff
@@ -0,0 +1,487 @@
...
+        }
+    };
+
+    // panic!("{}", tt);
+
+    tt
+}
+
+// Implements all elements of saveload common to structs of any type
+// (except unit structs like `struct Foo;` since they're auto-saveload);
```
torkleyy: I don't quite understand this sentence.

specs-derive/src/impl_saveload.rs
```diff
@@ -0,0 +1,487 @@
...
+    let mut saveload_generics = generics.clone();
+    saveload_generics.params.push(parse_quote!(MA));
+    match &mut saveload_generics.where_clause {
+        Some(clause) => clause.predicates.push(parse_quote!(
+            MA: ::serde::Serialize + ::serde::de::DeserializeOwned + ::specs::saveload::Marker,
+            for <'deser> MA: ::serde::Deserialize<'deser>
+        )),
+        clause => *clause = Some(parse_quote!(
+            where MA: ::serde::Serialize + ::serde::de::DeserializeOwned + ::specs::saveload::Marker,
+            for <'deser> MA: ::serde::Deserialize<'deser>
```
torkleyy: Why is this necessary? Isn't this guaranteed by `DeserializeOwned`?

specs-derive/src/impl_saveload.rs
```diff
@@ -0,0 +1,487 @@
...
+    } else if let Fields::Unnamed(_) = data.fields {
+        saveload_tuple_struct(
+            &data,
+            &name,
+            &saveload_name,
+            &saveload_generics,
+            &saveload_fields,
+        )
+    } else {
+        panic!("This derive does not support unit structs (Hint: they automatically derive FromDeserialize and IntoSerialize)");
```
torkleyy: Ah now I get the above sentence; both places should describe it more like "every unit struct can derive `Serialize` and `Deserialize`, thus has a blanket impl fulfill their `IntoSerialize` / `IntoDeserialize`".

torkleyy:
@WaDelma Are you going to be working on this?

WaDelma:
Yeah I'll work on this.

specs-derive/src/impl_saveload.rs
```diff
@@ -0,0 +1,487 @@
...
+        }
+        Type::Group(ty) => replace_entity_type(&mut *ty.elem),
+
+        Type::TraitObject(_) => {}
+        Type::ImplTrait(_) => {}
+        Type::Slice(_) => {
+            panic!("Slices are unsupported, use owned types like Vecs or Arrays instead")
+        }
+        Type::Reference(_) => panic!("References are unsupported"),
+        Type::Ptr(_) => panic!("Raw pointer types are unsupported"),
```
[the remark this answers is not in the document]
LinearZoetrope: I didn't want to deal with reference/pointer graphs within serialized components (**especially** with cycles).

The biggest problem is it's not exactly clear what to even **do** with types that hold references while deserializing, since you'd need to somehow deserialize the contents of the reference and then pass that handle back to the callsite as well.

The funny thing is, we have a pointer that solves all these problems for us -- Entity, which is just a generational weak pointer. So I see no reason to try and solve these issues when we can tell the user "either write your own `impl` to explain what you mean by "(de)serialize this pointer" or split this into multiple components and use an `Entity` ID lookup".

specs-derive/src/impl_saveload.rs
LinearZoetrope: Yeah, I think I wasn't thinking when I wrote this. TBH, we could probably just silently do nothing because of this instead of panicking. I think I decided to be explicit for some reason.

specs-derive/src/impl_saveload.rs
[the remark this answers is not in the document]
torkleyy: Oh right. I don't know what I was thinking there.

specs-derive/tests/saveload.rs
```diff
@@ -0,0 +1,70 @@
+// These are mostly just compile tests
+#![allow(dead_code)]
```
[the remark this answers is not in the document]
LinearZoetrope: See the overall thread because this is really important.

LinearZoetrope:
@torkleyy says

>Please just check the result of the method calls, so we've got some additional coverage and don't need to disable warnings. Would be nice to have some more "exotic" types involving entities, like Vec<Entity>.


I can tell you that `Vec<Entity>` doesn't work right now. I think this info got lost in the split, but back then I said:
----

Crap, I forgot code like

`impl IntoSerialize for C where C: Serialize+Clone` vs `impl IntoSerialize for Vec<T> where T: IntoSerialize` would require specialization to resolve (and even then IDK if specialization would solve this).

Not sure how to solve this from a design standpoint, seems like there are only really two options:

1. Hand-implement `IntoSerialize` and `FromDeserialize` for every native type to avoid a blanket impl, and expect others to `#[derive(Saveload)]` for their own types.
2. Extend the derive to have special casing for things like `Vec<T>` and generate specialized code

I feel like 1 is more painful but probably ultimately more flexible, it seems closer to how `serde` works, but is also a massive undertaking (not in difficulty so much in time and scale).

----

I'm still not sure how to fix it.

torkleyy:
Ah right I just read it. What I don't understand though: `Vec<Entity>` does not implement `Serialize` so it should also not implement `IntoSerialize`. So where do you need specialization here?

LinearZoetrope:
`Vec<Entity>` could probably be done on its own, but `Vec<T>` such that `T` may contain an entity can't.

torkleyy:
I don't think that'd be such a big issue.

## Suggested solution

* implement traits for `Vec<Entity>`
* other fields can either be solved by
   * creating a newtype with an implementation
   * providing an attribute like `#[saveload(data = "FooVecData", into = "into_foo_vec_data", from = "from_foo_vec_data"]`

specs-derive/src/impl_saveload.rs
```diff
@@ -0,0 +1,484 @@
...
+/// ) where MA: Serialize+Marker, for<'de> MA: Deserialize<'de>;
+/// ```
+///
+/// The generation for the `into` and `from` functions just constructs each field by calling `into`/`from` for each field in the input struct
+/// and placing it into the output struct:
+///
+/// ```nobuild
+///  fn into<F: FnMut(Entity) -> Option<MA>>(&self, mut ids: F) -> Result<Self::Data, Self::Error> {
+///      FooSaveloadData (
+///          e: IntoSerialize::into(&self.0, &mut ids)?
```
LinearZoetrope: Tuple struct, shouldn't have a name here.

WaDelma:
I stumbled into problem while trying to make custom derive to work with generics: I need to assert that `<T as FromDeserialize<MA>>::Data = <T as IntoSerialize<MA>>::Data` which triggers `error: equality constraints are not yet supported in where clauses (#20041)`.

So we would need to either wait for that to stabilize, merge the traits or figure some other workaround for it.

[This](https://play.rust-lang.org/?gist=ba67f27e46e2be9d4c6cbec77e362785&version=stable&mode=debug&edition=2015) is my current macro expansion

LinearZoetrope:
Merging the traits seems adequate, to me. I don't see much near-term utility in having them split with how all the `saveload` stuff currently works (as opposed to `serde` where there can be good reasons to implement only one of (De)serialize). If somebody has a compelling reason they can be split again.

WaDelma:
After merging the traits and getting code gen work I tried to implement `ConvertSaveload` (I am not fond of the name) for `Vec<Entity>`, but it cannot be done:
```rust
error[E0119]: conflicting implementations of trait `saveload::ConvertSaveload<_>` for type `std::vec::Vec<world::entity::Entity>`:
   --> src/saveload/mod.rs:187:1
    |
143 | / impl<C, M> ConvertSaveload<M> for C
144 | | where
145 | |     C: Clone + Serialize + DeserializeOwned,
146 | | {
...   |
162 | |     }
163 | | }
    | |_- first implementation here
...
187 | / impl<M> ConvertSaveload<M> for Vec<Entity>
188 | | where
189 | |     M: Serialize + DeserializeOwned,
190 | | {
...   |
206 | |     }
207 | | }
    | |_^ conflicting implementation for `std::vec::Vec<world::entity::Entity>`
    |
    = note: upstream crates may add new impl of trait `serde::Serialize` for type `std::vec::Vec<world::entity::Entity>` in future versions
    = note: upstream crates may add new impl of trait `serde::Deserialize<'de>` for type `std::vec::Vec<world::entity::Entity>` in future versions
```

specs-derive/Cargo.toml
```diff
@@ -9,8 +9,12 @@ keywords = ["gamedev", "parallel", "specs", "ecs", "derive"]
 license = "MIT/Apache-2.0"
 
 [dependencies]
-syn = "0.12"
+syn = {version="0.12", features=["extra-traits"]}
```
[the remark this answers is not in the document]
WaDelma: Tried to update this, but the previous derive code was using `synom` which was removed from re-exports in newer one and trying to import it manually didn't work out of the box. Will need more work to do this.

Zaerei:
Looks like since in the future they're allowing upstream impls of `Serialize` and `Deserialize` there could possible be a blanket impl conflict because, if an upstream crate did it, then suddenly the `impl<C>` would work? That's... bizarre and unfortunate.

Zaerei:
I've checked and annoyingly, even specialization doesn't solve this.

omni-viral:
Can addition of `C: Component` fix this?

WaDelma:
Adding `Component` bound seems to fix it, but I am not sure if it's right trait here.

Real solution would be adding bound for:
```rust
auto trait NoEntity {}

impl !NoEntity for Entity {}
```
but auto traits are unstable so that's out of picture

omni-viral:
I see. This trait is for deriving.
I have an idea.

torkleyy:
Since this seems rather experimental, please move this into an optional crate / hide it behind a feature flag. If this only works for nightly, that would also be a way to get a first implementation.

WaDelma:
I was going to leave it out of this PR anyways. Is there still something missing here otherwise?

specs-derive/src/impl_saveload.rs
```diff
@@ -0,0 +1,487 @@
+//! Contains implementations for #[derive(Saveload)]
+//! which derives both `IntoSerialize` and `FromDeserialize`.
+//!
+//! Since this requires a proxy "Data" `struct`/`enum`, these two need
+//! to be derived together or it begins requiring unwieldy attribute explosion.
+//!
+//! Currently there is one major limitation to this implementation: it cannot handle
+//! generic types such as `struct Foo<T> { x: T, e: Entity }`. I'm not good enough at
```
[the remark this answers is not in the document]
torkleyy: Just starting with the review, is this fixed now?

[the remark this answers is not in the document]
WaDelma: Yes. Forgot to remove push commit that removes that documentation

src/saveload/mod.rs
```diff
@@ -41,3 +46,140 @@ pub struct EntityData<M, D> {
...
+
+
+/// Converts a data type (usually a [`Component`]) into its serializable form
+/// and back to actual data from it's deserialized form.
+///
+/// This is automatically implemented for any type that is
+/// [`Clone`], [`Serialize`] and [`DeserializeOwned`].
+///
+/// Implementing this yourself is usually only needed if you
+/// have a component that points to another Entity, or has a field which does,
```
Zaerei: Entity -> [`Entity`] (and subsequent link at the bottom)

src/saveload/mod.rs
```diff
@@ -41,3 +46,140 @@ pub struct EntityData<M, D> {
...
+///  and you wish to [`Serialize`] it.
+///
+/// *Note*: if you're using `specs_derive`
+/// and your struct does not have a generic bound (i.e. `struct Foo<T>`),
```
Zaerei: The generic bound thing was fixed, so we can remove this part.

WaDelma:
"Please import the prelude.":
I removed import for that to make sure all things referred in custom derive use full path (previously there were inconsistencies here *because* prelude was used).
I am not sure if I should change everything to work like in #467 though

re relative paths:
Yeah I checked that they work. I think I can actually use `trait.Component.html` to refer them if they are in same crate

WaDelma:
Absolute links also break if the user is using locally generated documentation

WaDelma:
@torkleyy Should I change things inside the custom derive to work like in #467?
That and concern on the merging of trait/renaming it are only things left.

torkleyy:
@WaDelma I don't know what you mean by that rn. I'll review both PRs tomorrow if possible.

WaDelma:
Currently the custom derive does `::specs::Entity`, but in that PR the other custom derive was changed to do `Entity` so you have to import everything the derive needs. In exchange you get ability to (easily) use without direct specs dependency. (You can still make top level module named specs I think)

specs-derive/tests/saveload.rs
```diff
@@ -1,18 +1,21 @@
 #[macro_use]
 extern crate serde;
-extern crate specs;
+// Specs is renamed here so that the custom derive cannot refer specs directly
+extern crate specs as spocs;
 #[macro_use]
 extern crate specs_derive;
 
+use spocs::{Entity, saveload::{ConvertSaveload, Marker}, error::NoError};
```
torkleyy: `spocs`

WaDelma: Yes. Because I renamed `specs` to `spocs` at in extern crate. There is comment that explains why.

torkleyy: Oh ah :smile:

Zaerei: It's October,  obviously should be `spooks`.

WaDelma:
bors +
Ah yeah, this PR

WaDelma:
bors r+

Ah yeah, this PR.

PR #461, opened 2018-09-08
Saveload convenience macro

WaDelma:
Split from #434 
A macro which automatically will create SystemData for the most common usecase of SerializeComponents and DeserializeComponents. Unlike the existing tuple auto-implementation, this generates arbitrarily large structs that you can readily plop straight into your serialization or deserialization system. It also doesn't use recursion, so it shouldn't hit any nasty macro expansion recursion limits.

---
This change is [Reviewable](https://reviewable.io/reviews/slide-rs/specs/461)

### DIFF
--- a/tests/saveload_macro.rs
+++ b/tests/saveload_macro.rs
@@ -99,6 +99,8 @@ impl<'a> System<'a> for SerSys {
 
 #[test]
 fn saveload_macro() {
+    use specs::saveload::MarkedBuilder;
+
     let mut world = setup_world();
     let entity = world
         .create_entity()
--- a/tests/saveload_macro.rs
+++ b/tests/saveload_macro.rs
@@ -1,0 +1,145 @@
+#![cfg(feature = "serde")]
+
+#[macro_use]
+extern crate specs;
+#[macro_use]
+extern crate specs_derive;
+#[macro_use]
+extern crate shred_derive;
+#[macro_use]
+extern crate serde;
+extern crate ron;
+extern crate serde_json;
+extern crate shred;
+
+use specs::prelude::*;
+use specs::saveload::{U64Marker, U64MarkerAllocator};
+
+#[derive(Copy, Clone, Serialize, Deserialize, Component, Debug, PartialEq)]
+#[storage(VecStorage)]
+struct Pos {
+    x: f64,
+    y: f64,
+}
+
+#[derive(Copy, Clone, Serialize, Deserialize, Component, Debug, PartialEq)]
+#[storage(VecStorage)]
+struct Vel {
+    x: f64,
+    y: f64,
+}
+
+#[derive(Copy, Clone, Component, Saveload)]
+#[storage(VecStorage)]
+struct OwnsEntity(Entity);
+
+saveload_components!{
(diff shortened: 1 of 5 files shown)


WaDelma:
Ah didn't notice that test was using the custom derive.

tests/saveload_macro.rs
```diff
@@ -0,0 +1,147 @@
...
+}
+
+#[derive(Clone, Eq, PartialEq, Hash, Default, Debug)]
+struct Serialized {
+    json: String,
+}
+
+#[derive(SystemData)]
+struct DeSysData<'a> {
+    de: saveload_generated::DeData<'a, U64Marker, U64MarkerAllocator>,
```
torkleyy: We had a `WorldSerialize` before that had this convenience without macros on the user side.

torkleyy:
Btw @LinearZoetrope I love how you document everything in such detail :)

torkleyy:
Ah, there it is:

https://github.com/slide-rs/specs/pull/275/files#diff-9765da818432e066259a0076a4f94a6cR97

Used in action:

https://github.com/slide-rs/specs/pull/283/files#diff-32317e23ec50b4b4b1ad0bffba3a4a12dR86

LinearZoetrope:
The reason for the macro is to get around the tuple-size limitation you bump into with static types like `WorldSerialize`. If Rust had some way of generalizing over an arbitrary-sized tuple of types that all met some criterion it wouldn't matter. The intention was to just build something like a `WorldSerialize` at the user's request that was arbitrarily sized to their data instead of being stuck providing necessarily fixed-sized implementations.

It's unfortunately really tedious and error prone to hand-implement the relevant traits here in most cases because it's just copy+pasting a dozen `foo.bar.serialize(m)`, `foo.baz.serialize(m)` lines. This was motivated mostly by my current usecase where I have more than 18 component types or whatever the max autoderived value is atm.

torkleyy:
I personally like the approach we're taking shred:

* tuples in most cases
* struct with `#[derive]` if it gets too big
* tuples can be nested if you exceed the limit and you are ~~lazy~~ saving resources

LinearZoetrope:
That's reasonable. My other main concern is that the current, Post-`World(De)serialize` approach is that since it requires manually specifying the component mutability (e.g. `Read<Foo>`), it's fragile and annoying to keep the tuples in sync. You used to be able to do something like:

```
type SerComponents = (Foo, Bar, Baz);

// ...

WorldSerialize<..., SerComponents>
WorldDeserialize<..., SerComponents>
```

But now you have to manually do

```
type SerComponents = (Read<Foo>, Read<Bar>, Read<Baz>);
type DeserComponents = (Write<Foo>, Write<Bar>, Write<Baz>);
```

Things can break in really fun and baffling ways if these lists get out of sync (in fact, I think even *order* matters). A big part of the macro's intention was so I only had to write the list once.

torkleyy:
I'm not quite sure what to do with this one. @WaDelma Do you still want to continue this?

WaDelma:
I am not that sure about this macro and so I would be willing to drop it. But I haven't used the saveload stuff in real project so I am not 100% certain about it.

torkleyy:
Closing for now.

Issue #361, opened 2018-03-26
another storage to help synchronizing with external storage


thiolliere:
I use specs with nphysics (actually a fork on my own but 0.8 nphysics will impl things..) I have to synchronize RigidBodyHandle in specs world and in nphysics world.

My needs are:
* for insert everything is sync: the method to create RigidBodyHandle component takes nphysics world and create them altogether
* for remove I accept that some object are still alive in nphysics world but have no component in specs. and then I want to regularly sync things by removing those objects that have been deleted from specs world. But to do so I have issues:

I tried to implement using flagged storage in current master but I can't get the deleted component I just have the index of the entity deleted and that is not enough:
```rust
extern crate specs;

use specs::prelude::*;

#[derive(Debug)]
struct Vel(usize);

impl Component for Vel {
    type Storage = FlaggedStorage<Self, VecStorage<Self>>;
}

fn main() {
    let mut vels = vec![];
    let mut world = World::new();
    world.register::<Vel>();
    let mut removed_id = world.write::<Vel>().track_removed();

    vels.push(0.0);
    let e0 = world.create_entity().with(Vel(0)).build();

    {
        // remove an entity
        println!("entity removed:  {:?}", e0);
        world.delete_entity(e0).unwrap();

        // add an entity
        vels.push(0.3);
        let e1 = world.create_entity().with(Vel(1)).build();
        println!("entity created:  {:?}", e1);

        // try to find removed component
        let mut removed = BitSet::new();
        world.write::<Vel>().populate_removed(&mut removed_id, &mut removed);

        let entities = world.entities();
        for (e, _, v) in (&*entities, &removed, &world.read::<Vel>()).join() {
            // this does show we found new component instead
            println!("component of removed ? {:?}: {:?}", e, v);
        }
    }
}
```

The solution I'll implement is to create a new storage that keeps deleted component in some cache that I'll clear regularly (with removing their nphysic body).
This implementation is straightforward but might be needed by others so that is why I open this issue and also to know if you have better solution.

torkleyy:
There's no too obvious solution as far as I can tell with my limited knowledge of nphysics.

I think I need some more information to give a good recommendation. I'll look into it later.

thiolliere:
I don't think it is  related to nphysics, in my example I use an external storage of velocities.

[...]

PR #463, opened 2018-09-09
Add tests for tracking component insertions / removals

torkleyy:
@Rhuagh how was our strategy for keeping track of all the existing components?

Related: #361

---
This change is [Reviewable](https://reviewable.io/reviews/slide-rs/specs/463)

### DIFF
--- a/tests/tracked.rs
+++ b/tests/tracked.rs
@@ -1,0 +1,82 @@
+extern crate specs;
+
+use specs::prelude::*;
+
+struct MyComp;
+
+impl Component for MyComp {
+    type Storage = FlaggedStorage<Self>;
+}
+
+#[test]
+fn test_remove_insert() {
+    let mut world = World::new();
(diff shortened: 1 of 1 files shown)


GraemeWilde:
I saw your comment in #361, what part are you trying to figure out?

torkleyy:
I know how you could do it; I just thought there was a concise way to do
this. I'm considering to add a better API and / or change the existing one.

torkleyy:
Rhuagh suggested that we could merge all events into a single channel so we get the order of the events. I generally like the idea, I'm just worried it'll leave us with a worse access pattern.

torkleyy:
Another option would be to leave the API as-is and create a wrapper that does all the black magic that makes the current API difficult. Rhuagh's solution would be a lot cleaner though..

torkleyy:
Closing due to inactivity.

PR #467, opened 2018-09-13
Switched specs-derive to use Component in scope

jojolepro:
"Works on my pc".

So with this you need to have `Component` in scope for the derive to find it and use it. This allows to use it in projects without having to have specs as a hard dependency. <3

Also it might fix the issue with re-exporting the macro, but I didn't test that.

---
This change is [Reviewable](https://reviewable.io/reviews/slide-rs/specs/467)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -52,7 +52,7 @@ criterion = "0.2"
 ron = "0.4"
 rand = "0.5.5"
 serde_json = "1.0"
-specs-derive = { path = "specs-derive", version = "0.2.0" }
+specs-derive = { path = "specs-derive", version = "0.3.0" }
 
 [[example]]
 name = "async"
--- a/specs-derive/Cargo.toml
+++ b/specs-derive/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "specs-derive"
-version = "0.2.1"
+version = "0.3.0"
 authors = ["Eyal Kalderon <ebkalderon@gmail.com>"]
 description = "Custom derive macro for Specs components"
 documentation = "https://docs.rs/specs-derive"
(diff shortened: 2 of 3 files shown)


randomPoison:
I believe this is a breaking change, isn't it? Any code that uses `#[derive(Component)]` and doesn't explicitly bring `Component` into scope is going to fail.

As a user of specs (by way of Amethyst), I also don't love the prospect of having to manually import `Component` into scope for every file where I use `#[derive(Component)]`. This approach is pretty counter to the common pattern in custom derives to always use the full path internally, which means its also potentially confusing for new users who are expecting `#[derive(Component)]` to work the way most other derives do.

For the record, I'm not against this change. I think being able to derive `Component`, even if I have to manually bring it into scope, would be an improvement. But I also think we're going to eventually want a more robust solution.

jojolepro:
Of course, but before this change you needed to manually add specs to your Cargo.toml and add a "extern crate specs". In the case of amethyst where its already re-exported (amethyst::core::ecs), you end up with two versions of specs. If you are not careful, you'll get two different versions and get hard to track errors.

But yes a way of doing it with absolute paths would be nice indeed.
By the way, if you check at the source of the derive, in the lines I didn't touch DenseVecStorage is not imported as an absolute path either. https://github.com/jojolepro/specs/blob/derive/specs-derive/src/lib.rs#L58

:D

randomPoison:
You know what, I feel like I've run into that before and been annoyed about having to manually import `DenseVecStorage` 🤔

Xaeroxe:
@jojolepro tests failed, can you update the version usage as well?

torkleyy:
@Xaeroxe We should've held the merge off until 0.13..

Xaeroxe:
We did?  I'm planning a 0.13 soon.

PR #469, opened 2018-09-15
resolve method not defined error

Axelderan:
With this code so far, following along:
```
extern crate specs;
#[macro_use]
extern crate specs_derive;

use specs::{VecStorage, World};


#[derive(Component, Debug)]
#[storage(VecStorage)]
struct Position {
    x: f32,
    y: f32
}

#[derive(Component, Debug)]
#[storage(VecStorage)]
struct Velocity {
    x: f32,
    y: f32,
}

fn main() {
    let mut world = World::new();
    world.register::<Position>();

    let ball = world.create_entity().with(Position { x: 4.0, y: 7.0 }).build();
}
```
one gets this error:
```
error[E0599]: no method named `with` found for type `specs::EntityBuilder<'_>` in the current scope
  --> src/main.rs:26:38
   |
26 |     let ball = world.create_entity().with(Position { x: 4.0, y: 7.0 }).build();
   |                                      ^^^^
   |
   = help: items from traits can only be used if the trait is in scope
help: the following trait is implemented but not in scope, perhaps add a `use` for it:
   |
5  | use specs::Builder;
   |
```
This change just does what's suggested and we can compile fine.

I like to compile often, letting `rustc` check for typos and other errors.  I'm sure other beginners do also.

---
This change is [Reviewable](https://reviewable.io/reviews/slide-rs/specs/469)

### DIFF
--- a/book/src/02_hello_world.md
+++ b/book/src/02_hello_world.md
@@ -80,7 +80,7 @@ need to create a world in which to store all of our components.
 ## The `World`
 
 ```rust,ignore
-use specs::World;
+use specs::{World, Builder};
 
 let mut world = World::new();
 world.register::<Position>();


Axelderan:
Sure, I included the full code so you could do that.  It works where it didn't before.  Shrug.

I understand though that I might not be following along as the book intends one to.  It seems like the code examples are supposed to be digested in isolation, while the final, integrated example is at the end of the chapter.

No worries if this PR is not in keeping with the style.  I still find the book presents things in a good manner, as someone new to both Rust and ECS.

Thank you for the lib!

PR #472, opened 2018-09-18
Implement Join for Fetch/Read/Write resources

Aceeri:
So we can finally get rid of the confusion with `&*entities` and such.

---
This change is [Reviewable](https://reviewable.io/reviews/slide-rs/specs/472)

### DIFF
--- a/benches/storage_cmp.rs
+++ b/benches/storage_cmp.rs
@@ -54,7 +54,7 @@ where
             let entities = world.entities();
             let mut storage = world.write_storage::<C>();
 
-            for e in (&*entities).join() {
+            for e in (&entities).join() {
                 storage.remove(e);
             }
         },
@@ -87,7 +87,7 @@ where
             let entities = world.entities();
             let storage = world.read_storage::<C>();
 
-            for e in (&*entities).join() {
+            for e in (&entities).join() {
                 black_box(storage.get(e));
             }
         },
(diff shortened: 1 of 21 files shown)


src/join.rs
```diff
@@ -547,3 +549,74 @@ define_open!{A, B, C, D, E, F, G, H, I, J, K, L, M, N, O}
 define_open!{A, B, C, D, E, F, G, H, I, J, K, L, M, N, O, P}
 define_open!(A, B, C, D, E, F, G, H, I, J, K, L, M, N, O, P, Q);
 define_open!(A, B, C, D, E, F, G, H, I, J, K, L, M, N, O, P, Q, R);
+
+macro_rules! immutable_resource_join {
```
jojolepro: Just missing a tiny bit of docs here. ;)

benches/storage_cmp.rs
```diff
@@ -87,7 +87,7 @@ where
             let entities = world.entities();
             let storage = world.read_storage::<C>();
 
-            for e in (&*entities).join() {
+            for e in (&entities).join() {
```
randomPoison: If I recall correctly, the parenthesis here are also unnecessary, i.e. you can do `for e in entities.join() {}`. I've been able to do that with components, at least, so hopefully that still applies with `Entities` now 🤞

src/world/mod.rs
```diff
@@ -515,7 +515,7 @@ impl World {
     pub fn delete_all(&mut self) {
         use join::Join;
 
-        let entities: Vec<_> = (&*self.entities()).join().collect();
+        let entities: Vec<_> = (&self.entities()).join().collect();
```
randomPoison: Same goes for here, I think you can do `entities.join().collect()`.

PR #481, opened 2018-09-24
Add missing joins

Xaeroxe:
A few of these were missing in the book, this adds them.

---
This change is [Reviewable](https://reviewable.io/reviews/slide-rs/specs/481)

### DIFF
--- a/book/src/11_advanced_component.md
+++ b/book/src/11_advanced_component.md
@@ -42,11 +42,11 @@ impl<'a> System<'a> for Sys {
     
     fn run(&mut self, (drag, velocity, mut position): Self::SystemData) {
         // Update positions with drag
-        for (pos, vel, _) in (&mut position, &velocity, &drag) {
+        for (pos, vel, _) in (&mut position, &velocity, &drag).join() {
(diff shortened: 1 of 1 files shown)


Xaeroxe:
Seems we've got some extraneous failures and I can't override.  @torkleyy can you help me get this merged?  Do you have that power?

torkleyy:
@Xaeroxe Ah, thanks for these book improvements. Forgive my irregular visits to GitHub, I've disabled e-mail notifications since at one point they simply got too distracting.

Whenever you do these overrides, please make sure that there's no (unreported) bug in the nightly and, more importantly, bors isn't running for this repo. If it is, you might cause a crash and / or break our master branch. That's why most of the time I prefer to simply retry merging with bors, since that works best with the system (and there's no hurry, so yeah).

I'm actually surprised you're still that active here, I recall that you started building your game with another engine?

Xaeroxe:
I'm definitely not as active as I used to be, but I still have fun working
on this project so I provide small contributions here and there when I feel
like it.  Basically I'd like to use Amethyst for my indie game development
but I also recognize it's not quite what I need it to be yet.

Issue #482, opened 2018-09-26
Deprecate `Bundle`


torkleyy:
`Bundle` is hardly a useful or necessary trait and I don't see any point in maintaining it. As was pointed out in [an IRC chat](https://botbot.me/mozilla/rust-gamedev/2018-09-10/?tz=America/Los_Angeles), there already is a trait for this: `Fn(World)`.

Xaeroxe:
To be more precise: `FnOnce(&mut World)`.  But yeah

Aceeri:
It also would probably just be generally better for libraries to provide a `setup` method of their own rather than us trying to bring all those dependencies together.

torkleyy:
A clarification in the book would be nice.

ldesgoui:
I can cook something up for the setup chapter today-ish

I've noticed that all(?) examples don't call `Dispatcher.setup()` meaning the Resources/Components they require are not registered automatically. As far as I can tell, that's possibly because `main()` always creates a few entities before running the dispatcher, but that could be fixed by having `Dispatcher.setup()` before creating entities, then running it afterwards. This would remove the need to register components and resources.
As for registering systems in bulk, it was mentioned that a simple standalone function would do the job, and libraries could implement their own trait if they so desire

torkleyy:
Yes, exactly :+1: A PR to add this to the book would be tops!

ldesgoui:
I believe this can be closed now :)

PR #486, opened 2018-10-01
Deprecate world::Bundle

ldesgoui:
Hello,

`world::World::exec` can be used instead of implementing `world::Bundle`

Related issue: #482 

Thanks

---
This change is [Reviewable](https://reviewable.io/reviews/slide-rs/specs/486)


ldesgoui:
Should I also write the `CHANGELOG.md` changes?

ldesgoui:
Woops, I wasn't aware Rust had a deprecation feature. Will do

ldesgoui:
Along with the deprecation message, I'd like to add a useful message showing a way to use `FnOnce(&mut world::World)` instead, should I:
- Include everything in the #[deprecate]
- Write the example here and link it in the message
- Add a small segment to the book (not mentioning Bundle) and link it in the message

torkleyy:
Libraries should mostly use `System::setup` and optionally a standalone `setup` function. Using an extra trait for this (predefined or not) is mostly overkill IMO. Since this explanation is too long to be fully included in the deprecation message, I think linking a new book section is the best option.

PR #489, opened 2018-10-02
FlaggedStorage changes

Aceeri:
Related issue: https://github.com/amethyst/amethyst/issues/978

Uses a `ComponentEvent` enum instead of 3 separate channels.

---
This change is [Reviewable](https://reviewable.io/reviews/slide-rs/specs/489)

### DIFF
--- a/examples/track.rs
+++ b/examples/track.rs
@@ -12,8 +12,10 @@ impl Component for TrackedComponent {
 
 #[derive(Default)]
 struct SysA {
-    modified_id: Option<ReaderId<ModifiedFlag>>,
+    reader_id: Option<ReaderId<ComponentEvent>>,
+    inserted: BitSet,
     modified: BitSet,
+    removed: BitSet,
 }
 
 impl<'a> System<'a> for SysA {
@@ -21,11 +23,16 @@ impl<'a> System<'a> for SysA {
 
     fn setup(&mut self, res: &mut Resources) {
         Self::SystemData::setup(res);
-        self.modified_id = Some(WriteStorage::<TrackedComponent>::fetch(&res).track_modified());
+        self.reader_id = Some(WriteStorage::<TrackedComponent>::fetch(&res).track());
     }
 
     fn run(&mut self, (entities, tracked): Self::SystemData) {
-        tracked.populate_modified(&mut self.modified_id.as_mut().unwrap(), &mut self.modified);
+        self.modified.clear();
+        let events = tracked.channel().read(self.reader_id.as_mut().expect("ReaderId not found"));
+        self.modified.extend(events.filter_map(|event| match event {
+            ComponentEvent::Modified(index) => Some(index),
+            _ => None,
+        }));
 
         for (entity, _tracked, _) in (&entities, &tracked, &self.modified).join() {
             println!("modified: {:?}", entity);
--- a/examples/track.rs
+++ b/examples/track.rs
@@ -23,7 +23,7 @@ impl<'a> System<'a> for SysA {
 
     fn setup(&mut self, res: &mut Resources) {
         Self::SystemData::setup(res);
-        self.reader_id = Some(WriteStorage::<TrackedComponent>::fetch(&res).track());
+        self.reader_id = Some(WriteStorage::<TrackedComponent>::fetch(&res).register_reader());
     }
 
     fn run(&mut self, (entities, tracked): Self::SystemData) {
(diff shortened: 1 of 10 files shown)


examples/track.rs
```diff
@@ -12,20 +12,27 @@ impl Component for TrackedComponent {
 
 #[derive(Default)]
 struct SysA {
-    modified_id: Option<ReaderId<ModifiedFlag>>,
+    reader_id: Option<ReaderId<ComponentEvent>>,
+    inserted: BitSet,
     modified: BitSet,
+    removed: BitSet,
```
jojolepro: inserted and removed are not used.

examples/track.rs
```diff
@@ -12,24 +12,53 @@ impl Component for TrackedComponent {
...
 
     fn run(&mut self, (entities, tracked): Self::SystemData) {
-        tracked.populate_modified(&mut self.modified_id.as_mut().unwrap(), &mut self.modified);
+        self.events  = tracked.channel().read(self.reader_id.as_mut().expect("ReaderId not found")).map(|e| *e).collect();
+
+        self.modified.clear();
+        self.modified.extend(self.events.iter().filter_map(|event| match event {
+            ComponentEvent::Modified(index) => Some(index),
+            _ => None,
+        }));
```
WaDelma: These should probably be made into helper methods for `ComponentEvent`:
```rust
fn modified(&self) -> Option<Index>
```

WaDelma: Actually would it be better to iterate events only once and insert to each different `BitSet`?

examples/track.rs
```diff
@@ -57,13 +86,20 @@ fn main() {
 
     dispatcher.setup(&mut world.res);
 
-    for _ in 0..10000 {
+    for _ in 0..50 {
```
WaDelma: Was this reduced because it was too slow?

examples/track.rs
```diff
@@ -57,13 +86,20 @@ fn main() {
...
         world.create_entity().with(TrackedComponent(0)).build();
     }
 
     dispatcher.dispatch(&mut world.res);
     world.maintain();
 
+    let entities = (&world.entities(), &world.read_storage::<TrackedComponent>()).join().map(|(e, _)| e).collect::<Vec<Entity>>();
```
WaDelma: This line is kinda gnarly, but I cannot think how to make it better.

examples/track.rs
Aceeri: no, mainly for readability

examples/track.rs
jojolepro: .map(|t| t.0).collect::<Vec<_>>()

jojolepro: or you can keep the "e" and just change the collect

examples/track.rs
Aceeri: I'd rather just do the latter, the "helper" methods might be kind of confusing.

examples/track.rs
Aceeri: eh, I think it is clearer that it is de-structuring the tuple of the join and it only wants the entities. t.0 isn't very clear. If anything I'll just move the entities/read_storage to a line above it.

src/storage/flagged.rs
```diff
@@ -60,16 +61,21 @@ use world::{Component, Index};
...
-///         // This will only include inserted components from last read, note that this
-///         // will not include `insert` calls if there already was a pre-existing component.
-///         comps.populate_inserted(&mut self.inserted_id, &mut self.inserted);
+///         // Here we can populate the bitsets by iterating over the events.
+///         // You can also just iterate over the events without using a bitset, but this
+///         // allows us to use them in joins with components.
+///         {
+///             let events = comps.channel()
+///                 .read(&mut self.reader_id)
+///                 .collect::<Vec<&ComponentEvent>>();
```
WaDelma: Does this really need collect to a `Vec`? Seems odd that it cannot be used directly as iterator.

Aceeri: I suppose it could, I was just thinking of the previous time when I had to collect it to use it multiple times.

src/storage/track.rs
```diff
@@ -107,85 +47,20 @@ where
...
+    /// components.
+    pub fn channel_mut(&mut self) -> &mut EventChannel<ComponentEvent> {
+        unsafe { self.open() }.1.channel_mut()
     }
 
-    /// Returns the event channel tracking modified components mutably.
-    pub fn modified_mut(&mut self) -> &mut EventChannel<ModifiedFlag> {
-        &mut self.channels_mut().modify
+    /// Starts tracking component events.
+    pub fn track(&mut self) -> ReaderId<ComponentEvent> {
```
torkleyy: Could you call this `register_reader` or similar?

torkleyy:
Meh, conflict again :(

torkleyy:
bors r=WaDelma

Thanks for all the reviews!

PR #492, opened 2018-10-14
Added some comments to the saveload example

Telzhaak:
I hope they are all correct,
please verify.

### DIFF
--- a/examples/saveload.rs
+++ b/examples/saveload.rs
@@ -9,6 +9,11 @@ use specs::error::NoError;
 use specs::prelude::*;
 use specs::saveload::{DeserializeComponents, MarkedBuilder, SerializeComponents, U64Marker, U64MarkerAllocator};
 
+// This is an example of how the serialized data of two entities might look on disk.
+//
+// When serializing entities, they are written in an array of unnamed structs, each struct representing one entity.
+// The entity's marker and components are written as fields into these structs, knowing nothing about the original entity's id.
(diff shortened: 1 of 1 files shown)


examples/saveload.rs
```diff
@@ -9,6 +9,10 @@ use specs::error::NoError;
 use specs::prelude::*;
 use specs::saveload::{DeserializeComponents, MarkedBuilder, SerializeComponents, U64Marker, U64MarkerAllocator};
 
+// This is an example of how the serialized data of two entities might look on disk.
+//
+// When serializing entities, they are written in an array of unnamed structs, each struct representing one entity.
```
torkleyy: unnamed structs -> tuples

That's only if you pass a tuple of component storages though. Custom structs are possible, and will be easier with the derive PR.

examples/saveload.rs
```diff
@@ -64,12 +75,14 @@ impl fmt::Display for Combined {
...
 }
 
+// This returns the `ron::ser:Error` in form of the `Combined` enum, which can then be matched and displayed accordingly.
 impl From<ron::ser::Error> for Combined {
     fn from(x: ron::ser::Error) -> Self {
         Combined::Ron(x)
     }
 }
 
+// This returns nothing, in case a `NoError` is received.
```
torkleyy: This cannot be called, thus it cannot return.

Telzhaak:
I tried, but broke everything locally. So I decided to be save and do it here

PR #495, opened 2018-10-18
Make it possible to disable parallel feature

omni-viral:
This makes it possible to user specs in wasm environment

---
This change is [Reviewable](https://reviewable.io/reviews/slide-rs/specs/495)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,14 +24,14 @@ travis-ci = { repository = "slide-rs/specs" }
 crossbeam = "0.4.1"
 derivative = "1"
 fnv = "1.0"
-hibitset = { version = "0.5.2", features = ["parallel"] }
+hibitset = { version = "0.5.2", default-features = false }
 log = "0.4"
 mopa = "0.2"
-shred = "0.7.0"
+shred = { version = "0.7.0", default-features = false }
 shrev = "1.0.0"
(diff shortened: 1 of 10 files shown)


omni-viral:
> threading support in wasm

Does wasm support threading at all?

torkleyy:
Well, not "enough" yet. There are some experimental things supported only
in Chrome I believe, but for the most part it's currently unsupported.

On Oct 18, 2018 16:07, "Zakarum" <notifications@github.com> wrote:

threading support in wasm

Does wasm support threading at all?

—
You are receiving this because you commented.

Reply to this email directly, view it on GitHub
<https://github.com/slide-rs/specs/pull/495#issuecomment-431022621>, or mute
the thread

Xaeroxe:
Oh just re-read @torkleyy 's comment my bad

bors r-

omni-viral:
Without making code super-inconsistent I managed to reduce amount of `#[cfg(feature = "parallel")]` slightly.

torkleyy:
Wow.

*should we care about

omni-viral:
Emmm. I checked it with wasm :)

drager:
I can confirm this works. I tried out @omni-viral's branch with the same code that wasn't working before this PR and now it works.

Issue #508, opened 2018-10-30
`SerializeComponents::serialize` panics in a really vague way if a marker is invalid


expenses:
Basically, if an component contains an `Entity` but this referenced entity is deleted, then `SerializeComponent::serialize` panics with 

```
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', libcore/option.rs:345:21
stack backtrace:
   0: std::sys::unix::backtrace::tracing::imp::unwind_backtrace
             at libstd/sys/unix/backtrace/tracing/gcc_s.rs:49
   1: std::sys_common::backtrace::print
             at libstd/sys_common/backtrace.rs:71
             at libstd/sys_common/backtrace.rs:59
   2: std::panicking::default_hook::{{closure}}
             at libstd/panicking.rs:211
   3: std::panicking::default_hook
             at libstd/panicking.rs:227
   4: <std::panicking::begin_panic::PanicPayload<A> as core::panic::BoxMeUp>::get
             at libstd/panicking.rs:477
   5: std::panicking::continue_panic_fmt
             at libstd/panicking.rs:391
   6: std::panicking::try::do_call
             at libstd/panicking.rs:326
   7: core::ptr::drop_in_place
             at libcore/panicking.rs:77
   8: core::ptr::drop_in_place
             at libcore/panicking.rs:52
   9: <core::option::Option<T>>::unwrap
             at libcore/macros.rs:20
  10: <specs::world::entity::Entity as specs::saveload::ConvertSaveload<M>>::convert_into
             at /Users/ashley/.cargo/git/checkouts/specs-18d2bb2ce407da15/b22955b/src/saveload/mod.rs:175
  11: <specs_ron_bug::Parent as specs::saveload::ConvertSaveload<MA>>::convert_into
             at src/main.rs:11
  12: <(SP,) as specs::saveload::ser::SerializeComponents<E, M>>::serialize_entity::{{closure}}
             at /Users/ashley/.cargo/git/checkouts/specs-18d2bb2ce407da15/b22955b/src/saveload/ser.rs:128
  13: <core::option::Option<T>>::map
             at libcore/option.rs:414
  14: <(SP,) as specs::saveload::ser::SerializeComponents<E, M>>::serialize_entity
             at /Users/ashley/.cargo/git/checkouts/specs-18d2bb2ce407da15/b22955b/src/saveload/ser.rs:128
  15: specs::saveload::ser::SerializeComponents::serialize
[...]

PR #509, opened 2018-10-30
added better entity serialization errors

expenses:
See #508. These errors are a work in progress, if you've got a better idea please suggest it!

---
This change is [Reviewable](https://reviewable.io/reviews/slide-rs/specs/509)

### DIFF
--- a/specs-derive/src/impl_saveload.rs
+++ b/specs-derive/src/impl_saveload.rs
@@ -72,7 +72,7 @@ pub fn impl_saveload(ast: &mut DeriveInput) -> TokenStream {
 
         impl #impl_generics ConvertSaveload<MA> for #name #ty_generics #where_clause {
             type Data = #saveload_name #saveload_ty_generics;
-            type Error = NoError;
+            type Error = Error;
 
             fn convert_into<F>(&self, mut ids: F) -> Result<Self::Data, Self::Error>
             where
--- a/src/error.rs
+++ b/src/error.rs
@@ -59,6 +59,10 @@ pub enum Error {
     Custom(BoxedErr),
     /// Wrong generation error.
     WrongGeneration(WrongGeneration),
+    /// No marker was found for entity.
+    NoMarker,
+    /// No entity was found for marker
+    NoEntity,
 
     #[doc(hidden)]
     __NonExhaustive,
@@ -69,7 +73,8 @@ impl Display for Error {
         match *self {
             Error::Custom(ref e) => write!(f, "Custom: {}", e),
             Error::WrongGeneration(ref e) => write!(f, "Wrong generation: {}", e),
-
+            Error::NoMarker => write!(f, "No marker found for entity; has the entity been deleted?"),
+            Error::NoEntity => write!(f, "No entity found for marker; has the entity been deleted?"),
             Error::__NonExhaustive => unimplemented!(),
         }
     }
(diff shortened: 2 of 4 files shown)


src/saveload/mod.rs
```diff
@@ -172,13 +172,13 @@ where
     where
         F: FnMut(Entity) -> Option<M>,
     {
-        Ok(func(*self).unwrap())
+        Ok(func(*self).unwrap_or_else(|| panic!("No marker found for entity; has the entity been deleted?")))
```
LucioFranco: I believe you can use `.expect`

expenses: Oh yeah true, I'll do that in a sec

expenses:
@torkleyy yeah I was thinking that.

torkleyy:
Could you add a test to make sure we don't get a panic anymore?
Showing how to handle the error in the saveload example would also be useful.

I can assist you better in about 4 hours, but I suspect that doesn't work well with your time zone ^^'

specs-derive/src/impl_saveload.rs
```diff
@@ -72,7 +72,7 @@ pub fn impl_saveload(ast: &mut DeriveInput) -> TokenStream {
 
         impl #impl_generics ConvertSaveload<MA> for #name #ty_generics #where_clause {
             type Data = #saveload_name #saveload_ty_generics;
-            type Error = NoError;
+            type Error = Error;
```
expenses: This _might_ break some things, but the tests pass 🤷‍♀️

src/error.rs
```diff
@@ -96,14 +101,21 @@ impl StdError for Error {
...
+            Error::NoMarker => unimplemented!(),
+            Error::NoEntity => unimplemented!(),
             Error::__NonExhaustive => unimplemented!(),
         };
 
         Some(e)
     }
 }
 
+impl PartialEq for Error {
```
expenses: This is really hacky, but I want sure if I could impl PartialEq for BoxedErr.

tests/saveload.rs
```diff
@@ -69,7 +74,52 @@ mod tests {
         // The derive will work for all variants
         // so no need to test anything but unit
         black_box::<U64Marker, _>(AnEnum::Unit);
-        black_box::<U64Marker, _>(Generic(entity));
+        //black_box::<U64Marker, _>(Generic(entity));
```
expenses: I had to disable this black box test because of an error:
```
error[E0277]: the trait bound `tests::Generic<spocs::Entity>: spocs::saveload::ConvertSaveload<spocs::saveload::U64Marker>` is not satisfied
  --> tests/saveload.rs:77:9
   |
77 |         black_box::<U64Marker, _>(Generic(entity));
   |         ^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `spocs::saveload::ConvertSaveload<spocs::saveload::U64Marker>` is not implemented for `tests::Generic<spocs::Entity>`
   |
   = help: the following implementations were found:
             <tests::Generic<E> as spocs::saveload::ConvertSaveload<MA>>
note: required by `tests::black_box`
  --> tests/saveload.rs:125:5
   |
125|     fn black_box<M, T: ConvertSaveload<M>>(_item: T) {}
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

src/error.rs
```diff
@@ -96,14 +101,21 @@ impl StdError for Error {
         let e = match *self {
             Error::Custom(ref e) => e.as_ref(),
             Error::WrongGeneration(ref e) => e,
-
+            Error::NoMarker => unimplemented!(),
```
torkleyy: Needs to be changed before merge

src/error.rs
torkleyy: The `BoxedErr` variant should just always return `false`, just avoid implemeting `Eq`.

tests/saveload.rs
torkleyy: Hmm, this needs investigations, this shouldn't happen.

torkleyy:
@expenses I'm going to close this for now since it's gone stale, but in case you get to finishing it, please reopen it!

PR #515, opened 2018-11-22
Allow accessing fetched entities of a `Storage`

torkleyy:
I've added `Storage::entities` to allow making extension methods for `Storage` that return `Entity`.

This method is hidden because I'm 100% certain newcomers will be confused by this otherwise. This makes it less visible, yes, but I think that's worth for avoiding this confusion.

---
This change is [Reviewable](https://reviewable.io/reviews/slide-rs/specs/515)

### DIFF
--- a/src/storage/mod.rs
+++ b/src/storage/mod.rs
@@ -202,6 +202,14 @@ where
         &self.data.inner
     }
 
+    /// Returns the `EntitesRes` resource fetched by this storage.
+    /// **This does not have anything to do with the components inside.**
+    /// You only want to use this when implementing additional methods
+    /// for `Storage` via an extension trait.
+    pub fn fetched_entities(&self) -> &EntitiesRes {
+        &self.entities
+    }
+
     /// Tries to read the data associated with an `Entity`.
     pub fn get(&self, e: Entity) -> Option<&T> {
         if self.data.mask.contains(e.id()) && self.entities.is_alive(e) {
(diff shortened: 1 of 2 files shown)


src/storage/mod.rs
```diff
@@ -202,6 +202,15 @@ where
         &self.data.inner
     }
 
+    /// Returns the `EntitesRes` resource fetched by this storage.
+    /// **This does not have anything to do with the components inside.**
+    /// You only want to use this when implementing additional methods
+    /// for `Storage` via an extension trait.
+    #[doc(hidden)]
+    pub fn entitites(&self) -> &EntitiesRes {
```
jojolepro: s/entitites/entities

Xaeroxe:
I don't think it's worth it to hide this without demonstrated evidence it confuses people.  I think we should leave it visible until we get a lot of confused users.

torkleyy:
Hmm, I thought you're going to say something like this ^^
The problem is simply that users tend to use features that are more complicated than what they actually need. The fact that a storage fetches the entities resource is not known to many users, that's why I'm hiding it.

That's why I hid it. If you insist on this, the only option I see is to give it a longer name to make sure this is obvious.

Xaeroxe:
I'd definitely be in favor of the longer name option.  This way it would still come up in editor auto completes but there would be no documentation.  That's even more confusing.

PR #518, opened 2018-12-09
Add Join-able entries API to Storage

jjedelsky:
This makes it possible to read and insert component at same time inside join iteration.

`Storage::entries` would likely need more docs, however I can not find any really minimal example.

---
This change is [Reviewable](https://reviewable.io/reviews/slide-rs/specs/518)

### DIFF
--- a/src/storage/entry.rs
+++ b/src/storage/entry.rs
@@ -171,7 +171,7 @@ where
 /// Entry to a storage for convenient filling of components or removal based on whether
 /// the entity has a component.
 pub enum StorageEntry<'a, 'b: 'a, T: 'a, D: 'a> {
-    /// Entry variant that is returned if the entity does has a component.
+    /// Entry variant that is returned if the entity has a component.
     Occupied(OccupiedEntry<'a, 'b, T, D>),
     /// Entry variant that is returned if the entity does not have a component.
     Vacant(VacantEntry<'a, 'b, T, D>),
@@ -182,7 +182,7 @@ where
     T: Component,
     D: DerefMut<Target = MaskedStorage<T>>,
 {
-    /// Inserts a component if the entity does not contain a component.
+    /// Inserts a component if the entity does not have it already.
     pub fn or_insert(self, component: T) -> &'a mut T {
         self.or_insert_with(|| component)
     }
--- a/src/storage/entry.rs
+++ b/src/storage/entry.rs
@@ -188,7 +188,9 @@ where
     }
 
     /// Inserts a component using a lazily called function that is only called
-    /// when inserting the component.
+    /// when inserting the component. Ensures this entry has a value and if not, 
+    /// inserts one using the result of the passed closure. Returns a reference 
+    /// to the value afterwards.
     pub fn or_insert_with<F>(self, default: F) -> &'a mut T
     where
         F: FnOnce() -> T,
--- a/src/storage/entry.rs
+++ b/src/storage/entry.rs
@@ -1,0 +1,201 @@
+use super::*;
+use crate::join::Join;
+use hibitset::BitSetAll;
+
+impl<'e, T, D> Storage<'e, T, D>
+where
+    T: Component,
+    D: DerefMut<Target = MaskedStorage<T>>,
+{
+    /// Returns an entry to the component associated to the entity.
+    ///
+    /// Behaves somewhat similarly to `std::collections::HashMap`'s entry api.
+    ///
+    /// ## Example
+    ///
+    /// ```rust
+    /// # extern crate specs;
+    /// # use specs::prelude::*;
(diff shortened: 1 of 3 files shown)


src/storage/entry.rs
```diff
@@ -0,0 +1,201 @@
...
+    D: DerefMut<Target = MaskedStorage<T>>,
+{
+    /// Get a mutable reference to the component associated with the entity.
+    pub fn get_mut(&mut self) -> &mut T {
+        unsafe { self.storage.data.inner.get_mut(self.id) }
+    }
+
+    /// Converts the `OccupiedEntry` into a mutable reference bounded by
+    /// the storage's lifetime.
+    pub fn into_mut(self) -> &'a mut T {
```
WaDelma: This looks suspicious. Wouldn't this let you peddle the reference to a variable which can then be accessed in a join? Or is there something I am missing here.

src/storage/entry.rs
```diff
@@ -0,0 +1,201 @@
...
+            self.storage.data.inner.insert(self.id, component);
+            self.storage.data.inner.get_mut(self.id)
+        }
+    }
+}
+
+/// Entry to a storage for convenient filling of components or removal based on whether
+/// the entity has a component.
+pub enum StorageEntry<'a, 'b: 'a, T: 'a, D: 'a> {
+    /// Entry variant that is returned if the entity does has a component.
```
WaDelma:
```suggestion
    /// Entry variant that is returned if the entity has a component.
```

src/storage/entry.rs
```diff
@@ -0,0 +1,201 @@
...
+    /// Entry variant that is returned if the entity does not have a component.
+    Vacant(VacantEntry<'a, 'b, T, D>),
+}
+
+impl<'a, 'b, T, D> StorageEntry<'a, 'b, T, D>
+where
+    T: Component,
+    D: DerefMut<Target = MaskedStorage<T>>,
+{
+    /// Inserts a component if the entity does not contain a component.
```
WaDelma:
```suggestion
    /// Inserts a component if the entity does not have it already.
```

src/storage/entry.rs
```diff
@@ -0,0 +1,201 @@
...
+    pub fn or_insert(self, component: T) -> &'a mut T {
+        self.or_insert_with(|| component)
+    }
+
+    /// Inserts a component using a lazily called function that is only called
+    /// when inserting the component.
```
WaDelma: I am not sure how to improve this, but it sounds wrong (to me)

src/storage/entry.rs
WaDelma: Ah it was there before. It would be good to have some explanation why it's safe.

torkleyy:
> Should this have warning that it iterates through all entities and should be joined with other components?

I think that should be addressed by the `is_constrained` check, but we should better check.

src/storage/entry.rs
```diff
@@ -0,0 +1,201 @@
...
+                action: "attempting to get an entry to a storage",
+                actual_gen: gen,
+                entity: e,
+            })
+        }
+    }
+
+    /// Returns a `Join`-able structure that yields all indices, returning
+    /// `Entry` for all elements
+    pub fn entries<'a>(&'a mut self) -> Entries<'a, 'e, T, D> {
```
torkleyy: Example code would be nice to illustrate that you should join this with another storage.

src/storage/entry.rs
torkleyy: "Ensures this entry has a value and if not, inserts one using the result of the passed closure. Returns a reference to the value afterwards."

torkleyy:
Could you add tests for this?

Maybe one checking the general functionality, and another one for the unconstrained "warning"?

src/storage/entry.rs
jjedelsky: Full signature for this function is `pub fn into_mut(self: OccupiedEntry<'a, 'b, T, D>) -> &'a mut T`,
so the returned reference can not outlive the entry which can not outlive Entries which can not outlive Storage.
References are mutable, so rust won't allow you to create 2 Entries to the same storage.
However you could misuse __unsafe__ Join::get to create `StorageEntry` more than once, but:
- it is possible for any Join over mutable data
- and Join::get is __unsafe__

jjedelsky:
I have added test for general functionality. Is there any test checking for unconstrained warning for `MaybeJoin`? I am not sure how to test it cross-platform, because the warning is written to stdout.

src/storage/entry.rs
jjedelsky: Updated

src/storage/entry.rs
jjedelsky: Added

PR #522, opened 2018-12-26
[DO NOT MERGE] Updating resources

KuSpa:
There is currently no API (at least in `World`) for checking if a resource already exists. Fetching may result in a panic, which is bad or a `Default`, which adds boilerplate, because we need to check against the default what would look like this: 
```rust
// we assume the Resource has a Default, so we won't panic
if world.read_resource::<Res>() == Res::default() {
    *world.write_resource() = Res::init();// <- costly initialization like loading from disk or calculate something heavy like 3+4 ^.^
}
```
Using `world.add_resource(Resd::init())` overrides the actual resource if existing, which is bad for obvious reasons.
This makes updating resources less clear, especially in [`State>>on_start`](https://www.amethyst.rs/doc/latest/doc/amethyst/trait.State.html#method.on_start).

---
This change is [Reviewable](https://reviewable.io/reviews/slide-rs/specs/522)

### DIFF
--- a/src/world/mod.rs
+++ b/src/world/mod.rs
@@ -440,6 +440,13 @@ impl World {
         }
     }
 
+    /// Adds a resource to the world, if the resource does not exist yet. See `add_resource` for more information.
+    pub fn add_resource_soft<T: Resource>(&mut self, res: T) {
+        if !self.res.has_value::<T>() {
+            self.add_resource(res);
+        }
+    }
+
     /// Fetches a component's storage for reading.
     ///
     /// ## Panics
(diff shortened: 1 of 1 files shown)


torkleyy:
This is part of the bigger issue of `Resources` / `World`.

There is such an API for `Resources`: [`Resources::entry`](https://docs.rs/specs/0.14.1/specs/prelude/struct.Resources.html#method.entry)

So, given a `World` called `world`: `world.res.entry::<MyResource>()`

torkleyy:
> This is part of the bigger issue of Resources / World.

To elaborate, `World` is basically an extension of `Resources`. I feel this makes the API a bit inconsistent because

1) you cannot use the `World` API if you have just `Resources`
2) you have both the `World` and the `Resources` API to deal with

Actually, gonna move this into an issue.

torkleyy:
@KuSpa Can we close this now or is there anything else left you want to address?

torkleyy:
Closing this for now. Feel free to re-open if you think this should be left open.

PR #529, opened 2019-01-03
Add `Storage::with_mut`

### LINKED ISSUE #336: Add way of getting multiple mutable references from storages
Add API that allows getting multiple mutable references from a storage in a way that preserves correctness.
If we had integer generics we could provide generic API:
```rust
fn get_muts<const N: usize>(&mut self, [Entity; N]) -> [C; N];
```
Right now we could provide more specific instances of that.

The API should be restricted to `DistinctStorage`s.

torkleyy:
Fixes #336

---
This change is [Reviewable](https://reviewable.io/reviews/slide-rs/specs/529)

### DIFF
--- a/src/storage/mod.rs
+++ b/src/storage/mod.rs
@@ -255,6 +255,152 @@ where
         }
     }
 
+    /// Tries to retrieve `e` from the storage and passes a mutable reference to `and_then`.
+    /// This is useful for accessing multiple components mutably.
+    ///
+    /// # Parameters
+    ///
+    /// * `e`: The `Entity` you want to retrieve the component for. If it is not alive (was deleted)
+    ///   or there is no component for `e` in this storage, `None` will be passed to `and_then`.
+    /// * `and_then`: A closure executed independently of the existence of the component. It will
+    ///   be passed `self` as a first argument, and an optional reference to the component (in
+    ///   case it existed and is alive). The closure may return a value of type `R`, which will
+    ///   be returned from this method.
+    ///
+    /// # How this works
+    ///
+    /// This implementation removes and re-inserts the component at `e` to prove to the compiler
+    /// the accessed components are distinct (or you will get a `None`).
+    ///
+    /// # Recommended usage
+    ///
+    /// Check entities for equality prior to using this method; that way, if you get `None`, the
+    /// error is limited to a missing component.
+    ///
+    /// # Examples
+    ///
+    /// ```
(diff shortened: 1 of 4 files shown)


torkleyy:
Will add convenience methods in a follow up.

src/storage/mod.rs
```diff
@@ -255,6 +255,152 @@ where
         }
     }
 
+    /// Tries to retrieve `e` from the storage and passes a mutable reference to `and_then`.
```
WaDelma: I was really confused first by both `e` and `and_then` when I first read this. Mostly because I didn't read the function signature. If were going to use them in documentation I would rename `e` to `entity`. (`and_then` was confusing because it's name used for `Option`s `flat_map`/`bind`)

src/storage/mod.rs
```diff
@@ -255,6 +255,152 @@ where
...
+    ///
+    /// # How this works
+    ///
+    /// This implementation removes and re-inserts the component at `e` to prove to the compiler
+    /// the accessed components are distinct (or you will get a `None`).
+    ///
+    /// # Recommended usage
+    ///
+    /// Check entities for equality prior to using this method; that way, if you get `None`, the
+    /// error is limited to a missing component.
```
WaDelma: s/equality/inequality/

src/storage/mod.rs
WaDelma: This is probably not that huge of a deal, but somehow I was confused by it.

torkleyy:
> I am kinda wondering about the semantics. Is there use case for calling the closure if the component is missing? Removing the call should simplify the example.

Are you saying we should not call `f` at all if it's missing? That might actually be a good idea, yes...

torkleyy:
I think the drawback of not calling `f` would be that convenience methods fetching multiple components at once would need to branch on every single entity.

src/storage/mod.rs
```diff
@@ -255,6 +255,186 @@ where
         }
     }
 
+    /// Tries to retrieve `e` from the storage and passes a mutable reference to `and_then`.
```
WaDelma:
```suggestion
    /// Tries to retrieve `entity` from the storage and passes a mutable reference to `f`.
```

torkleyy:
What I don't like about this solution:

* does not use the `Entry` API
* requires operating on the components inside the closure
* does not allow nice inserting of missing components; bad usability

PR #531, opened 2019-01-03
Fix `FlaggedStorage` docs

aleksijuvani:
The documentation for `FlaggedStorage` became out of date with #489. I'm not entirely familiar with this API, so please review to see if my understanding of the changes is correct. I have however made sure that the code sample compiles.

The bit about handling `ComponentEvent::Removed` first is no longer necessary because the events are in the same "channel" and in chronological order, right?

---
This change is [Reviewable](https://reviewable.io/reviews/slide-rs/specs/531)

### DIFF
--- a/book/src/12_tracked.md
+++ b/book/src/12_tracked.md
@@ -6,9 +6,9 @@ modified.
 
 To avoid a lot of unnecessary computation when updating components it
 would be nice if we could somehow check for only those entities that are updated
-and recalculate only those. 
+and recalculate only those.
 
-We might also need to keep an external resource in sync with changes 
+We might also need to keep an external resource in sync with changes
 to components in Specs `World`, and we only want to propagate actual changes, not
 do a full sync every frame.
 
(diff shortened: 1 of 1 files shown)


aleksijuvani:
cc @torkleyy

book/src/12_tracked.md
```diff
@@ -30,73 +30,52 @@ impl Component for Data {
...
+        let events = data.channel().read(self.reader_id.as_mut().unwrap());
+
+        // Note that we could use separate bitsets here, we only use one to
+        // simplify the example
+        for event in events {
+            match event {
+                ComponentEvent::Modified(id) | ComponentEvent::Inserted(id) => {
+                    self.dirty.add(*id);
+                }
+                ComponentEvent::Removed(_) => (),
```
torkleyy: I think another comment here would be good:

```suggestion
                 // We don't need to take this event into account since
                 // removed components will be filtered out by the join;
                 // if you want to, you can use `self.dirty.remove(*id);`
                 // so the bit set only contains IDs that still exist
                 ComponentEvent::Removed(_) => (),
```

torkleyy:
> The bit about handling ComponentEvent::Removed first is no longer necessary because the events are in the same "channel" and in chronological order, right?

Yes, exactly :+1:

Issue #523, opened 2018-12-26
Make `World` an extension of `Resources`


torkleyy:
`World` is basically an extension of `Resources`. I feel this makes the API a bit inconsistent because

1) you cannot use the `World` API if you have just `Resources`
2) you have both the `World` and the `Resources` API to deal with

### Trait

Make `World` a trait that is implemented for `Resources`. The `new` function can be implemented

1) using a free function (`new_world() -> Resources`)
2) doing a `impl World { pub fn new() -> Resources }`

### Free standing functions

Provide functions like

* `ǹew_world`
* `read_component<T>(&Resources)`
* etc

torkleyy:
This would be a breaking change unfortunately, although we can limit its impact if we choose the former.

azriel91:
I'd vote for the extension trait, also today I learned that you can have an inherent method implemented on a trait type.

jojolepro:
trait over resources

impl World for Resources

KuSpa:
Just using a Trait without reducing either API, Resources would have `fetch` as well as `read_storage`, which is exactly the same. 
To not to confuse the user, I would remove the wrapping methods or rename Resources' methods. Doing either would heavily impact everyone, who's using the other API.

TimonPost:
Concrete steps to clarify this issue.

1. [ ] Rename `shred::Resources` to `shred::World`
2. [ ] Remove methods from `specs::World` into `specs::WorldExt`
	- read_storage / write_storage should be removed completely.

torkleyy:
This is still relevant.

torkleyy:
Fixes by #550

PR #550, opened 2019-01-22
Refactor of `specs::World`

TimonPost:
This PR is a fix for #523. I did an other [related PR](https://github.com/slide-rs/shred/pull/109) in 'shred' to rename `shred::Resources` to `shred::World`. This is also the reason why I am using a git link in 'Cargo.toml' as 'shred' dependency.

`specs::World` is basically an extension of `Resources`, this means that it is basically, a wrapper calling underlying methods of `shred::Resources`.

A better scenario would be to have a trait with methods, currently in `specs::World` so that we can implement this trait for `shred::Resources`. 

We did the following changes to make this API better:

Refactor 'shred'
- Rename `shred::Resources` to `shred::World`

Refactored 'specs'.
- Remove `World` struct.
- Moved all `World` method definitions into `WorldExt` trait.
- Implemented `WorldExt` for `shred::World`
- Removed the methods `read_storage / write_storage` since they are not needed.

```
struct shred::World {}
````
```
trait WorldExt {
     // method definitions of 'specs::World' except  `read_storage / write_storage` 
}

impl WorldExt for shred::World {
   // all methods of current 'specs::World' except from 'read_storage / write_storage'
}
```

**Note that this PR has breaking API changes**

Thinks you should be checking for are incorrect documentations since some renaming is done on both `shred::Resource` and `specs::World`

---
This change is [Reviewable](https://reviewable.io/reviews/slide-rs/specs/550)


torkleyy:
Thank you for all the work! Fixed the last error :)
bors r+

PR #562, opened 2019-03-07
implement get_mut_or_default

### LINKED ISSUE #561: Implement GenericWriteStorage::get_mut_or_default
## Description

Implement `GenericWriteStorage::get_mut_or_default`, which either fetches an existing component, or calls `Default::default` on the component type, inserts the newly created struct in to the storage, and returns a mutable reference to it.

This can never fail.

## Motivation

I found myself writing a lot of code along the lines of

```rust
if let Some(stun) = data.stun.get_mut(entity) {
    stun.time += 1;
} else {
    data.stun.insert(entity, Stun { time: 1 }).unwrap();
}
```

Verbose, and extremely repetitive. The new function offers a shorthand:

```rust
data.stun.get_mut_or_default(entity).time += 1;
```

## Drawbacks

An extra function to maintain. However, it builds on the existing methods of `GenericWriteStorage`, so the maintenance impact should be near zero.`

## Unresolved questions

---

Opened a PR for this.

JaniM:
Fixes #561

## Checklist

* [ ] I've added a demonstration of the new feature to one or more examples (didn't find any relevant examples)

## API changes

New function: `GenericWriteStorage::get_mut_or_default` for Default component types. Not breaking.

---
This change is [Reviewable](https://reviewable.io/reviews/slide-rs/specs/562)

### DIFF
--- a/src/storage/generic.rs
+++ b/src/storage/generic.rs
@@ -83,6 +83,13 @@ pub trait GenericWriteStorage {
     /// Get mutable access to an `Entity`s component
     fn get_mut(&mut self, entity: Entity) -> Option<&mut Self::Component>;
 
+    /// Get mutable access to an `Entity`s component. If the component does not exist, it
+    /// is automatically created using `Default::default()`.
+    /// 
+    /// # Panics
+    /// Panics if the entity is dead.
+    fn get_mut_or_default(&mut self, entity: Entity) -> &mut Self::Component where Self::Component: Default;
+
     /// Insert a component for an `Entity`
     fn insert(&mut self, entity: Entity, comp: Self::Component) -> InsertResult<Self::Component>;
 
@@ -101,6 +108,14 @@ where
 
     fn get_mut(&mut self, entity: Entity) -> Option<&mut Self::Component> {
         WriteStorage::get_mut(self, entity)
+    }
+
+    fn get_mut_or_default(&mut self, entity: Entity) -> &mut Self::Component where Self::Component: Default {
+        if !self.contains(entity) {
+            self.insert(entity, Default::default()).expect("Insertion of default component failed");
+        }
+        // At this point it is reasonably safe to assume the component exists.
+        self.get_mut(entity).expect("Fetching a default component failed")
     }
 
     fn insert(&mut self, entity: Entity, comp: Self::Component) -> InsertResult<Self::Component> {
@@ -124,6 +139,14 @@ where
 
     fn get_mut(&mut self, entity: Entity) -> Option<&mut Self::Component> {
         WriteStorage::get_mut(*self, entity)
+    }
+
+    fn get_mut_or_default(&mut self, entity: Entity) -> &mut Self::Component where Self::Component: Default {
+        if !self.contains(entity) {
+            self.insert(entity, Default::default()).expect("Insertion of default component failed");
+        }
+        // At this point it is reasonably safe to assume the component exists.
+        self.get_mut(entity).expect("Fetching a default component failed")
     }
 
     fn insert(&mut self, entity: Entity, comp: Self::Component) -> InsertResult<Self::Component> {
--- a/src/storage/tests.rs
+++ b/src/storage/tests.rs
@@ -136,7 +136,7 @@ mod test {
         type Storage = NullStorage<Self>;
     }
 
-    #[derive(PartialEq, Eq, Debug)]
+    #[derive(PartialEq, Eq, Debug, Default)]
     struct Cvec(u32);
     impl From<u32> for Cvec {
         fn from(v: u32) -> Cvec {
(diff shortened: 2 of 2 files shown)


src/storage/generic.rs
```diff
@@ -103,6 +106,13 @@ where
         WriteStorage::get_mut(self, entity)
     }
 
+    fn get_mut_or_default(&mut self, entity: Entity) -> &mut Self::Component where Self::Component: Default {
+        if self.get(entity).is_none() {
+            self.insert(entity, Default::default()).unwrap();
```
Moxinilian: I don't think an unwrap should be best here. Maybe at least an expect.
The one under does not need an except, but maybe an explanation in the comment explaining why it will never be triggered.

JaniM: Is there any situation where the expect could be hit?

Moxinilian: I believe the specs does not guarantee it would never happen. It's not very expensive to make sure it doesn't go wrong.

JaniM: Is the only possible error case that the component has been added between the check and the insertion? In that case we could ignore the result entirely.

Moxinilian: No, this would not even be an error, insert overwrites the previous component.
@torkleyy What do you think?

azriel91: +1 to `expect()` instead of `unwrap()`.

> Is the only possible error case that the component has been added between the check and the insertion?
it's a mutable borrow of `self`, so nothing can insert it between

by the way, [`WriteStorage` is a `Storage`](https://github.com/slide-rs/specs/blob/d6f3c7a82baa864c3c603f699037848362c424f1/src/storage/data.rs#L203), and has a [`contains`](https://github.com/slide-rs/specs/blob/d6f3c7a82baa864c3c603f699037848362c424f1/src/storage/mod.rs#L235) method, so perhaps it expresses the intent better.

```rust
if !self.contains(entity) {
    self.insert(entity, Default::default()).expect("..");
}
self.get_mut(entity).expect("..")
```

<details>
<summary>The alternate implementation feels a bit verbose</summary>

```rust
if let Some(component) = self.get_mut(entity) {
    component
} else {
    self.insert(entity, Default::default()).expect("..");
    self.get_mut(entity).expect("..")
}
```

</details>

Same comment for the below implementation

src/storage/tests.rs
```diff
@@ -295,6 +295,39 @@ mod test {
...
+        let mut w = World::new();
+        let mut s: Storage<T, _> = create(&mut w);
+
+        for i in 0..500 {
+            if let Err(err) = s.insert(Entity::new(i, Generation::new(1)), (i + 1290).into()) {
+                panic!("Failed to insert component into entity! {:?}", err);
+            }
+        }
+
+        for i in 0..1_000 {
```
azriel91: hm, what are these loops and magic numbers? am finding I have to think quite a bit to understand the logic, and then think again to figure out what its intent is.

if the loops are necessary, it's worth writing the rationale in a comment
* is it to get the entity index high enough?
* can we use a smaller backing storage to achieve the same effect?

not obvious to understand what the test is doing 🤔:

1. insert 500 entities with `n + 1290` as its value.
2. mutate 1000 entities with `get_mut_or_default()` to be `n + n + 710`
3. assert the first 500 are `2n + 2000`
4. assert the last 500 are `n + 710`

ah 💡! while it is logically correct and does test the effect, it's considerable mental effort to turn the code back into understanding. could you try to make it easier, otherwise my head is like this: 🤕, haha

---

**edit:** just expanded the diff, and saw the *existing code* is like that. oh man 😂

src/storage/generic.rs
```diff
@@ -83,6 +83,9 @@ pub trait GenericWriteStorage {
     /// Get mutable access to an `Entity`s component
     fn get_mut(&mut self, entity: Entity) -> Option<&mut Self::Component>;
 
+    /// Get mutable access to an `Entity`s component. If the component does not exist, it is automatically created using `Default::default()`.
```
azriel91: rustfmt's gonna cry on this line length, possibly on the next one

JaniM:
Squashed & pushed with relevant modifications.

src/storage/tests.rs
JaniM: I removed the magic numbers & commented it.

src/storage/generic.rs
JaniM: Switched to expect.

src/storage/generic.rs
```diff
@@ -103,6 +111,14 @@ where
         WriteStorage::get_mut(self, entity)
     }
 
+    fn get_mut_or_default(&mut self, entity: Entity) -> &mut Self::Component where Self::Component: Default {
+        if !self.contains(entity) {
+            self.insert(entity, Default::default()).expect("Insertion of default component failed");
```
torkleyy: This will panic for dead entities.

src/storage/generic.rs
```diff
@@ -126,6 +142,14 @@ where
         WriteStorage::get_mut(*self, entity)
     }
 
+    fn get_mut_or_default(&mut self, entity: Entity) -> &mut Self::Component where Self::Component: Default {
+        if !self.contains(entity) {
+            self.insert(entity, Default::default()).expect("Insertion of default component failed");
```
torkleyy: Same here

src/storage/generic.rs
JaniM: Oh boy. I didn't think of that.

JaniM:
Marked that it will panic if the entity is dead. Another option is to return an Option, which'd make the usage slightly less ergonomic, but in most cases still nice enough.

JaniM:
^ Opened a separate PR for that. Pick one.

PR #583, opened 2019-05-02
Big VS Small component benchmarks.

jojolepro:
## Checklist

* [X] I've added tests for all code changes and additions (where applicable)

## API changes
N/A

---
This change is [Reviewable](https://reviewable.io/reviews/slide-rs/specs/583)

### DIFF
--- a/src/world/entity.rs
+++ b/src/world/entity.rs
@@ -1,9 +1,9 @@
 use std::{
     fmt,
-    num::NonZeroI32,
     sync::atomic::{AtomicUsize, Ordering},
 };
 
+use nonzero_signed::NonZeroI32;
 use hibitset::{AtomicBitSet, BitSet, BitSetOr};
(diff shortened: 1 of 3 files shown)


src/world/entity.rs
```diff
@@ -1,9 +1,9 @@
 use std::{
     fmt,
-    num::NonZeroI32,
     sync::atomic::{AtomicUsize, Ordering},
 };
 
+use nonzero_signed::NonZeroI32;
```
torkleyy: I'm confused by this; it's now deprecated since std has support for it.

jojolepro: the import was failing for me

torkleyy: Have you updates your Rust toolchain?

jojolepro: fixed thanks

PR #584, opened 2019-05-02
Uuid Marker

jojolepro:
## Checklist

* [X] I've added tests for all code changes and additions (where applicable)
* [X] Usage of new public items is shown in the API docs

## API changes

Added UuidMarker and UuidMarkerAllocator.
They work the same way their U64 counterparts do.

Added "uuid_entity" feature gate.

---
This change is [Reviewable](https://reviewable.io/reviews/slide-rs/specs/584)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -34,6 +34,7 @@ shred-derive = "0.6.0"
 tuple_utils = "0.3"
 rayon = { version = "1.0.0", optional = true }
 nonzero_signed = "1.0.1"
+uuid = { version = "0.7.4", optional = true, features = ["v4", "serde"] }
 
 serde = { version = "1.0", optional = true, features = ["serde_derive"] }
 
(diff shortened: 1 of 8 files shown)


src/saveload/uuid.rs
```diff
@@ -0,0 +1,98 @@
...
+    }
+
+    /// Get the current uuid.
+    pub fn uuid(&self) -> &Uuid {
+        &self.uuid
+    }
+
+    /// Set a new uuid.
+    pub fn set_uuid(&mut self, uuid: Uuid) {
+        self.uuid = uuid;
```
torkleyy: I don't think we should permit this. Won't this invalidate the mapping?

Cargo.toml
```diff
@@ -34,16 +34,20 @@ shred-derive = "0.6.0"
...
 
 serde = { version = "1.0", optional = true, features = ["serde_derive"] }
 
 [features]
 default = ["parallel"]
 parallel = ["rayon", "shred/parallel", "hibitset/parallel"]
 nightly = ["shred/nightly"]
+uuid_entity = ["uuid", "serde"]
+stdweb = ["uuid/stdweb"]
+wasm-bindgen = ["uuid/wasm-bindgen"]
```
torkleyy: Is this something we have to expose?

jojolepro: if you want to support both stdweb (cargo-web) and wasm then yes.

torkleyy:
Thanks!
bors r+

PR #594, opened 2019-05-17
Replace U64Marker with SimpleMarker<T: ?Sized>

### LINKED ISSUE #592: Turn U64Marker/U64MarkerAllocator into generics over a "marker" trait
## Description

`U64Marker` becomes `U64Marker<T: ?Sized>`, `U64MarkerAllocator` turns into `U64MarkerAllocator<T: ?Sized>`, allowing the following:

```rust
struct Networked;
struct FileSaved;

fn setup(world: &mut World) {
    world.register::<U64Marker<Networked>>;
    world.register::<U64Marker<FileSaved>>;
    world.add_resource(U64MarkerAllocator::<Networked>::default());
    world.add_resource(U64MarkerAllocator::<FileSaved>::default());
}
```

## Motivation

The provided implementation for Marker is most likely good enough for a large majority of the use-cases, however the inability to use it more than once is limiting.

## Drawbacks

* Is it a breaking change?

It is a breaking change

Arguably, the added set of carets for the generic type makes the code busier.
ZST, etc might not be the friendliest to newcomers.
I don't see how performance could be impacted.

## Unresolved questions

I'm not the most knowledgeable in terms of generics, isn't it possible to have "default" generics? If so, the changes would only be additive and not breaking.
If it comes down to breaking changes, a rename should be considered, I feel that `SimpleMarker` could be more appropriate.

---

I should be able to spend time on a PR implementing this if accepted

ldesgoui:
Fixes #592 

## Checklist

## API changes

Breaking change: 

`specs::saveload::U64Marker` is replaced with `SimpleMarker<T>`
`specs::saveload::U64MarkerAllocator` is replaced with `SimpleMarkerAllocator<T>`
This change allows multiple instances of SimpleMarker(Allocator) to be used in a single World

`SimpleMarker` is no longer `PartialOrd + Ord` because `derivative` doesn't implement those traits, however they are trivial to implement, up to you to decide whether I should

---
This change is [Reviewable](https://reviewable.io/reviews/slide-rs/specs/594)

### DIFF
--- a/src/saveload/marker.rs
+++ b/src/saveload/marker.rs
@@ -26,7 +26,6 @@ pub trait MarkedBuilder {
     ///     saveload::{MarkedBuilder, SimpleMarker, SimpleMarkerAllocator},
     /// };
     ///
-    /// #[derive(Clone, Debug, Hash, PartialEq, Eq)]
     /// struct NetworkSync;
     ///
     /// fn mark_entity<M: Builder + MarkedBuilder>(markable: M) -> Entity {
@@ -70,7 +69,6 @@ impl<'a> MarkedBuilder for LazyBuilder<'a> {
     ///     saveload::{MarkedBuilder, SimpleMarker, SimpleMarkerAllocator},
     /// };
     ///
-    /// #[derive(Clone, Debug, Hash, PartialEq, Eq)]
     /// struct NetworkSync;
     ///
     /// let mut world = World::new();
@@ -116,7 +114,6 @@ impl<'a> EntityResBuilder<'a> {
     ///     saveload::{SimpleMarker, SimpleMarkerAllocator},
     /// };
     ///
-    /// #[derive(Clone, Debug, Hash, PartialEq, Eq)]
     /// struct NetworkSync;
     ///
     /// let mut world = World::new();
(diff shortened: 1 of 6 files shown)


src/saveload/marker.rs
```diff
@@ -357,57 +366,65 @@ pub trait MarkerAllocator<M: Marker>: Resource {
...
-impl Component for U64Marker {
+impl<T: ?Sized + Send + Sync + 'static> Component for SimpleMarker<T> {
     type Storage = DenseVecStorage<Self>;
 }
 
-impl Marker for U64Marker {
-    type Allocator = U64MarkerAllocator;
+impl<T> Marker for SimpleMarker<T>
+where
+    T: 'static + ?Sized + Clone + Debug + Eq + Hash + Send + Sync,
```
ldesgoui: According to https://doc.rust-lang.org/std/marker/struct.PhantomData.html, those traits are implemented by PhantomData when `T: ?Sized`, except it wouldn't compile, mentioning that `T` needed to be `Clone + Debug + ...`

Is there a fix to this? I assume I'm doing something wrong

ldesgoui:
Tutorial changes can be previewed here: https://deploy-preview-594--specs.netlify.com/docs/tutorials/13_saveload.html#marker-and-markerallocator

src/saveload/marker.rs
ldesgoui: Oh, should it be
`where SimpleMarker<T>: Clone + Debug + Eq + Hash + Send + Sync` instead?

ldesgoui: Oh, I think I'm slamming my head against this: https://github.com/rust-lang/rust/issues/26925

src/saveload/marker.rs
```diff
@@ -357,57 +363,70 @@ pub trait MarkerAllocator<M: Marker>: Resource {
     fn maintain(&mut self, _entities: &EntitiesRes, _storage: &ReadStorage<M>);
 }
 
-/// Basic marker implementation usable for saving and loading
-#[derive(Clone, Copy, Debug, Hash, PartialEq, Eq, PartialOrd, Ord, Serialize, Deserialize)]
-pub struct U64Marker(u64);
+/// Basic marker implementation usable for saving and loading, uses `u64` as identifier
+#[derive(Derivative, Serialize, Deserialize)]
+#[derivative(Clone, Copy, Debug, Hash, PartialEq, Eq)]
+pub struct SimpleMarker<T: ?Sized>(u64, PhantomData<T>);
```
torkleyy: I think the second field should be skipped from the serialization. Maybe you could even make it transparent so we don't have a single-field tuple and just a `u64`.

PR #608, opened 2019-06-25
Recommend maybe() rather than EntitiesRes + get

Tomcc:
I discovered `maybe()` by chance because this tutorial recommends doing things another way, but I much prefer `maybe()`.
It's shorter, doesn't require `EntitiesRes`, and I assume it may be faster because it doesn't have to do a bunch of random lookups with `get()`.

So I tried to edit the documentation to include it, let me know if it sucks :)

It would be nice to explain `mask()` and `restrict()` in this section as well because those are really handy with `FlaggedStorage`s.

## Checklist


Tomcc:
@torkleyy sorry, I completely forgot about this one :P

Anyway, I added a section about `get()` and an example where you definitely need to use it AFAIK.

azriel91:
bors try

Issue #541, opened 2019-01-04
Allow joining over `FlaggedStorage` in parallel


torkleyy:
## Problem

We cannot join over a `FlaggedStorage` as usual because it does not fulfill the `DistinctStorage` guarantee (since it pushes events when accessed).

## Potential solution

We can push the events prior or after the actual access and keep it out of any parallel logic. This seems to be the technically most reasonable solution, but might be very hard to get right, especially with the API we have.

torkleyy:
cc @Moxinilian Would you be interested in taking a look?

Moxinilian:
I think the easiest solution would be to extend the parallel iteration process to call a storage-specific method prior to parallel iteration. This method would pass the joined mask bitset to the storages. We could push all the edit events at that time.

I think the best way to implement this is by extending the Join trait with a pre-process method (breaking change should be avoidable by defaulting it to a no-op). The FlaggedStorage would implement that method by doing the event push logic (and forwarding the call to the wrapped storage).

The `Join` trait would therefore get a new safe method `fn prepare(&mut self, mask: T::Mask)` to which the iteration bitset is passed. This method would be implemented for joinable tuples as a call to all `prepare` of the joinables with the mask. The default implementation should be a no-op. As this is statically determined, LLVM should be able to truncate those no-op calls, ensuring this is zero-cost for non-flagged storages.

The FlaggedStorage wrapper type would do the flagging of entities there using the mask (this could even make flagging more efficient considering it will be an optimizable bitset AND, maybe?). It would implement Join and ParJoin manually. In the immutable implementation of Join, the `prepare` method will only forward it to the wrapped storage, but in the mutable iteration, it will first register the masked entities as dirty, then forward it to the wrapped storage.

This prepare method would be called [here](https://github.com/slide-rs/specs/blob/16ac0fe85162a7377856c553b952755c144fbcc2/src/join/mod.rs#L297) for serial and [here](https://github.com/slide-rs/specs/blob/16ac0fe85162a7377856c553b952755c144fbcc2/src/join/par_join.rs#L46) for parallel.

Notice that this has the drawback of being inefficient, as the final mutable iteration might have filtered out some of the entities. It seems like the current FlaggedStorage doesn't do it either, so at least for now I believe it is acceptable.

Do you think this would be a reasonable fix?

Xaeroxe:
I realize this might come off as flippant but serious question has anyone ever seriously used a parallel join?

Moxinilian:
[...]

PR #613, opened 2019-07-09
Start and Stop signal registration for the FlaggedStorage

AndreaCatania:
Sometimes you may want to perform some operations on the storage, but you don't want
that these operations produce any event.

You can use the function `storage.set_event_emission(false)` to suppress the event writing for
of any action. When you want to re activate them you can simply call
`storage.set_event_emission(true)`

This change will be used to fix this: https://github.com/amethyst/amethyst/issues/1795

## Checklist

### DIFF
--- a/src/storage/flagged.rs
+++ b/src/storage/flagged.rs
@@ -172,6 +172,18 @@ pub struct FlaggedStorage<C, T = DenseVecStorage<C>> {
     phantom: PhantomData<C>,
 }
 
+impl<C, T> FlaggedStorage<C, T> {
+    #[cfg(feature = "storage-event-control")]
+    fn emit_event(&self) -> bool {
+        self.event_emission
+    }
+
+    #[cfg(not(feature = "storage-event-control"))]
+    fn emit_event(&self) -> bool {
+        true
+    }
+}
+
 impl<C, T> Default for FlaggedStorage<C, T>
 where
     T: TryDefault,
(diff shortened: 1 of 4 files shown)


jaynus:
Wouldn't setting `storage.set_event_emission(false)` accidentally allow for https://github.com/slide-rs/specs/issues/541 in some cases?

AndreaCatania:
Do you mean that when the emission is false, the parallel joint is allowed?

I think that this change doesn't fix or make worse that scenario, because the main point of that issue
is find a way to join in parallel the flagged storages and write the events.

So I would not consider this as a fix or a partial fix.

jojolepro:
bors try

torkleyy:
> Sometimes you may want to perform some operations on the storage, but you don't want
that these operations produce any event.

Why do you want that? It seems to me in this case the requirement stems from unclean design. `FlaggedStorage` is just meant to be a very simple wrapper which just gives you all the events.

If you want a more custom storage, please consider making your own instead of adding it to Specs; I'm trying to keep Specs small, and stabilize it.

AndreaCatania:
The reason is that sometimes a particular `System` perform some operations on the storage and these don't have to be propagated.
Like the parenting hierarchy resolution of the `Transform` component in `Amethyst` here https://github.com/amethyst/amethyst/issues/1795 .

Create a new Storage to solve this problem is doable, but mean also copy an paste 80% of the actual `FlaggedStorage`.

If you and the others are ok with it, I can create a new storage type that is capable of start and stop event signaling and so solve that issue.

AndreaCatania:
@torkleyy However this functionality is not so invasive, and only add a feature that can be useful even out of Amethyst.

Is always a good thing be able to `start` and `stop` event submission.

AndreaCatania:
Hello guys! Any idea about this PR?

jojolepro:
I'm okay with merging this, but if @torkleyy doesn't want to, then it would make sense to create this type of storage in an external crate and use it in amethyst.

The main downside is that using an external crate will require more maintenance.

AndreaCatania:
I waited a bit to hear the @torkleyy thoughts on this;
I assume that he don't want to merge it, so I'll create a new crate in the next few days.

_(By the way, I think that use a different crate for this is not the optimal solution in this case)_

AndreaCatania:
Yes I can to it.

> I think it would be fine to merge it when it's disabled by default; 

Is it already disabled by default.

> Can we put this under a feature flag?

Would you like to use a flag (i32 - bit wise operations) instead of a boolean?

AndreaCatania:
Re-thinking about your comment, probably your idea is to use the conditional building using the features. Let me know which one

jojolepro:
feature flag = "conditional building" = `features = ["somethingsomething"]`

AndreaCatania:
@jojolepro Thanks for the clarification!
@torkleyy Done!

src/storage/flagged.rs
```diff
@@ -166,16 +201,37 @@ impl<C: Component, T: UnprotectedStorage<C>> UnprotectedStorage<C> for FlaggedSt
...
+                self.channel.single_write(ComponentEvent::Inserted(id));
+            }
+        }
+        #[cfg(not(feature = "storage-event-control"))]
         self.channel.single_write(ComponentEvent::Inserted(id));
         self.storage.insert(id, comp);
     }
 
     unsafe fn remove(&mut self, id: Index) -> C {
+        #[cfg(feature = "storage-event-control")]
```
torkleyy: Can you please define a function which just returns `true` in case of a disabled feature?

AndreaCatania:
@torkleyy Done, now looks much better.

AndreaCatania:
Morning, any idea about this?

PR #620, opened 2019-07-27
impl Clone for JoinIter

### LINKED ISSUE #593: Implement Clone for JoinIter
## Description

Implement `Clone` for `JoinIter` when it is safe to do so.

## Motivation

I am currently trying to build a small collision system which simply checks for each entity pair if they collide, `JoinIter` implemented `Clone` I would be able to write it like this:

```rust
impl<'s> System<'s> for CollisionSystem {
    type SystemData = (
        Entities<'s>,
        ReadStorage<'s, Collider>,
        ReadStorage<'s, Transform>,
    );

    fn run(&mut self, (entities, colliders, transforms): Self::SystemData) {
        let mut collisions = Vec::new();

        let mut iter = (&entities, &colliders, &transforms).join();
        while let Some(a) = iter.next() {
            for b in iter.clone() // check all following entities {
                if is_collision((a.1, a.2), (b.1, b.2)) {
                    self.collisions.push(Collision(a.0, b.0))
                }
            }
        }

        for collision in collisions {
            // ...
        }
    }
}
```

I currently collect the `JoinIter` into a temporary `Vec` which does work, even if it is somewhat unsatisfying. If there is a prettier solution, please tell me as I only recently started using specs/amethyst.

## Drawbacks

* Is it a breaking change?

  - No

  - I don't think so

## Unresolved questions

How can we detect that the `JoinIter` iterates over a mutable collection, in which case
`Clone` must not get implemented.

---

I would gladly try to implement this if the unresolved question can be solved

lcnr:
Closes #593 

## Checklist

## API changes

implement `Clone` for `JoinIter` where possible.
Updates hibitset version as this requires access to `BitIter::clone` and `BitIterAnd::clone`.
This is not a breaking change

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+# Unreleased
+
+* Implement `Clone` for some `JoinIter`. ([#620])
+* Bump `hibitset` to `0.6.3`. ([#620])
+
+[#620]: https://github.com/slide-rs/specs/pull/620
+
 # 0.16.0
 
 * Update `syn`, `quote` and `proc-macro2` to `1.0`. ([#648])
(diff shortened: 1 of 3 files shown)


lcnr:
requires https://github.com/slide-rs/hibitset/pull/52

lcnr:
requires https://github.com/amethyst/hibitset/issues/54

azriel91:
bors try

lcnr:
@azriel91 should be ready for merge

WaDelma:
bors try

PR #621, opened 2019-07-27
fix link in docs

lcnr:


src/world/lazy.rs
```diff
@@ -49,6 +49,8 @@ impl<'a> Builder for LazyBuilder<'a> {
     /// Finishes the building and returns the built entity.
     /// Please note that no component is associated to this
     /// entity until you call [`World::maintain`].
```
torkleyy: You can just remove the backicks and it should work (rustdoc will resolve the path).

Issue #553, opened 2019-02-13
Idea: OpenCL-backed storage


michaelmelanson:
I have [a scientific simulation / numerical computation project](https://github.com/michaelmelanson/spiking-neural-net) I've written using Specs. But last night I did a spike where I prototyped an version that offloaded the calculations to the GPU via OpenCL, and I saw a 100x performance improvement over the Specs (i.e, CPU-based) version. That's kind of hard to ignore.

It occurred to me that if Specs had a component storage type backed by OpenCL buffers, then you could write Systems that apply OpenCL kernels to the data, and thereby make a Specs application that offloads some calculations to the GPU. You'd get to keep Specs' entity management, as well as have other systems that do CPU-based work in the same application, while shifting the heaviest lifting to the GPU.

Has any thinking been done on this front, or any related work? What major challenges do you see in pulling this off?

torkleyy:
This sound really interesting! Unfortunately, I don't have any experience with OpenCL.

I'm wondering how much we'd have to tie such an implementation to a specific OpenCL backend.

torkleyy:
_further activity_

OvermindDL1:
Still very relevant!

michaelmelanson:
This is awesome work, thanks!

willglynn:
#634 landed in `master`, bringing [slice access to selected storages](https://github.com/slide-rs/specs/blob/master/docs/tutorials/src/05_storages.md#slices). This is sufficient for my OpenCL needs. Usage:

```rust
impl<'a> System<'a> for SomeSystem {
    type SystemData = (WriteStorage<'a, Something>, );

    fn run(&mut self, (mut some_storage, ): Self::SystemData) {
        // 1. get a mutable slice of the component storage
        let slice = some_storage.unprotected_storage_mut().as_mut_slice();

        // 2. process the slice with OpenCL or whatever

        // 3. there is no step 3
    }
}
```

Does anyone here need `specs` to do something more before closing this issue?

jkelleyrtp:
Awesome work! Does anyone have some example code I can work off of? I'd like to add webgl/wgpu backend so specs can actually be multithreaded in the browser.

willglynn:
[...]

PR #634, opened 2019-08-28
Add slice access

willglynn:
This PR adds slice access to certain storages. Slices allow the user to bolt on an accelerator framework like OpenCL (#553) for high throughput component processing. My thoughts behind this particular design are https://github.com/slide-rs/specs/issues/553#issuecomment-525099211 and https://github.com/slide-rs/specs/issues/553#issuecomment-525362663.

## Checklist

## API changes

Non-breaking, strictly additive:

* Add `DenseVecStorage<T>::as_slice() -> &[T]` and `as_mut_slice() -> &mut [T]`.
* Add `VecStorage::unsafe_slice() -> &[T]` and `unsafe_mut_slice() -> &mut [T]`.
* Add `DefaultVecStorage<T>`, which works like `VecStorage<T>`, requires `T: Default`, and provides `as_slice() -> &[T]` and `as_mut_slice() -> &mut [T]`.
* Add `as` or `unsafe` `[_mut]_slice` directly on `Storage` as appropriate.

### DIFF
--- a/src/storage/storages.rs
+++ b/src/storage/storages.rs
@@ -11,6 +11,18 @@ use crate::{
     world::Index,
 };
 
+/// Some storages can provide slices to access the underlying data.
+///
+/// The underlying data may be of type `T`, or it may be of a type
+/// which wraps `T`. The associated type `Element` identifies what
+/// the slices will contain.
+pub trait SliceAccess<T> {
+    type Element;
+
+    fn as_slice(&self) -> &[Self::Element];
+    fn as_mut_slice(&mut self) -> &mut [Self::Element];
+}
+
 /// BTreeMap-based storage.
 #[derive(Derivative)]
 #[derivative(Default(bound = ""))]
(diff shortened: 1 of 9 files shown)


WaDelma:
I think those `unsafe_slice`s should return raw pointers as this is constructing partially initialized slice which I think is UB. Not 100% sure about this though.

WaDelma:
Well if the whole `Vec` is full then it would be safe yes. But it's practically useless. I would provide way of getting raw pointers and let the user construct slice out of it if they want.

WaDelma:
If you have `VecStorage<bool>` that contains uninitialised values, this would create reference to slice of booleans that have different values than `0` or `1` and I think having such reference might be UB. But I cannot be certain as there is no memory model yet. What I am certain that having `&bool` to uninitialized bool is insta UB (and so &T of uninitialised value is easily UB).

https://www.reddit.com/r/rust/comments/95vxdy/understanding_ub_with_stdmemuninitialized/

https://github.com/rust-lang/rust/issues/53491

willglynn:
I rearranged my changes and force-pushed the branch for this PR. The first commit adds `as_slice()`/`as_mut_slice()` to `DenseVecStorage` and adds `DefaultVecStorage`. This is sufficient for my use (and likely others' use) per #553. 

The second commit reworks `VecStorage` to use `MaybeUninit` per my previous comment and makes it expose `as_slice() -> &[MaybeUninit<T>]`. These changes are related (it would be _nice_ to slice `VecStorage`), but they are separable, and I'm happy to move that commit to a new PR if desired.

WaDelma:
Note that for f32 every bit pattern isn't allowed: Signaling NaN is a UB.

I am not sure if slice is just syntactic sugar from compilers POV. But yeah probably constructing slice isn't UB.

I think having `MaybeUninit` is good direction to go to as it gives more confidence on the correctness of this stuff. I am not against bumping the minimum rust version myself, but cannot say if that's for others.

willglynn:
> Note that for f32 every bit pattern isn't allowed: Signaling NaN is a UB.

🤔 Do you mean making such an `f32`, dereferencing such an `&f32`, or performing operations with such an `f32`?

[`f32::from_bits(v: u32)`](https://doc.rust-lang.org/std/primitive.f32.html#method.from_bits) says it's equivalent to `mem::transmute::<u32, f32>(v)`, and it mentions some platform-specific gotchas around NaN, but that's implementation-defined behavior and not undefined behavior.

> I think having `MaybeUninit` is good direction to go to as it gives more confidence on the correctness of this stuff. I am not against bumping the minimum rust version myself, but cannot say if that's for others.

:+1:

WaDelma:
Related: https://github.com/slide-rs/specs/issues/646

willglynn:
I note in https://github.com/slide-rs/specs/issues/646#issuecomment-536056356 why I now believe that slices must always refer to initialized elements in current Rust, contrary to my earlier comments here.

WaDelma:
This PR looks otherwise good, but `DenseVecStorage` should probably be changed to also have `MaybeUninit`.

willglynn:
Ah, the `Vec<T>`s is dense and always initialized, it's the the `Vec<Index>`es which are not. On it.

src/storage/storages.rs
```diff
@@ -83,12 +96,40 @@ unsafe impl<T> DistinctStorage for HashMapStorage<T> {}
...
+/// cannot be compared with indices from any other storage, and
+/// a particular entity's position within this slice may change
+/// over time.
 #[derive(Derivative)]
 #[derivative(Default(bound = ""))]
 pub struct DenseVecStorage<T> {
     data: Vec<T>,
-    entity_id: Vec<Index>,
-    data_id: Vec<Index>,
+    entity_id: Vec<MaybeUninit<Index>>,
```
WaDelma: Does `entity_id` need to contain `MaybeUninit`? We don't `set_len` on it, right?

willglynn: Oops, got over-zealous.

PR #654, opened 2019-10-23
Fix book link in README.md

wenyangfu:
Update book link to the correct location in README,


wenyangfu:
Ah, bummer - how do I merge this in? 😕

PR #668, opened 2020-01-09
Generic maskedstorage

### LINKED ISSUE #664: Turn MaskedStorage into a safe wrapper around UnprotectedStorage
At the moment, `MaskedStorage` requires a `Component` bound on its type parameter, and it's missing two of the methods that `UnprotectedStorage` offers. I think it would make the type more flexible, with relatively little work, if the following changes were made:

* Remove the `T: Component` bound, instead adding a second type parameter `S` for the storage, with a `S: UnprotectedStorage<T>` bound.
* Add the missing methods `.get` and `.get_mut` that forward to the corresponding `UnprotectedStorage` methods.

With these changes, the `MaskedStorage` type would be just a wrapper around `UnprotectedStorage` but with safe methods instead of unsafe ones. This will allow it to be used for other purposes than component storage, anything that needs an ID for lookup would work. An example is Amethyst, which currently uses `UnprotectedStorage` for asset storage, but which has had to build its own safety wrappers. If `MaskedStorage` were changed in this way, users could use that type instead, without safety concerns.

Rua:
Fixes #664 

## API changes

This changes three things:
* `UnprotectedStorage` has its item type as an associated type rather than as a generic type parameter.
* The type parameter of `MaskedStorage` is now the inner storage, rather than the item type.
This allows any item to be stored in a `MaskedStorage`, rather than only `Component`s.
* `MaskedStorage` now has `contains`, `get` and `get_mut` methods.

### DIFF
--- a/src/storage/data.rs
+++ b/src/storage/data.rs
@@ -118,17 +118,18 @@ use crate::{
 /// Note that you can also use `LazyUpdate` , which does insertions on
 /// `World::maintain`. This allows more concurrency and is designed
 /// to be used for entity initialization.
-pub type ReadStorage<'a, T> = Storage<'a, T, Fetch<'a, MaskedStorage<T>>>;
+#[allow(type_alias_bounds)]
+pub type ReadStorage<'a, T: Component> = Storage<'a, T, Fetch<'a, MaskedStorage<T, T::Storage>>>;
 
 impl<'a, T> SystemData<'a> for ReadStorage<'a, T>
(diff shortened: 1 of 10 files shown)


Rua:
There's two errors remaining that I can't figure out. They seem to have to do with the definitions of `ReadStorage` and `WriteStorage`, which are type aliases. If I substitute them for their definitions at the site of the error, the error goes away.

PR #670, opened 2020-01-19
Remove derivative dependency

allphyr:
## Checklist

* n/a I've added tests for all code changes and additions (where applicable)
* n/a I've added a demonstration of the new feature to one or more examples
* n/a I've updated the book to reflect my changes
* n/a Usage of new public items is shown in the API docs

Derivative is a quite costly dependence in term of build time, so we can remove it at small cost

### DIFF
--- a/src/saveload/marker.rs
+++ b/src/saveload/marker.rs
@@ -1,6 +1,10 @@
 //! Provides `Marker` and `MarkerAllocator` traits
 
-use std::{collections::HashMap, fmt::Debug, hash::Hash, marker::PhantomData};
+use std::{collections::HashMap,
+    fmt::{self, Debug},
+    hash::{Hash, Hasher},
+    marker::PhantomData
+};
 
 use serde::{de::DeserializeOwned, Deserialize, Serialize};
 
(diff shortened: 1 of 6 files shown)


src/saveload/marker.rs
```diff
@@ -362,11 +365,38 @@ pub trait MarkerAllocator<M: Marker>: Resource {
...
+
+impl<T: ?Sized> Hash for SimpleMarker<T> {
+    fn hash<H: Hasher>(&self, state: &mut H) {
+        self.0.hash(state);
+    }
+}
+
+impl<T: ?Sized> Debug for SimpleMarker<T> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "SimpleMarker({:?}, PhantomData)", self.0)
```
azriel91: Think it would be better to use [`Formatter::debug_tuple`](https://doc.rust-lang.org/std/fmt/struct.Formatter.html#method.debug_tuple) -- I think it handles writing for both normal debug and pretty-print debug.
This *might* work:

```rust
fmt.debug_tuple("SimpleMarker")
    .field(&self.0)
    .field(&format_args!("_"))
    .finish()
```

src/saveload/marker.rs
```diff
@@ -387,8 +417,7 @@ where
 }
 
 /// Basic marker allocator, uses `u64` as identifier
-#[derive(Derivative)]
-#[derivative(Clone, Debug)]
+#[derive(Clone, Debug)]
```
azriel91: Gotta manually implement this, the `std` derive would add a `Clone` and `Debug` bound to `T`

allphyr:
Do you need something else in order to merge this ?

azriel91:
Depends -- I'm happy to merge incremental improvements like these (with one approval), but checking with the other members.

PR #674, opened 2020-01-24
Remove send+sync requirement if not parallel

### LINKED ISSUE #673: Components require `Send` + `Sync` even when parallel feature is disabled
## Problem

`specs` does not allow you to use resources that aren't `Send` + `Sync`, even if the parallel feature is disabled.

This is an issue for me because I'm using `specs` in an application that is compiled to WebAssembly.

On the WebAssembly target one of my resources is a `wasm_bindgen::JsValue` - which is a type that at this time can't implement `Send` safely.

## Potential solution

Right now `Component` is defined as

```rust
pub trait Component: Any + Sized {
    /// Associated storage type for this component.
    type Storage: UnprotectedStorage<Self> + Any + Send + Sync;
}
```

Perhaps we can change this to

```rust
pub trait Component: Any + Sized {
    /// Associated storage type for this component.
    #[cfg(feature = "parallel")
    type Storage: UnprotectedStorage<Self> + Any + Send + Sync;
    #[cfg(not(feature = "parallel"))
    type Storage: UnprotectedStorage<Self> + Any;
}
```

I'd be happy to work on this (or any better approach)

chinedufn:
Closes #673

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,7 +26,7 @@ crossbeam-queue = "0.2"
 hashbrown = "0.6.0"
 hibitset = { version = "0.6.1", default-features = false }
 log = "0.4"
-shred = { version = "0.10.0", default-features = false }
+shred = { version = "0.10.1", default-features = false }
 shrev = "1.1.1"
 tuple_utils = "0.3"
 
@@ -54,7 +54,7 @@ criterion = "0.3"
 ron = "0.5"
 rand = "0.7"
 serde_json = "1.0"
-shred = { version = "0.10.0", features = ["shred-derive"] }
+shred = { version = "0.10.1", default-features = false, features = ["shred-derive"] }
 specs-derive = { path = "specs-derive", version = "0.4.0" }
 
 [[example]]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "specs"
-version = "0.15.1"
+version = "0.16.0"
 description = """
 Specs is an Entity-Component-System library written in Rust.
 """
--- a/src/world/lazy.rs
+++ b/src/world/lazy.rs
@@ -11,61 +11,49 @@ impl<T> Default for Queue<T> {
 }
 
 #[cfg(feature = "parallel")]
-mod traits_internal {
-    use crate::{Component, Entity, World};
-
-    pub trait LazyUpdateInternal: Send + Sync {
-        fn update(self: Box<Self>, world: &mut World);
-    }
-
-    pub trait FnLazyUpdate: FnOnce(&World) + Send + Sync + 'static {}
-    impl<F> FnLazyUpdate for F where F: FnOnce(&World) + Send + Sync + 'static {}
-
-    pub trait FnLazyUpdateMut: FnOnce(&mut World) + Send + Sync + 'static {}
-    impl<F> FnLazyUpdateMut for F where F: FnOnce(&mut World) + Send + Sync + 'static {}
-
-    pub trait ComponentBound: Component + Send + Sync {}
-    impl<C> ComponentBound for C where C: Component + Send + Sync {}
-
-    pub trait IntoIteratorBound<C>:
-        IntoIterator<Item = (Entity, C)> + Send + Sync + 'static
-    {
-    }
-    impl<I, C> IntoIteratorBound<C> for I
-    where
-        I: IntoIterator<Item = (Entity, C)> + Send + Sync + 'static,
-        C: ComponentBound,
-    {
-    }
+pub trait LazyUpdateInternal: Send + Sync {
+    fn update(self: Box<Self>, world: &mut World);
 }
 
 #[cfg(not(feature = "parallel"))]
-mod traits_internal {
-    use crate::{Component, Entity, World};
-
-    pub trait LazyUpdateInternal {
-        fn update(self: Box<Self>, world: &mut World);
-    }
-
-    pub trait FnLazyUpdate: FnOnce(&World) + 'static {}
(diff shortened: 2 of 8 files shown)


Cargo.toml
```diff
@@ -27,7 +27,7 @@ derivative = "1"
 hashbrown = "0.6.0"
 hibitset = { version = "0.6.1", default-features = false }
 log = "0.4"
-shred = { version = "0.9.3", default-features = false }
+shred = { version = "0.10.1", default-features = false }
```
chinedufn: `shred 0.10.1` needs to be deployed before this PR can be merged

Cargo.toml
```diff
@@ -39,7 +39,6 @@ uuid = { version = "0.7.4", optional = true, features = ["v4", "serde"] }
 [features]
 default = ["parallel"]
 parallel = ["rayon", "shred/parallel", "hibitset/parallel"]
-nightly = ["shred/nightly"]
```
chinedufn: A breaking change so I bumped the minor version of `specs`

We could also leave this feature in and just make it `nightly = []` if we don't want to bump to `0.16` just yet - then I can just make this PR bump to `0.15.2`

Whatever works for `specs`!

azriel91:
Heya, took a while to get back, but the changes in both PRs look good to me.

In terms of repository ownership, I don't yet feel like I'm "the one", else I'd be happy to take them.
@WaDelma since @torkleyy's focusing on other things, would you mind me making decisions at this level? am also happy to tag you whenever this kind of decision needs to be made (usually directional, rather than technical).

WaDelma:
@chinedufn tbh I have never felt like being "the one" either and have been bit afraid of stepping on other peoples toes accidentally, so I am happy to let you be it. I can participate on things thouh!

azriel91:
@chinedufn I've tried getting `specs` to be tested with the `shred` change in #677, but Rust isn't happy yet:

```bash
$ cargo test --no-default-features --workspace                                                                                                                                                                                                
# ..
error: aborting due to 55 previous errors
# ..
```

Could you check what's blocking that? I haven't merged `shred`'s PR because I'd like to see it working end-to-end first (instead of merge/release and needing to yank versions).

chinedufn:
@azriel91 sorry about that - https://github.com/amethyst/specs/pull/674/commits/d74d02b724529994c729ea8a5c4a3f897e2d1655 makes the following pass

```
cargo test --workspace && \
  cargo test --no-default-features --workspace
```

examples/cluster_bomb.rs
```diff
@@ -1,9 +1,11 @@
 extern crate rand;
```
chinedufn: Opted for the quick approach to get things passing with the parallel feature disabled by just panicking.

I thought that the work of making the examples work in parallel and non-parallel could be left for the future since these wouldn't have worked in `not(parallel)` before anyway

examples/full.rs
```diff
@@ -1,4 +1,6 @@
+#[cfg(feature = "parallel")]
```
chinedufn: Same idea here - quick and dirty approach of panicking when not parallel

tests/no_parallel.rs
```diff
@@ -0,0 +1,14 @@
+#![cfg(not(feature = "parallel"))]
```
chinedufn: @azriel91 addresses https://github.com/amethyst/specs/pull/674#pullrequestreview-352585509 - thanks for the feedback!

azriel91:
ah oops, I did similar changes in #677 (but didn't force push to your branch -- in case you still wanted it).

oh. just discovered why -- `[dev-dependencies]` didn't have `default-features = false` for `shred`. Shall push that one over this branch soon (shall release `shred`).

azriel91:
**Goal:** Support `Component`s that are **not** `Send + Sync`.

There is a test in `tests/no_parallel.rs` to prove the goal is reached when `"parallel"` is disabled. Existing tests prove when `"parallel"` is enabled.

* **Ideal solution:** [Trait aliases](https://github.com/rust-lang/rust/issues/41517), in beta but not stable, and very little signs of movement.

    ```rust
    #![cfg(feature = "parallel")]
    trait ComponentBound = Component + Send + Sync;
    trait ComponentBound = Component;
    ```

* **Duplicate code:** Copy code per `feature = "parallel"` and `not(feature = "parallel")`. Simple but ugly.
* **Poor man's trait aliases:** Implemented in e1477cf, but not ideal, because users have to specify the type of `world` in their closures: 1ed2912, which is very bad<sup>TM</sup>.
* **Macro to generate per feature fn (current state):** Kind of like the duplication option without the duplication, but macro is ugly to maintain.

@chinedufn @WaDelma I think it's good to have both of you review this (again). The summary above should capture the options I've tried.

src/world/lazy.rs
```diff
@@ -10,6 +10,51 @@ impl<T> Default for Queue<T> {
...
+pub trait LazyUpdateInternal: Send + Sync {
+    fn update(self: Box<Self>, world: &mut World);
+}
+
+#[cfg(not(feature = "parallel"))]
+pub trait LazyUpdateInternal {
+    fn update(self: Box<Self>, world: &mut World);
+}
+
+macro_rules! parallel_feature {
```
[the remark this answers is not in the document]
azriel91: ah of course! shall do that in the morning

azriel91:
Just realized that's no test for a storage that isn't `Send`, if you like, could you add one? That'll help the PR get through a little faster.

chinedufn:
If the `type = VecStorage<Self>` and `Self` isn't send then isn't the storage non send?

Or am I missing something here?

azriel91:
ah that's correct :v: -- it backs onto `Vec` which only `impl`s `Send` when T is `Send`

PR #677, opened 2020-02-11
Non send sync resources

azriel91:
Attempt at #673 / get #674 to be tested.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+# 0.16.0
+
+* Components and resources no longer need to be `Send + Sync` if parallel feature is disabled ([#673])
+* Remove the `nightly` feature ([#673])
+
+[#672]: https://github.com/slide-rs/specs/pull/672
+
 # 0.15.1
 
 * Benchmark uses `nalgebra` instead of `cgmath`. ([#619]).
--- a/tests/no_parallel.rs
+++ b/tests/no_parallel.rs
@@ -1,0 +1,29 @@
+#![cfg(not(feature = "parallel"))]
+
+use std::rc::Rc;
+
+use specs::{storage::VecStorage, Builder, Component, World, WorldExt};
+
+#[derive(PartialEq)]
+struct CompNonSend(Rc<u32>);
+
+impl Component for CompNonSend {
+    type Storage = VecStorage<Self>;
+}
+
+#[test]
+fn non_send_component_is_accepted() {
+    let mut world = World::new();
+    world.register::<CompNonSend>();
+
+    let entity = world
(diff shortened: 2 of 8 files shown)


tests/no_parallel.rs
```diff
@@ -0,0 +1,27 @@
...
+}
+
+#[test]
+fn non_send_component_is_accepted() {
+    let mut world = World::new();
+    world.register::<CompNonSend>();
+
+    let entity = world
+        .create_entity()
+        .with(CompNonSend(Rc::new(123)))
```
azriel91: @chinedufn I've made this test different to the one in `shred` -- in `shred` it tests resources, but in `specs` we have to test components.

This part doesn't compile yet -- even if the backing `UnprotectedStorage` doesn't have the `Send + Sync` requirement, the `Component` itself needs to have that relaxed. But when relaxing it in the `Builder` trait (which it is done above), the constraint ripples to other types as well (many, many ripples).

Since there's many ripples, and it's not maintainable to create two versions of the same code (one with and one without `Send + Sync`), I kind of want to wait for [trait aliases](https://github.com/rust-lang/rust/issues/41517) to be stabilized so that we can do something like:

```rust
#![cfg(feature = "parallel")]
trait ComponentBound = Component + Send + Sync;
trait ComponentBound = Component;
```

and use `ComponentBound` where it's needed.

azriel91: Wait a minute, do we need to support the `Component` being `!Send + !Sync`? or is it enough to support the storage not being `Send + Sync`?

CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
+# 0.16.0
+
+* Components and resources no longer need to be `Send + Sync` if parallel feature is disabled ([#673])
```
azriel91: **TODO:** Likely should be component *storages*. `Component`s still have to be `Send + Sync`.

tests/no_parallel.rs
azriel91: Okay, the experiment shows it's possible -- the test passes. But the duplication is still unwieldly.
Shall attempt a less duplicated approach in a bit.

azriel91: e1477cf removes the duplication, but forces closures to explicitly specify the type of the `world` variable: 1ed2912, which is an ergonomic breakage I would like to avoid.

azriel91: okay, so b4e83ce uses a macro to generate the feature-specific methods (with/without `Send + Sync`). Gonna do a quick summary for reviewers

azriel91:
Bringing discussion back to #674.

Issue #679, opened 2020-02-11
Release 0.16.0


azriel91:
There are a number of useful, yet unreleased changes since `specs 0.15.1`. This issue tracks the release so that people may pick up the changes without needing to depend on a git dependency.

Tasks:

* [x] Update `CHANGELOG.md`.
* [x] Release `specs`.
* [x] Release `specs-derive`.

azriel91:
Released:

* :label: tag: https://github.com/amethyst/specs/tree/0.16.0
* :card_file_box: GH releases: https://github.com/amethyst/specs/releases/tag/0.16.0
* :package: crates.io: https://crates.io/crates/specs/0.16.0

erlend-sh:
Sweet! I’ll spread the word.

One tiny thing first: For some reason it’s not tagged as the Latest release, so it appears only has a hidden tag in the Releases page.

azriel91:
ah, looks like you have to manually create a "Release" from a tag. Have just done that :v:

PR #680, opened 2020-02-11
Meta/679/release 0 16 0

azriel91:
Prepare a release #679.

I've omitted typos / dependency bumps from the changelog where it's not public API.

Shall close that when the tag is created and the crate is published.


azriel91:
bors r-

(shall bump `specs-derive` version as well)

PR #687, opened 2020-02-18
Add derive flag

### LINKED ISSUE #684: All-inclusive `derive` feature
## Description

Add another `derive` flag that enables all the available `*-derive` flags.

## Motivation

I've spent non-trivial time searching why `#[derive(SystemData)]` does not work for me when I've already included the `specs-derive` feature. It is „hidden“ in `shred-derive` ‒ but, using shred is somewhat internal thing of specs so this is non-obvious. The documentation (book) doesn't really speak about specific features much either.

Including it as `derive` (same as serde and others), it is much easier to guess and it would allow all derives at once, saving such surprise.

## Drawbacks

* Is it a breaking change?

No

This might have slight impact on first compilation time to compile additional derive crate (eg. if I need just one and not the other). But as they both share the heavy dependencies (syntax, quote), it should be small. It's build-time only, no impact for runtime.

Users would still be able to enable only one of the originals if it mattered to them.

I have the actual change ready, I'd be willing to go through the documentation and mention the new flag.

vorner:
* There are currently two distinct *-derive features and it is
  uncomfortable to hunt for them.
* They exist mostly due to technical reasons, as they bring two
  different crates. The price of the other if one is already present is
  relatively low (the cost is by depending on syntax and quote and they
  share them).
* This is in the convention of having the `derive` feature flag as serde
  and other packages have.
* The original two are still preserved, so users still can enable only
  one of them.

Closes #684.

## Checklist

* [x] I've added tests for all code changes and additions (where applicable) (not applicable)
  Should I update the examples to use this feature, or leave them be with the current `specs-derive`?

I'm not sure this completely applies. I've updated the note in docs about features, but otherwise left the `specs-derive` as the main suggestion how to use it by users and expect this will be more like for people who go more by guess than by documentation. But if you prefer, I can update all the examples, etc, to use this new feature flag instead.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,8 +2,11 @@
 
 * Deprecate `error::NoError` in favor of `std::convert::Infallible` ([#688])
 * Use `#[non_exhaustive]` for `error::Error`. Note this bumps the minimum supported rust version to 1.40 ([#688]).
+* Add the `derive` feature that enables all derive-related smaller features
+  (`specs-derive` and `shred-derive` currently). ([#687])
 
 [#688]: https://github.com/amethyst/specs/pull/688
+[#687]: https://github.com/amethyst/specs/pull/687
 
 # 0.16.1 (2020-02-18)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,8 +5,8 @@
 * Add the `derive` feature that enables all derive-related smaller features
   (`specs-derive` and `shred-derive` currently). ([#687])
 
-[#688]: https://github.com/amethyst/specs/pull/688
 [#687]: https://github.com/amethyst/specs/pull/687
+[#688]: https://github.com/amethyst/specs/pull/688
 
 # 0.16.1 (2020-02-18)
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -39,7 +39,7 @@
 //! }
 //! ```
 //!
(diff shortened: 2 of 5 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
+# Unreleased
+
+* Add the `derive` feature that enables all derive-related smaller features
+  (`specs-derive` and `shred-derive` currently). ([#687])
+
+[#687]: https://github.com/amethyst/specs/pull/687
```
vorner: Should I be adding (and guessing the PR number) manually, or is there an automatic way how this happens?

Also, the other links seem to be to slide-rs/specs, but this repo is amethyst/specs. Is that a previous place where it lived?

azriel91: Heya, it's manual for now. Normally I would look at the most recent issues/PR and add one (ya, not a great method).

Yeap it used to live in `slide-rs` before moving into `amethyst` (shall update the links in a later PR).

src/lib.rs
```diff
@@ -39,7 +39,7 @@
 //! }
 //! ```
 //!
-//! Or alternatively, if you enable the `specs-derive` feature, you can use a
+//! Or alternatively, if you enable the `specs-derive` or `derive` feature, you can use a
```
azriel91: I think it's alright just mentioning the `"derive"` feature without mentioning `specs-derive`.

vorner:
Thank you :-)

PR #688, opened 2020-02-19
Cleanup error types

allphyr:
## Checklist

* *n/a* I've added tests for all code changes and additions (where applicable)
* *n/a* I've added a demonstration of the new feature to one or more examples
* *n/a* Usage of new public items is shown in the API docs

## API changes

- Use `#[non_exhaustive]` for `error::Error` instead of `__NonExhaustive` trick. **This is a breaking change, and requires a minimal Rust version of 1.40** (previously 1.38). This notably enables to add new error types without it being a breaking change.
- Change `std::error::Error` implementation of different error types, to use "state of the art" functions instead of deprecated ones.
- Deprecated `NoError` and use `std::convert::Infallible`, and re-export the latter as `NoError` for a smoother transition (maybe we should remove completly). This will also allow a smoother transition to never type when it is stabilized.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -105,7 +105,7 @@ impl Display for WrongGeneration {
     fn fmt(&self, f: &mut Formatter) -> FmtResult {
         write!(
             f,
-            "Tried to {} entity {:?}, but the generation is not valid anymore; it should be {:?}",
+            "Tried to {} entity {:?}, but the generation is no longer valid; it should be {:?}",
             self.action, self.entity, self.actual_gen
         )
     }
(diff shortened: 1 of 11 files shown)


src/error.rs
```diff
@@ -122,33 +105,14 @@ impl Display for WrongGeneration {
     fn fmt(&self, f: &mut Formatter) -> FmtResult {
         write!(
             f,
-            "Tried to {} entity {:?}, but the generation is wrong; it should be {:?}",
+            "Tried to {} entity {:?}, but the generation is not valid anymore; it should be {:?}",
```
azriel91: pedantic :stuck_out_tongue:: `no longer valid`

allphyr: Thanks !

PR #691, opened 2020-03-19
`LazyUpdate::exec` and `LazyUpdate::exec_mut` should not require `F: …

kyren:
…Sync`

`F` must already be `FnOnce`, and the `Sync` requirement is too strict, it
should only require `Send` when the `parallel` feature is enabled.

Unfortunately I couldn't come up with a clean way of using or modifying the
`parallel_feature` macro to do this, so I was forced to copy the doc comments
and implementation of exec / exec_mut.  If you would rather make an additional
`parallel_feature` macro or modify that in some way let me know.

## Checklist

## API changes

I don't *think* this is a breaking API change since it's strictly reducing the bounds of two LazyUpdate methods.  It also removes requirements on two pub traits but they are marked "internal" so hopefully they shouldn't matter?

### DIFF
--- a/src/world/lazy.rs
+++ b/src/world/lazy.rs
@@ -11,7 +11,7 @@ impl<T> Default for Queue<T> {
 }
 
 #[cfg(feature = "parallel")]
-pub trait LazyUpdateInternal: Send + Sync {
+pub trait LazyUpdateInternal: Send {
     fn update(self: Box<Self>, world: &mut World);
 }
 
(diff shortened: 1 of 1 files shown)


WaDelma:
You could add new trait `MaybeSend` (or some better name) that either does or does not have `Send` as super trait depending on feature

azriel91:
Yeap, removing the bounds doesn't break API, though I find the `fn exec` taking a `&mut World` to be a bug... ish, introduced probably by me -- see [`lazy.rs@a5459b2:L262-L271`](https://github.com/amethyst/specs/pull/674/files#diff-a5459b249cd8c2255bfbc424048f8278L262-L271), though before that it already held a `&mut World` and passed it as `&World` to the function.

Fixing *that* would be a breaking change, but people can simply use `exec_mut` anyway.

I'm just wondering if we *should* relax `Sync` for when `"parallel"` is enabled.

PR #701, opened 2020-05-24
Helper attributes for fine grained ConvertSaveload control.

Azaril:
## Checklist

## API changes

This change adds additional control to ConvertSaveload derive behavior to customized code generation for the conversion data type. This allows for support of serde traits and non-serializable types. This was previously impossible as the attributes would be identical between the original data type and the conversion data type. This was due to serde traits being invalid on the original data type even if valid on the conversion data type.

Two new helper attributes are added to the ConvertSaveload derive behavior:

1.  'convert_save_load_skip_convert' bypasses calling convert_into/convert_from completly. This is needed as it's not possible to implement a blanket ConvertSaveLoad for 'Clone' types as this behavior is likely not desired and will also create conflicting trait implementation due to being too broad.
2. 'convert_save_load_attr' allows for attribute expansion only on the derived data type. This behaves similar too 'cfg_attr' in that the internal attributes is evaluated and attached in place. This can be applied to fields as well as enum variants.

### DIFF
--- a/specs-derive/src/impl_saveload.rs
+++ b/specs-derive/src/impl_saveload.rs
@@ -7,8 +7,8 @@
 
 use proc_macro2::{Span, TokenStream};
 use syn::{
-    DataEnum, DataStruct, DeriveInput, Field, GenericParam, Generics, Ident, Type, WhereClause,
-    WherePredicate, Attribute
+    Attribute, DataEnum, DataStruct, DeriveInput, Field, GenericParam, Generics, Ident, Type,
+    WhereClause, WherePredicate,
 };
 
 /// Handy collection since tuples got unwieldy and
@@ -102,7 +102,7 @@ pub fn impl_saveload(ast: &mut DeriveInput) -> TokenStream {
 
 struct FieldMetaData {
     field: Field,
-    skip_field: bool
+    skip_field: bool,
 }
 
 /// Implements all elements of saveload common to structs of any type
@@ -189,10 +189,10 @@ fn saveload_named_struct(
         let field_ident = &field.ident;
 
         if field_meta.skip_field {
-            quote!{ #field_ident: self.#field_ident.clone() }
+            quote! { #field_ident: self.#field_ident.clone() }
         } else {
-            quote!{ #field_ident: ConvertSaveload::convert_into(&self.#field_ident, &mut ids)? }
-        }        
+            quote! { #field_ident: ConvertSaveload::convert_into(&self.#field_ident, &mut ids)? }
+        }
     });
 
     let ser = quote! {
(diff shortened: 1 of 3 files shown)


specs-derive/src/impl_saveload.rs
```diff
@@ -175,30 +172,52 @@ fn saveload_named_struct(
     name: &Ident,
     saveload_name: &Ident,
     generics: &Generics,
-    saveload_fields: &[Field],
+    saveload_fields: &[FieldMetaData],
 ) -> (TokenStream, TokenStream, TokenStream) {
     let (_, ty_generics, where_clause) = generics.split_for_impl();
 
+    let fields = saveload_fields.iter().map(|f| f.field.clone());
```
azriel91, line 179: does `fields` need to be owned? If I remember correctly, they should be able to be converted into tokens even if they are `&Field`s.

oh, if it's to do with lifetimes, wrapping it in `{}` should work:

```rust
let struct_def = {
    let fields = saveload_fields.iter().map(|f| f.field);

    quote! {
        struct #saveload_name #ty_generics #where_clause {
            #( #fields ),*
        }
    };
};
```

specs-derive/src/impl_saveload.rs
```diff
@@ -234,39 +253,88 @@ fn saveload_tuple_struct(
...
     saveload_name: &Ident,
     generics: &Generics,
-    saveload_fields: &[Field],
+    saveload_fields: &[FieldMetaData],
 ) -> (TokenStream, TokenStream, TokenStream) {
     use syn::Index;
 
     let (_, ty_generics, where_clause) = generics.split_for_impl();
 
+    let fields = saveload_fields.iter().map(|f| f.field.clone());
```
azriel91, line 262: same comment re:cloning field

specs-derive/src/impl_saveload.rs
```diff
@@ -234,39 +253,88 @@ fn saveload_tuple_struct(
...
-    let field_ids = saveload_fields.iter().enumerate().map(|(i, _)| Index {
-        index: i as u32,
-        span: data.struct_token.span,
-    });
-    let tmp = field_ids.clone();
+    let field_ser = saveload_fields
+        .iter()
+        .enumerate()
+        .map(|(i, field_meta)| {
+            let field_ids = Index {
```
azriel91, line 274: nitpick, can this be singular now

specs-derive/src/impl_saveload.rs
```diff
@@ -301,85 +369,137 @@ fn saveload_tuple_struct(
...
 
                 big_match_ser = quote! {
                     #big_match_ser
-                    #name::#ident( #( ref #field_ids ),* ) => #saveload_name::#ident( #( ConvertSaveload::convert_into(#field_ids_2, &mut ids)? ),* ),
+                    #name::#ident( #( ref #field_ser_ids ),* ) => #saveload_name::#ident( #(#field_ser),* ),
                 };
 
-                let field_ids = tmp;
-                let field_ids_2 = field_ids.clone();
+                let field_de_ids = saveload_fields
```
azriel91, line 480: hm, this looks exactly the same as `field_ser_ids`.
How about just going:
```rust
// Both serialization and deserialization semantically use the same values.
field_de_ids = field_ser_ids;
```

specs-derive/src/impl_saveload.rs
```diff
@@ -301,85 +369,137 @@ fn saveload_tuple_struct(
...
-                    .map(|(i, _)| Some(Ident::new(&format!("field{}", i), data.enum_token.span)))
-                    .collect();
-                let field_ids_2 = field_ids.clone();
-                let tmp = field_ids.clone();
+                    .map(|(i, _)| make_field_ident(i))
+                    .collect::<Vec<_>>();
+
+                let field_ser = saveload_fields
+                    .iter()
+                    .enumerate()
```
azriel91, line 463: can save on an `enumerate()` and `make_field_ident` by zipping:

```rust
saveload_fields
    .iter()
    .zip(field_ser_ids.iter())
    .map(|(field_ident, field_meta)| {
        if field_meta.skip_field {
            quote! { #field_ident.clone() }
        } else {
            quote! { ConvertSaveload::convert_into(#field_ident, &mut ids)? }
        }
    })
```

same for deserialization

PR #707, opened 2020-08-09
Allow reentrant laziness

WaDelma:
## Checklist

## API changes

None

## Description
Before this change if one tried to do `World::system_data<Read<LazyUpdate>>()` within `LazyUpdate::exec` on would get the following error:
```
thread 'main' panicked at 'Already borrowed mutably: InvalidBorrow', /home/delma/.cargo/registry/src/github.com-1ecc6299db9ec823/shred-0.7.2/src/cell.rs:103:9
stack backtrace:                                                                                                                                              
   0: backtrace::backtrace::libunwind::trace                                                             
             at /cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.46/src/backtrace/libunwind.rs:86
   1: backtrace::backtrace::trace_unsynchronized                                                         
             at /cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.46/src/backtrace/mod.rs:66 
   2: std::sys_common::backtrace::_print_fmt                                                             
             at src/libstd/sys_common/backtrace.rs:78                                                    
   3: <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt                  
             at src/libstd/sys_common/backtrace.rs:59                                                    
   4: core::fmt::write
             at src/libcore/fmt/mod.rs:1069
   5: std::io::Write::write_fmt
             at src/libstd/io/mod.rs:1504
   6: std::sys_common::backtrace::_print
             at src/libstd/sys_common/backtrace.rs:62
   7: std::sys_common::backtrace::print
             at src/libstd/sys_common/backtrace.rs:49
   8: std::panicking::default_hook::{{closure}}
             at src/libstd/panicking.rs:198
   9: std::panicking::default_hook
             at src/libstd/panicking.rs:218
  10: std::panicking::rust_panic_with_hook
             at src/libstd/panicking.rs:511
  11: rust_begin_unwind
             at src/libstd/panicking.rs:419
  12: core::panicking::panic_fmt
             at src/libcore/panicking.rs:111
  13: core::option::expect_none_failed
             at src/libcore/option.rs:1268
  14: <alloc::boxed::Box<F> as core::ops::function::FnOnce<A>>::call_once
  15: specs::world::World::maintain
```
This problem arised in our project when I refactored code such that piece of code so that previously wasn't used inside `LazyUpdate::exec` was now used there. The piece of code is also used outside of `LazyUpdate` so just removing `LazyUpdate` usage in it is non-trivial. 

This PR changes the behavior to allow reentrant `LazyUpdate` usage. After the change reentrant `LazyUpdate::exec`:s will be applied in the order they are created on the same call of `World::maintain`.  Another option was to execute them on the next call to maintain. This option wasn't chosen to make applying lazy updates as soon as possible as `LazyUpdate` is not meant for delaying, but for giving more flexibility.

I think allowing this kind of reentrancy makes using `LazyUpdate` more robust.

### DIFF
--- a/src/world/lazy.rs
+++ b/src/world/lazy.rs
@@ -1,7 +1,7 @@
 use crossbeam_queue::SegQueue;
 
 use crate::{prelude::*, world::EntitiesRes};
-use std::mem;
+use std::sync::Arc;
 
 struct Queue<T>(SegQueue<T>);
 
@@ -153,7 +153,7 @@ where
 /// so there's no need to get `LazyUpdate` mutably.
 /// This resource is added to the world by default.
 pub struct LazyUpdate {
-    queue: Queue<Box<dyn LazyUpdateInternal>>,
+    queue: Arc<Queue<Box<dyn LazyUpdateInternal>>>,
 }
 
 impl Default for LazyUpdate {
@@ -313,7 +313,7 @@ impl LazyUpdate {
         {
             self.queue
                 .0
-                .push(Box::new(|w: &mut World| f(w)));
+                .push(Box::new(f));
         }
 
         /// Lazily executes a closure with mutable world access.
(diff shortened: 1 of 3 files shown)


WaDelma:
I think that the semantics of applying reentrant `LazyUpdate` calls immediately is probably the better way to go. The purpose of `LazyUpdate` is not to delay actions, but allow more flexibility and executing it's modifications as early as possible is most likely better. I'll look at how that could be implemented later.

WaDelma:
Now the alternative of applying reentrant lazy updates on the same maintain call is implemented.

src/world/lazy.rs
```diff
@@ -376,24 +375,14 @@ impl LazyUpdate {
         LazyBuilder { entity, lazy: self }
     }
 
-    /// Allows to temporarily take the inner queue.
-    pub(super) fn take(&mut self) -> Self {
+    pub(super) fn clone(&self) -> Self {
```
WaDelma, line 378: We could probably just derive `Clone` instead of doing this. Does that have any down sides? Well at least we cannot remove that later without breaking change.

azriel91: I think it wouldn't be too different; the current way just prevents a `LazyUpdate` from being passed when there's a `Clone` bound, but I doubt that's a common use case.

WaDelma:
bors r+

Previous bors timed out

PR #753, opened 2022-07-02
prepare release `0.18.0` and switch to the recommended license format…

xMAC94x:


AngelOnFira (changes requested):
After the other info is added to the changelog, lgtm :)

xMAC94x:
Not a huge fan of just printing the full merge history:
 - unsafe block scope is nice, but has no external influence, its just cleanup
 - extra examples are also not externally important, when someone is deciding wether they wanna upgrade or not it doesnt matter.
  - even the github actions technically dont fall in that category, i am tempting to remove them. 
  - fmt is also just internal stuff, no consumer of the api cares about that.
  
New contributors would be funny, but at this point of the crate it was never done in history and i dont know if we should start with it and bloat our Changelog with it.

The link to the full changelog sounds doable, i could include that

AngelOnFira (approved):

PR #754, opened 2022-07-10
Add missing licenses to specs-derive crate root

torkleyy:
Fixes https://github.com/amethyst/specs/issues/735


xMAC94x (approved):
seems like other repos also have duplicate LICENSE files ✅
SHA1 of Apache License is the same as in the root folder ✅
MIT license only differs in the Copyright year ✅

PR #756, opened 2022-07-12
Bump shred to 0.14.0

### LINKED ISSUE #746: Migrate to Rust 2021
Perhaps it is time to migrate to Rust 2021?
I believe we are currently at 2018

I just want to put an issue out here so we can track progress.

torkleyy:
## API changes

* Public dependency `shred` changed
* Resources no longer provide `mopa::Any` trait implementation

## Other changes

* Bump MSRV to 1.60.0 and test it in CI
* Bump edition to 2021
* Remove `hashbrown` dependency

Fixes #746


xMAC94x:
is there a reason why to remove hashbrown completly ?

torkleyy:
Hmm, I just saw they use different hashing algorithms by default. We should consider using a different algorithm than the default (we don't need HashDOS protection).

xMAC94x:
> Hmm, I just saw they use different hashing algorithms by default. We should consider using a different algorithm than the default (we don't need HashDOS protection).

yeah, i had something similar in mind. that stds is crypto-secure by default and we basically dont need that and could go for an algorithm that is very good for hashing of usize (or whatever we need to hash mostly).

Additionally hashbrown has got newer releases since 2019, and i am not sure if we keep udpating std s ones or not

Cargo.toml
```diff
@@ -55,7 +53,7 @@ criterion = "0.3.1"
 ron = "0.7.1"
 rand = "0.8"
 serde_json = "1.0.48"
-shred = { version = "0.13.0", default-features = false, features = ["shred-derive"] }
+shred = { version = "0.14.0", default-features = false, features = ["shred-derive"] }
```
xMAC94x, line 56: 0.14.1

Cargo.toml
```diff
@@ -12,21 +12,19 @@ keywords = ["gamedev", "ecs", "entity", "component"]
 categories = ["concurrency", "game-engines"]
 license = "MIT OR Apache-2.0"
 authors = ["slide-rs hackers"]
-exclude = ["bors.toml", ".travis.yml"]
-edition = "2018"
+include = ["src", "README.md", "LICENSE-MIT", "LICENSE-APACHE"]
+edition = "2021"
+rust-version = "1.60.0"
```
xMAC94x, line 17: the feature `nightly` was added some time back to introduce GAT and not increase the minimal rust version. we should check if its possible to just enable this by default now and if the required features are stabelized in 1.60.0

CHANGELOG.md
```diff
@@ -1,4 +1,8 @@
-# Unreleased
+# 0.19.0 (2022-07-12)
+
+* Bump shred to version `0.14.1`, MSRV to 1.60.0 ([shred changelog][shred-changelog], [#756])
```
xMAC94x, line 3: i really would like to see some benchmarks on either specs or shred level to show that there is no performance regression to hashbrown. because all we did so far was assuming on some hash-algorithms that it will prob be fine.

But for us as veloren we prob won't update to this new versions if we cannot be sure that the performance has no regression in some benchmarks of the crate itself.

torkleyy: > i really would like to see some benchmarks on either specs or shred level to show that there is no performance regression to hashbrown. because all we did so far was assuming on some hash-algorithms that it will prob be fine.

If you check the source code of `libstd`, you can see that `std::collections::HashMap` is just a wrapper around `hashbrown`:

```
use hashbrown::hash_map as base;

// ...

#[cfg_attr(not(test), rustc_diagnostic_item = "HashMap")]
#[stable(feature = "rust1", since = "1.0.0")]
#[rustc_insignificant_dtor]
pub struct HashMap<K, V, S = RandomState> {
    base: base::HashMap<K, V, S>,
}
```

I also checked the `rust-lang/rust` repo, it uses the latest version of `hashbrown`:

```rust
hashbrown = { version = "0.12", default-features = false, features = ['rustc-dep-of-std'] }
```

> [source](https://github.com/rust-lang/rust/blob/431c6f84990cdd08b68d46efff2efc75549934fe/library/std/Cargo.toml#L22)

I agree that we should add more benchmarks, but right now I don't have the capacity to do that.  However, since `hashbrown` is included in `std` anyways (just with a different default hasher) and we use the same hasher `hashbrown` uses by default, this change shouldn't affect performance at all, I just wanted to remove unnecessary dependencies.

Cargo.toml
torkleyy: I tried compiling with the `nightly` feature on stable, the necessary language features are not yet stabilized. I remember it's been forever that GAT are "about to be stabilized", but definitely worth keeping an eye out :)

Cargo.toml
torkleyy: done

xMAC94x (approved):

torkleyy:
Thanks for the reviews @xMAC94x :)

Issue #647, opened 2019-09-27
JoinIter::get allows mutable aliasing without the user writing any unsafe code.


allen-marshall:
## Description

Using the `JoinIter::get` method, it is possible for a user of `specs` to create two mutable references to the same component data, without the user writing any `unsafe` code. Here is a simple program that illustrates the issue:

```rust
use specs::{Builder, Component, DenseVecStorage, Join, World, WorldExt, WriteStorage};

#[derive(Default)]
struct TestComponent {
    value: u32,
}

impl Component for TestComponent {
    type Storage = DenseVecStorage<TestComponent>;
}

fn main() {
    let mut world = World::new();
    world.register::<TestComponent>();
    let entity = world.create_entity().with(TestComponent::default()).build();
    world.maintain();

    let mut storage: WriteStorage<TestComponent> = world.write_storage();
    let entities = world.entities();

    let mut join_iter = (&mut storage).join();
    let aliased_ref_0 = join_iter.get(entity, &entities).unwrap();
    let aliased_ref_1 = join_iter.get(entity, &entities).unwrap();

    println!("aliased_ref_0 is initially {}.", aliased_ref_0.value);
    aliased_ref_1.value += 1;
    println!(
        "After change to aliased_ref_1, aliased_ref_0 is now {}.",
        aliased_ref_0.value
    );
}
```

When I run this program, I get the following output:

```
aliased_ref_0 is initially 0.
After change to aliased_ref_1, aliased_ref_0 is now 1.
```

The issue doesn't seem too hard to avoid as long as I only use `JoinIter` for its intended purpose as an iterator, but even so, the fact that I can violate Rust's aliasing rules without writing `unsafe` seems like an issue. Perhaps `JoinIter::get` needs to be made private to the `specs` crate, or maybe it just needs to be marked `unsafe`? I imagine the same issue can occur with `JoinIter::get_unchecked`, though I haven't tested it.

## Meta

Rust version: 1.37.0 (2018 edition)
Specs version / commit: 0.15.1
Operating system: Ubuntu 18.04.3 LTS 64-bit

## Reproduction

1. Compile the Rust program given above.
2. Run the program.
3. See in the output that the value at aliased_ref_0 changed even though aliased_ref_0 wasn't mutated directly.

## Expected behavior

I would expect the above code not to compile, at least not without modifying it to include an `unsafe` block.

WaDelma:
Here is version of that where the lifetimes are explicit:
```rust
[...]

PR #765, opened 2023-03-01
Safety

Imberflur:
Fixes https://github.com/amethyst/specs/issues/647

See API changes section for description.

## Checklist

## Todo
* [x] Add more to PR description
* [x] Fix this https://github.com/amethyst/specs/pull/765#discussion_r1131532958
* [ ] ~~LendJoin uses `nougat` since I started working on it before GATs stabilized. I kept it because I thought it might be useful to add more iterator combinators (some combinators are currently impossible to implement with actual GATs). However, I think it would probably be prudent to remove this and use regular GATs for simplicity.~~ So I wanted to remove `nougat` but it turns out that I can't implement `for_each` without it (without either running into requiring `Self: 'static` due to `for<'_>` hrtb or not being able to have lifetimes in the GAT (`LendJoin::Type<'next>`) that aren't `'next` and aren't `'static`.
* [x] May want to create a quick 0.19.0 release first? See https://github.com/amethyst/specs/issues/764
* [x] Compare performance to make sure no significant regressions were introduced (and to satisfy my curiosity)
    * [x] Benchmarks (I compared several versions of `specs` using `ecs_bench_suite`, there is _maybe_ a 3% increase in `add_remove`, there was a significant increase (10+%) in deletion times due to `shred::MetaTable` changes, so I introduce a `nightly` feature that enables a more efficient implementation using `ptr_metadata`, `fetch` heavy code shows improvements due to the use of `atomic_refcell` in `shred` instead of the custom atomic cell implementation) 
    * [x] Veloren server
* [x] Publish new versions of `hibitset` and `shred` to crates.io and use those instead of git dependencies.
* [x] Update changelog

## API changes

This is a breaking change.

This PR fixes several soundness issues that I encountered when working on #737. These include:
* The `Join` implementation for `specs::storage::Entries` allows creating aliasing `&mut` references to the underlying storage (i.e. when not using the `JoinIter` as a lending/streaming iterator).
* The `Join` implementation and API of `specs::storage::RestrictedStorage` similarly can create mutable aliasing references to storages as well as to a specific component.
* The `Join` implementation of `Storage` allowed `DerefFlaggedStorage` to create aliasing `&mut` references to the internal events channel.
* `JoinIter::get` allows creating aliasing mutable refs, see #647. 
* Parallel iteration creates aliasing `&mut Join::Value` references.
* Under stacked borrows certain Join::get implementations were invalidating previously returned mutable references (e.g. by creating a slice reference to the whole underlying storage).

Many (but not all) of these issues stem from artificially lengthening the lifetime of the `&mut Join::Value` within `Join::get` implementations. We now avoid this and leverage alternative mechanisms including interior mutability and lending iteration.

In more detail, we:
* Introduce `LendJoin` trait or lending (aka streaming) iteration of joined values.
* Make some things like `Entries` now only implement `LendJoin` (i.e. remove unsound `Join` implementations).
* Implement `LendJoin` for everything that implements `Join`(manually, not a blanket impl since there are difference in the implementation).
* Move `JoinIter::get` to `JoinLendIter::get`.
* Add the `RepeatableLendGet` trait (which `JoinLendIter::get` requires). This is to facilitate destructive implementations of `LendJoin` where getting the same index more than once would be unsound (i.e. literally just the owned implementation of `LendJoin` for  `ChangeSet` which removes values as it iterates).
* Refactor `ParJoin` trait to not rely on `Join` implementation so we have additional flexibility to use different implementations and keep `&mut Join::Value` in `Join::get`.
* Make `Join`  an `unsafe` trait. The new `LendJoin` trait is also `unsafe` to implement.
* `Storage` no longer implements mutable `Join`s (the non-lending variant) for all `UnprotectedStorage`s. Instead this implementation requires that the storage implements `SharedGetMutStorage`. The trait which provides a `shared_get_mut(&self, id: Index)` method. Notice this takes a shared reference. Storages that can soundly implement this wrap their components in `UnsafeCell` to allow constructing a mutable reference from a shared reference.
* Revamp many pieces of safety documentation.
* Add `#![deny(unsafe_op_in_unsafe_fn)]` to make it easier to identify and document unsafe operations.
    
To try to catch any remaining UB, I ran the available tests under [Miri ](https://github.com/rust-lang/miri). This also identified some issues in dependencies for which I have submitted PRs to fix:
* https://github.com/amethyst/shred/pull/223
* https://github.com/amethyst/shred/pull/226
* https://github.com/amethyst/hibitset/pull/61
    
(We need to publish new versions of these to crates.io)

This PR adds Miri to the CI.

Additionally, `specs` exposed a `nightly` cargo feature that enabled additional APIs using GATs. Since GATs are now stabilized, I bumped the MSRV to be able to eliminate this feature and remove a bunch of `cfg` complexity.
   
Also I introduced `AccessMut` trait which is similar to `DerefMut` except it requires explicit use. The associated type`UnprotectedStorage::AccessMut<'_>` now requires `AccessMut` instead of `DerefMut`. This is to faciliate my work in https://github.com/amethyst/specs/pull/737 where I am exploring a flagged storage type that makes generating modification events more explicit. A blanket implementation of `AccessMut` for anything implementing `DerefMut` is included.

### DIFF
--- a/.github/workflows/deny.yml
+++ b/.github/workflows/deny.yml
@@ -15,13 +15,11 @@ jobs:
           - bans licenses sources
 
     # Prevent sudden announcement of a new advisory from failing ci:
-    # continue-on-error: ${{ matrix.checks == 'advisories' }}
-    # TODO: temp to allow git deps
-    continue-on-error: true
+    continue-on-error: ${{ matrix.checks == 'advisories' }}
 
 
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
       - uses: EmbarkStudios/cargo-deny-action@v1
         with:
           command: check ${{ matrix.checks }}
--- a/.github/workflows/deny.yml
+++ b/.github/workflows/deny.yml
@@ -15,7 +15,10 @@ jobs:
           - bans licenses sources
 
     # Prevent sudden announcement of a new advisory from failing ci:
-    continue-on-error: ${{ matrix.checks == 'advisories' }}
+    # continue-on-error: ${{ matrix.checks == 'advisories' }}
+    # TODO: temp to allow git deps
+    continue-on-error: true
+
 
     steps:
       - uses: actions/checkout@v2
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,13 +1,19 @@
+# Unreleased
+
+* MSRV to 1.65.0 [#7xx].
+
+[#7xx]: https://github.com/amethyst/specs/pull/7xx
+
 # 0.19.0 (2023-06-10)
 
 * Bump MSRV to 1.65.0 ([#766])
 * Added index where entity deletion stopped to the error returned from `WorldExt::delete_entities` ([#766])
 * Fix bug where deleting an entity with the wrong generation could clear the components of an existing entity. ([#766])
-* Bump shred to version `0.14.1`, MSRV to 1.60.0 ([shred changelog][shred-changelog], [#756])
+* Bump shred to version `0.14.1`, MSRV to 1.60.0 ([shred changelog][shred_changelog], [#756])
 
+[shred_changelog]: https://github.com/amethyst/shred/blob/master/CHANGELOG.md#0141-2022-07-14
 [#756]: https://github.com/amethyst/specs/pull/756
 [#766]: https://github.com/amethyst/specs/pull/766
-[shred-changelog]: https://github.com/amethyst/shred/blob/6b754812e304cf6c63ba0364a82a7e0e5025aaa4/CHANGELOG.md#0140-2022-07-12
 
 # 0.18.0 (2022-07-02)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,12 @@
 # Unreleased
 
-* MSRV to 1.70.0 [#7xx].
+* MSRV to 1.70.0 ([#765])
+* Significant refactor of `Join` and related traits to alleviate soundness
+  issues. Includes introduction of a lending/streaming join via the `LendJoin`
+  trait which is the new common denominator implemented by joinable types.
+  ([#765])
 
-[#7xx]: https://github.com/amethyst/specs/pull/7xx
+[#765]: https://github.com/amethyst/specs/pull/765
 
 # 0.19.0 (2023-06-10)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Unreleased
 
-* MSRV to 1.65.0 [#7xx].
+* MSRV to 1.70.0 [#7xx].
 
 [#7xx]: https://github.com/amethyst/specs/pull/7xx
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -62,26 +62,27 @@ shred = { git = "https://github.com/Imberflur/shred", branch = "metatable-fix", 
 specs-derive = { path = "specs-derive", version = "0.4.1" }
 
 [[example]]
+name = "async"
+[[example]]
 name = "basic"
-
-[[example]]
-name = "full"
-
-[[example]]
-name = "cluster_bomb"
-
 [[example]]
 name = "bitset"
-
 [[example]]
-name = "track"
-
+name = "cluster_bomb"
+[[example]]
+name = "full"
+[[example]]
+name = "lend_join"
+test = true
 [[example]]
 name = "ordered_track"
-
 [[example]]
 name = "saveload"
 required-features = ["serde"]
+[[example]]
+name = "slices"
+[[example]]
+name = "track"
 
 [[bench]]
 name = "benches_main"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,11 +24,11 @@ ahash = "0.7.6"
 crossbeam-queue = "0.3"
 # waiting on PR and new version to be published
 # hibitset = { version = "0.6.3", default-features = false }
-hibitset = { path = "../hibitset", default-features = false }
+hibitset = { git = "https://github.com/amethyst/hibitset", default-features = false }
 log = "0.4.8"
 # waiting on PR and new version to be published
 # shred = { version = "0.14.1", default-features = false }
-shred = { path = "../shred", default-features = false }
+shred = { git = "https://github.com/Imberflur/shred", branch = "metatable-fix", default-features = false }
 shrev = "1.1.1"
 tuple_utils = "0.4.0"
 nougat = "0.2.3"
@@ -58,7 +58,7 @@ ron = "0.7.1"
 rand = "0.8"
 serde_json = "1.0.48"
 # shred = { version = "0.14.1", default-features = false, features = ["shred-derive"] }
-shred = { path = "../shred", default-features = false, features = ["shred-derive"] }
+shred = { git = "https://github.com/Imberflur/shred", branch = "metatable-fix", default-features = false, features = ["shred-derive"]}
 specs-derive = { path = "specs-derive", version = "0.4.1" }
 
 [[example]]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,13 +22,9 @@ autobenches = false
 [dependencies]
 ahash = "0.7.6"
 crossbeam-queue = "0.3"
-# waiting on PR and new version to be published
-# hibitset = { version = "0.6.3", default-features = false }
-hibitset = { git = "https://github.com/amethyst/hibitset", default-features = false }
+hibitset = { version = "0.6.4", default-features = false }
 log = "0.4.8"
-# waiting on PR and new version to be published
-# shred = { version = "0.14.1", default-features = false }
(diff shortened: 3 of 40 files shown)


src/storage/mod.rs
```diff
@@ -365,6 +400,40 @@ where
...
+    /// # Safety
+    ///
+    /// May only be called if `id` is not present in the mask.
+    #[inline(always)]
+    unsafe fn not_present_insert(&mut self, id: Index, value: T) {
+        // SAFETY: The mask was previously empty, so it is safe to
+        // insert. We immediately add the value to the mask below and
+        // unwinding from the `insert` call means that we don't need to
+        // include the value in the mask. `BitSet::add` won't unwind on 32-bit
+        // and 64-bit platforms since OOM aborts and any overflow in capacity
```
Imberflur, line 414: > `BitSet::add` won't unwind on 32-bit and 64-bit platforms since OOM aborts

Seems like this might not always be the case, e.g. see https://blog.rust-lang.org/2023/03/09/Rust-1.68.0.html

Imberflur: Fixed

Cargo.toml
```diff
@@ -14,19 +14,24 @@ license = "MIT OR Apache-2.0"
...
 # the `storage_cmp` and `storage_sparse` benches are called from `benches_main`
 autobenches = false
 
 [dependencies]
 ahash = "0.7.6"
 crossbeam-queue = "0.3"
-hibitset = { version = "0.6.3", default-features = false }
+# waiting on PR and new version to be published
+# hibitset = { version = "0.6.3", default-features = false }
+hibitset = { git = "https://github.com/amethyst/hibitset", default-features = false }
```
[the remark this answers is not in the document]
Imberflur: now waiting on new version to be published iirc

Imberflur:
Exciting news! I finally got a chance to update veloren to use this and profile it and I don't see any particular regressions. Might try to post some tracy pictures later.

I think this should be ready to merge soon.

Imberflur:
Here are some profiling results in veloren, "this trace" is before changes here and "external trace" is after. I focused on two systems. "character_behavior" which has a join over a lot of component types and "phys" which has multiple joins over fewer component types of which some joins are parallel ones. It seems like there are no regressions and potentially a slight improvement (the profiling conditions have room to be more strictly controlled so I would not trust that this improvement is as significant as it appears here).  

image  

image

src/bitset.rs
```diff
@@ -2,39 +2,83 @@
 //!
 //! Normally used for `Join`s and filtering entities.
 
-#![cfg_attr(rustfmt, rustfmt_skip)]
+// TODO: rustfmt bug (probably fixed in next rust release)
+// #![cfg_attr(rustfmt, rustfmt::skip)]
```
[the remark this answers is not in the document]
Imberflur: seems good!

.github/workflows/deny.yml
```diff
@@ -15,7 +15,10 @@ jobs:
           - bans licenses sources
 
     # Prevent sudden announcement of a new advisory from failing ci:
-    continue-on-error: ${{ matrix.checks == 'advisories' }}
+    # continue-on-error: ${{ matrix.checks == 'advisories' }}
+    # TODO: temp to allow git deps
+    continue-on-error: true
+
```
Imberflur, line 21: should be able to remove this

CHANGELOG.md
```diff
@@ -1,13 +1,19 @@
+# Unreleased
+
+* MSRV to 1.70.0 [#7xx].
+
+[#7xx]: https://github.com/amethyst/specs/pull/7xx
```
Imberflur, line 5: Need to update this

src/storage/mod.rs
```diff
@@ -503,75 +738,96 @@ where
...
+        &mut *self
+    }
+}
+
 /// Used by the framework to quickly join components.
 pub trait UnprotectedStorage<T>: TryDefault {
     /// The wrapper through with mutable access of a component is performed.
-    #[cfg(feature = "nightly")]
-    type AccessMut<'a>: DerefMut<Target = T>
+    type AccessMut<'a>: AccessMut<Target = T>
```
Imberflur, line 765: I switched this from `DerefMut` to a new `AccessMut` trait to allow types that are automatically deferenceable. This is for a flagged storage variant that I'm planning that would be more explicit about when a component is marked as mutated. Most cases aren't effected by this unless they are generic over the component type. `AccessMut` also still has `Deref` as a super trait like `DerefMut` does.

src/storage/mod.rs
```diff
@@ -591,11 +847,52 @@ pub trait UnprotectedStorage<T>: TryDefault {
...
     unsafe fn drop(&mut self, id: Index) {
-        self.remove(id);
+        // SAFETY: Requirements passed to the caller.
+        unsafe { self.remove(id) };
     }
 }
 
+/// Used by the framework to mutably access components in contexts where
+/// exclusive access to the storage is not possible.
+pub trait SharedGetMutStorage<T>: UnprotectedStorage<T> {
```
Imberflur, line 862: The base `UnprotectedStorage` trait means that we can implement `LendJoin` for it. `SharedGetMutStorage` enables implementing `Join`. And finally `DistinctStorage` enables `ParJoin`.

src/storage/restrict.rs
```diff
@@ -146,152 +158,470 @@ where
...
+/// Pairs a storage with an index, meaning that the index is guaranteed to
+/// exist.
+///
+/// Yielded by `join`/`par_join` on `&mut storage.restrict_mut()`.
+pub struct PairedStorageWriteShared<'rf, C: Component> {
+    index: Index,
+    storage: SharedGetOnly<'rf, C, C::Storage>,
 }
 
-impl<'rf, 'st, C, S, B, Restrict> PairedStorage<'rf, 'st, C, S, B, Restrict>
```
Imberflur, line 256: With the additional complexity of adding the `LendJoin` case it was much easier to have several separate paired storage types rather than using generic marker parameters.

src/storage/restrict.rs
```diff
@@ -146,152 +158,470 @@ where
...
+/// let mut b = joined.next().unwrap();
+/// // sound since Pos::Storage is a DistinctStorage
+/// std::thread::scope(|s| {
+///     s.spawn(move || {
+///         a.get_mut();
+///     });
+/// });
+/// b.get_mut();
+/// ```
+fn _dummy() {}
```
Imberflur, line 505: compile_fail test to check for mistakes

src/storage/storages.rs
```diff
@@ -154,112 +178,207 @@ impl<T> SliceAccess<T> for DenseVecStorage<T> {
...
 unsafe impl<T> DistinctStorage for DenseVecStorage<T> {}
 
 /// A null storage type, used for cases where the component
 /// doesn't contain any data and instead works as a simple flag.
-pub struct NullStorage<T>(T);
+pub struct NullStorage<T>(PhantomData<T>);
 
-impl<T> UnprotectedStorage<T> for NullStorage<T>
-where
-    T: Default,
```
Imberflur, line 220: NullStorage no longer requires `T: Default`, since we can soundly conjure up ZSTs provided that they represent instances that this has previously taken ownership of

Imberflur (commented):
Here are some notes that will hopefully help add some context for reviewers

tests/no_parallel.rs
```diff
@@ -1,4 +1,5 @@
 #![cfg(not(feature = "parallel"))]
+// TODO: ensure we run these with miri too
```
Imberflur, line 3: Looks like a TODO to address in this PR

docs/tutorials/src/12_tracked.md
```diff
@@ -117,8 +117,8 @@ fetch the component as mutable if/when needed.
 for (entity, mut comp) in (&entities, &mut comps.restrict_mut()).join() {
     // Check whether this component should be modified, without fetching it as
     // mutable.
-    if comp.get_unchecked().condition < 5 {
-         let mut comp = comp.get_mut_unchecked();
+    if comp.get().condition < 5 {
+         let mut comp = comp.get_mut();
```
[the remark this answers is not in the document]
cpetig: The _unchecked suffix always reminds me of unsafe access (e.g. https://doc.rust-lang.org/std/primitive.slice.html#method.get_unchecked ), so this rename is good.

xMAC94x (approved):
hard to check such a massive PR. i had a look through all of it and am positive that its an improvement. LGTM

Imberflur:
@xMAC94x thanks for taking a look :heart: 

I will wait till this weekend in case anyone else is interested in reviewing. If anyone interested needs more time just let me know.

docs/tutorials/src/12_tracked.md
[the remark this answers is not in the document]
Imberflur: > Update: I later realized that get indeed remains unsafe, and the restriction that the item must exists sounds very similar to the unchecked variant of slice. 

The method discussed here has always been safe. There are other unsafe methods with the same name that this PR touches, so maybe you are thinking of one of those.

zesterer:
I just wanted to say: thanks for pushing this all the way to completion :) I would have reviewed, but I've been busy with other things recently. Nice work!

PR #766, opened 2023-06-06
Delete components after killing entities to avoid clearing components when the entity has the wrong generation

Imberflur:
## Checklist

* ~~[ ] I've added a demonstration of the new feature to one or more examples~~
* ~~[ ] I've updated the book to reflect my changes~~
* ~~[ ] Usage of new public items is shown in the API docs~~

## API changes

`WorldExt::delete_entities` method now includes the index of the entity with the wrong generation to indicate to the user where it stopped (since entities after that are not deleted).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-# 0.19.0 (2022-07-15)
+# 0.19.0 (2023-06-10)
 
 * Bump MSRV to 1.65.0 ([#766])
 * Added index where entity deletion stopped to the error returned from `WorldExt::delete_entities` ([#766])
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # 0.19.0 (2022-07-15)
 
+* Bump MSRV to 1.65.0 ([#766])
 * Added index where entity deletion stopped to the error returned from `WorldExt::delete_entities` ([#766])
 * Fix bug where deleting an entity with the wrong generation could clear the components of an existing entity. ([#766])
 * Bump shred to version `0.14.1`, MSRV to 1.60.0 ([shred changelog][shred-changelog], [#756])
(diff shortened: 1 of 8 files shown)


zesterer:
I was about to ping @Imberflur to discuss this, then realised you'd just changed your profile picture :grinning:

CHANGELOG.md
```diff
@@ -1,8 +1,12 @@
 # 0.19.0 (2022-07-15)
 
+* Bump MSRV to 1.65.0 ([#766])
+* Added index where entity deletion stopped to the error returned from `WorldExt::delete_entities` ([#766])
+* Fix bug where deleting an entity with the wrong generation could clear the components of an existing entity. ([#766])
```
xMAC94x, line 5: should that belong to the new version ?

Imberflur: the latest release is 0.18.0 so I think this is the new version

xMAC94x: Urrrrrggg,
weird situation: Somehow it got the change https://github.com/amethyst/specs/commit/474cdeaee761b38d0e4d0ac3e8bd73e027b9da02 to do an release, but the actuall release never happened. noone pushed it to crates.io or did a release tag or github release.
I dont have full permissions to do all of that.
So we can now decide to what we do. Either release a 0.19 as intended back then. or just add it here. but in that case could you change back the Title to `Unreleased` and remove the wrong date ?

Imberflur: Oof, good news is that I have preemptively acquired the permissions for releasing a new version when working on my other major PR (that still needs a few finishing touches).

I could change it back to `Unreleased` or just bump the date so we can put out a new release after merging this? (I'm in favor of getting a new release out since there is an issue requesting it)

xMAC94x: both is fine :)

Issue #769, opened 2023-10-23
ahash 7.5 dep yanked


juliusl:
## Description

```sh
error: failed to select a version for the requirement `ahash = "^0.7.6"`
candidate versions found which didn't match: 0.8.5, 0.8.4, 0.4.1, ...
location searched: crates.io index
required by package `specs v0.20.0`
```

https://crates.io/crates/ahash/versions -- 7.5 is yanked

Related: 
https://github.com/tkaitchuck/aHash/issues/174

## Meta

Rust version:
Specs version / commit:
Operating system:

Imberflur:
It looks like a 0.7.7 was published. Does this still fail to select a version?

juliusl:
@Imberflur yeah the 0.7.7 backport fixed things

PR #749, opened 2022-06-21
xmac94x/GitHub actions

xMAC94x:
Pure CI change, just get rid of travis and go to actions (i hope this works)

PR #770, opened 2023-10-23
Bump ahash dependency version

joshlf:
The previous version had a bug, and was yanked.

## Checklist

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,7 +20,7 @@ rust-version = "1.70.0"
 autobenches = false
 
 [dependencies]
-ahash = "0.7.6"
+ahash = "0.8.5"
 crossbeam-queue = "0.3"
 hibitset = { version = "0.6.4", default-features = false }
 log = "0.4.8"


juliusl:
track #769

juliusl:
@joshlf shred needs to be bumped too

joshlf:
> @joshlf shred needs to be bumped too

To 0.15.1?

joshlf:
Oh sorry, do you mean that I should open a separate PR to the shred repo? https://github.com/amethyst/shred

zesterer:
Thanks all. It's unfortunate that this is required since the original issue in question isn't of particular concern to `specs`, but ah well.

zesterer:
Thanks all. It's unfortunate that this is required since the original issue in question isn't of particular concern to `specs`, but ah well.

The licensing issue caught by CI is a problem, however. I'd argue the fault here lies with `ahash` since it's pulling in a dependency (`zerocopy`) that is not compatible with its own license.

joshlf:
Would it be an acceptable solution to add BSD-2-Clause as an allowed license? BSD-3-Clause is already allowed.

zesterer:
It might be, yes. @kvark @Imberflur Do either of you have thoughts on this?

Imberflur:
@zesterer 

`cargo deny` CI was added in https://github.com/amethyst/specs/pull/749 and there was a bit discussion on it here https://github.com/amethyst/specs/pull/749#discussion_r903036768. I think the current set of allows matches what was needed for the dependencies at the time.

I'm not familiar with what kind of licensing effect would arise from allowing BSD-2-Clause and I don't have any particular opinions here.

It is possible to add exceptions for specific crates (rather than a blanket allow for a license), such that taking on new dependencies with that license requires a conscious decision: https://embarkstudios.github.io/cargo-deny/checks/licenses/cfg.html#the-allow-field

We could potentially allow this for now and if having BSD-2-Clause deps ends up being an issue later, find a path to remove them?

Also a 0.7.7 version of ahash was published so this PR should not be essential for dealing with the other versions being yanked. Ofc, it is nice to update to the latest version regardless.

joshlf:
> @zesterer
> 
> `cargo deny` CI was added in #749 and there was a bit discussion on it here [#749 (comment)](https://github.com/amethyst/specs/pull/749#discussion_r903036768). I think the current set of allows matches what was needed for the dependencies at the time.
> 
> I'm not familiar with what kind of licensing effect would arise from allowing BSD-2-Clause and I don't have any particular opinions here.
> 
> It is possible to add exceptions for specific crates (rather than a blanket allow for a license), such that taking on new dependencies with that license requires a conscious decision: https://embarkstudios.github.io/cargo-deny/checks/licenses/cfg.html#the-allow-field
> 
> We could potentially allow this for now and if having BSD-2-Clause deps ends up being an issue later, find a path to remove them?
> 
> Also a 0.7.7 version of ahash was published so this PR should not be essential for dealing with the other versions being yanked. Ofc, it is nice to update to the latest version regardless.

I found [this](https://imagej.net/licensing/bsd#modified-bsd-license) on the difference between the licenses:

> The Modified or New BSD (or BSD 3-clause) license is the same as BSD-2, but with an additional clause prohibiting the names of the authors from being used to endorse or promote products relating to the software.

Indeed, using the text of the 2- and 3-clause variants from [here](https://choosealicense.com/licenses/bsd-2-clause/) and [here](https://choosealicense.com/licenses/bsd-3-clause/), this appears to be an accurate description:

```text
$ diff 2_CLAUSE 3_CLAUSE 
1c1
< BSD 2-Clause License
---
> BSD 3-Clause License
13a14,17
> 
> 3. Neither the name of the copyright holder nor the names of its
>    contributors may be used to endorse or promote products derived from
>    this software without specific prior written permission.
```

IANAL, but I'd assume that if your project is already comfortable with BSD-3-Clause, then the same logic would apply for being comfortable with a license which imposes strictly fewer restrictions on your project.

juliusl:
@joshlf sorry for the late reply, To clarify I'm just pointing it out to note that shred needs to be updated and released before this can be merged. Ran into it when I while I was patching for this in some of my other repos.

joshlf:
> @joshlf sorry for the late reply, To clarify I'm just pointing it out to note that shred needs to be updated and released before this can be merged. Ran into it when I while I was patching for this in some of my other repos.

Oh gotcha, makes sense!

Imberflur:
Superceded by #781. Sorry for the delay here, and thanks for resolving the licensing issue! (even if BSD-2-Clause might have been fine since BSD-3-Clause is already allowed)

PR #774, opened 2024-04-22
Fix #773

### LINKED ISSUE #773: specs-derive better 'storage' macro parameter
Hi and thanks for specs.

## Description
Currently it is not possible to specify a storage type with specs-derive that expects special generics. This is because specs-derive assumes `<Self> `https://github.com/amethyst/specs/blob/708c3ac436e9cad285babd61caf54fa15a0edace/specs-derive/src/lib.rs#L73 and this cannot be overwritten. It should be possible to specify other generics.
## Motivation
 It is currently not possible to specify e.g. `FlaggedStorage` with the 'storage' macro.

```rust
#[derive(Component)]
#[storage(DerefFlaggedStorage<Self,DenseVecStorage<Self>>)]
pub struct Foo {}
```

Although 'Component' can be implemented traditionally, it is irritating if this has to be done for some types of storage and not for others. This also means additional work, which the derive macro is designed to avoid from the outset.
## Drawbacks
 None

* Is it a breaking change?
   No - it is not
   No - I would actually argue, it implements expected behaviour.
## Unresolved questions
None

---

I asked about this on [gitter](https://matrix.to/#/!TiVTQzYllgpxfEpvjI:gitter.im/$wnvO9dns5z9nIyAzF2scAqgyi-OJGfOkAQ8X5QGiwmA?via=gitter.im&via=matrix.freyachat.eu&via=matrix.org), and will create a pull request, which 'fixes' this request.

-ClaasJG

ClaasJG:
Hi,

This fixes #773 . 

## Checklist

* [ ] I've added tests for all code changes and additions *(where applicable)*

I thought this was already possible based on the resources mentioned in the list above.
Therefore, I have not changed anything and don't think anything needs to be added / changed.

-ClaasJG

### DIFF
--- a/specs-derive/src/lib.rs
+++ b/specs-derive/src/lib.rs
@@ -15,7 +15,7 @@ extern crate syn;
 use proc_macro::TokenStream;
 use syn::{
     parse::{Parse, ParseStream, Result},
-    DeriveInput, Path,
+    DeriveInput, Path, PathArguments,
 };
 
 mod impl_saveload;
@@ -68,9 +68,14 @@ fn impl_component(ast: &DeriveInput) -> proc_macro2::TokenStream {
         })
         .unwrap_or_else(|| parse_quote!(DenseVecStorage));
 
+    let additional_generics = match storage.segments.last().unwrap().arguments {
+        PathArguments::AngleBracketed(_) => quote!(),
+        _ => quote!(<Self>),
+    };
+
     quote! {
         impl #impl_generics Component for #name #ty_generics #where_clause {
-            type Storage = #storage<Self>;
+            type Storage = #storage #additional_generics;
         }
     }
 }
--- a/specs-derive/src/lib.rs
+++ b/specs-derive/src/lib.rs
@@ -28,7 +28,17 @@ mod impl_saveload;
 /// use specs::storage::VecStorage;
 ///
 /// #[derive(Component, Debug)]
-/// #[storage(VecStorage)] // This line is optional, defaults to `DenseVecStorage`
+/// #[storage(VecStorage<Self>)] // This line is optional, defaults to `DenseVecStorage<Self>`
+/// struct Pos(f32, f32, f32);
+/// ```
+///
+/// When the type parameter is `<Self>` it can be omitted i.e.:
+///
+///```rust,ignore
+/// use specs::storage::VecStorage;
+///
+/// #[derive(Component, Debug)]
+/// #[storage(VecStorage)] // Equals to #[storage(VecStorage<Self>)]
 /// struct Pos(f32, f32, f32);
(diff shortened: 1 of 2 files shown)


specs-derive/src/lib.rs
```diff
@@ -15,7 +15,7 @@ extern crate syn;
 use proc_macro::TokenStream;
 use syn::{
     parse::{Parse, ParseStream, Result},
-    DeriveInput, Path,
+    DeriveInput, Path, PathArguments,
 };
 
 mod impl_saveload;
```
Imberflur, line 21: Can you add an example onto the derive macro docs below demonstrating the new usage that this enables and noting how it differs from the case where `<Self>` is assumed?

Imberflur (approved):
Hi! Thanks for this improvement, it looks great!

I left one suggestion for improving the docs.

specs-derive/src/lib.rs
```diff
@@ -68,9 +68,14 @@ fn impl_component(ast: &DeriveInput) -> proc_macro2::TokenStream {
         })
         .unwrap_or_else(|| parse_quote!(DenseVecStorage));
 
+    let additional_generics = match storage.segments.last().unwrap().arguments {
+        PathArguments::AngleBracketed(_) => quote!(),
+        _ => quote!(\<Self>)
```
Imberflur, line 73:
```suggestion
        _ => quote!(<Self>)
```

ClaasJG: I do not know how this happens - sry

ClaasJG:
I added an explanation, that #[storage(...)] accepts storage types with type parameters.
English is not my first language and the text feels clunky to me.

specs-derive/src/lib.rs
ClaasJG: I misunderstood I think.
007d5bb7cbeb77a0fcaf6386708b86389e9d4d96 adds (maybe unnecessary) Info to the tutorial.
fd59c30ba7daa86fe3ed30ae75d9943ff3665329 adds info to the docs.

Imberflur: I just meant the docs, but the other addition looks fine to me! Sorry, I would have put my comment directly on the section where the docs are but GitHub's review doesn't allow that.

ClaasJG:
I dont know how to fix the failing check on nightly, nor am I sure, it is caused by this pull request.

Imberflur:
> I dont know how to fix the failing check on nightly, nor am I sure, it is caused by this pull request.

No worries, it isn't caused by this PR.

I'm planning to do some quick testing to confirm works with the new derive logic, and will merge this after I do that.

Imberflur (approved):

PR #781, opened 2024-05-13
update aHash and dev dependencies

xMAC94x:
Note: aHash is a patch version higher than #770 should not have the license issues

## Checklist


Imberflur (approved):
Yay! Looks good to merge. Need to publish a new shred version soon so we can depend on the bumps there too.

